PR #2, opened 2016-09-07
Core-8176 and Core-8168 bug fixes

sri2k1us:
Core-8168 disable save button when viewing metadata on read-only resources
Core-8176 fix refresh button in analyses window when using search filter. 

Apologies opening a single PR for 2 bugs. Will get right the next time :)


ash3rz:
LGTM :shipit:

psarando:
I didn't catch any issues :shipit:

PR #5, opened 2016-09-14
Revive most DataSearchQueryBuilder tests with JSONAssert

ianmcorvidae:
I wasn't quite able to figure out how to revive `testBuildQuery()` since that doesn't assert equality, but containment. But, this is still a big improvement, and brings back most of the functionality, so I'm putting it up.


strootman:
I ran the tests, they pass. I debugged the assertions, just to see that they are doing what they appear to be doing, and they are.
The new dependency is added in the test compile scope.

LGTM
:shipit:

PR #6, opened 2016-09-14
CORE-3701 Resize oversize window from saved session

ash3rz:
If a user saves their session while viewing the DE with a big monitor, then logs back in while using a smaller monitor, their windows will be too big to access any of the window buttons (close, minimize, etc.).

Now their saved windows will be resized (if needed) to fit the desktop view so they can have access to the entire window.


sri2k1us:
:shipit:

ash3rz:
Thanks for the reviews!

PR #7, opened 2016-09-14
CORE-4364 Disable rating private apps

ash3rz:
Currently, it's possible to attempt to rate an app in Apps Under Development which results in the user getting an error that they cannot rate private apps.  It is not possible to rate external apps, but it does visually look like you could rate an external app (it has the little white glove cursor when hovering).

This PR makes it so that hovering over the private apps and external apps rating cells has the regular black pointer cursor to convey that they are not clickable.  This also adds a tooltip to the private app rating cell that lets the user know that rating private apps is not supported.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/theme/base/client/apps/cells/AppRatingCellDefaultAppearance.java
+++ b/de-lib/src/main/java/org/iplantc/de/theme/base/client/apps/cells/AppRatingCellDefaultAppearance.java
@@ -27,6 +27,8 @@ public class AppRatingCellDefaultAppearance implements AppRatingCell.AppRatingCe
         @ClassName("apps_icon")
         String appsIcon();
 
+        String disabledRating();
+
         @ClassName("disabled_unrate_button")
         String disabledUnrateButton();
     }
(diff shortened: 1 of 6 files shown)


sri2k1us:
Looks good to me :shipit:

de-lib/src/main/java/org/iplantc/de/theme/base/client/apps/cells/AppRatingCellDefaultAppearance.java
```diff
@@ -181,6 +188,13 @@ public void render(SafeHtmlBuilder sb, App app) {
         int total = app.getRating().getTotal();
         // Build five rating stars
         for (int i = 0; i < ratings.size(); i++) {
+            if (!app.isPublic() || app.getAppType().equalsIgnoreCase(App.EXTERNAL_APP)) {
+                sb.append(templates.imgCell("Rating-" + i,
+                                            ratings.get(i),
+                                            resources.css().disabledRating(),
+                                            resources.whiteStar().getSafeUri()));
```
strootman: Should the Star even be shown? Looking at these changes, it almost seems like a bug that hovering over the stars doesn't work.

ash3rz: I guess that's up for debate.  IMO users would be more confused that there are no stars and think it's a bug or think their private apps could never be rated versus having a tooltip that tells them why they can't currently rate those apps.  And then with external apps we let them know it's just not supported.

PR #8, opened 2016-09-14
CORE-4012 Disable Copy Link button after link deleted

ash3rz:
When a user goes to Create a Public Link to a file, in the Manage Data Links dialog that opens, the Copy Link button will remain enabled when no link is selected (in particular, after a link has been deleted).  Clicking on the Copy Link button then opens the Copy Link dialog with no URL.

This PR does some minor refactoring and then disables the Copy Link button after a link has been deleted.


ash3rz:
Thanks for the review!

PR #9, opened 2016-09-15
CORE-5543 update hint texts in data window.

sri2k1us:


sri2k1us:
LOL. I don't how the hell i got that number. It should be CORE-5543.

sri2k1us:
Thanks for the review.

PR #11, opened 2016-09-19
CORE-8323 fix NPE when  'New Folder' is clicked without selecting

sri2k1us:
a parent folder.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/diskResource/client/views/toolbar/DiskResourceViewToolbarImpl.java
+++ b/de-lib/src/main/java/org/iplantc/de/diskResource/client/views/toolbar/DiskResourceViewToolbarImpl.java
@@ -2,7 +2,6 @@ package org.iplantc.de.diskResource.client.views.toolbar;
 
 import org.iplantc.de.client.models.UserInfo;
 import org.iplantc.de.client.models.diskResources.DiskResource;
-import org.iplantc.de.client.models.diskResources.DiskResourceAutoBeanFactory;
 import org.iplantc.de.client.models.diskResources.File;
 import org.iplantc.de.client.models.diskResources.Folder;
 import org.iplantc.de.client.models.viewer.InfoType;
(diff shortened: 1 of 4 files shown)


de-lib/src/main/java/org/iplantc/de/diskResource/client/views/toolbar/DiskResourceViewToolbarImpl.java
```diff
@@ -495,7 +498,13 @@ void onMoveToTrashClicked(SelectionEvent<Item> event) {
 
     @UiHandler("newFolderMi")
     void onNewFolderClicked(SelectionEvent<Item> event) {
-        presenter.onCreateNewFolderSelected(selectedFolder);
+        if(selectedFolder == null) {
+            Folder parent = drAbFactory.folder().as();
```
ash3rz: I think we're trying to keep AutoBean stuff out of the views.  Could we move this new logic to the presenter (and update/add tests)?

de-lib/src/main/java/org/iplantc/de/diskResource/client/views/toolbar/DiskResourceViewToolbarImpl.java
```diff
@@ -495,7 +498,13 @@ void onMoveToTrashClicked(SelectionEvent<Item> event) {
...
+            parent.setPath(UserInfo.getInstance().getHomePath());
+            presenter.onCreateNewFolderSelected(parent);
```
ash3rz: If you're feeling ambitious, we could probably refactor this to using events instead of having a reference to the presenter 😝 Not needed for this PR though.

de-lib/src/main/java/org/iplantc/de/diskResource/client/views/toolbar/DiskResourceViewToolbarImpl.java
sri2k1us: Good point. I have refactored code and added some tests.

de-lib/src/main/java/org/iplantc/de/diskResource/client/views/toolbar/DiskResourceViewToolbarImpl.java
sri2k1us: 👍

ash3rz:
LGTM 👍

sri2k1us:
Thanks!

PR #13, opened 2016-09-19
CORE-8163 Add more static IDs to Belphegor

ash3rz:
This PR is to add more static IDs to Belphegor as requested by QA.

* Each tab in Belphegor should have its own ID (switched to `DETabPanel` class instead of Sencha's `TabPanel` which allows for the tabs to have IDs)
* The App Catalog tab didn't have any IDs so those were added
  * I created `DETree` abstract class which allows any child class to specify a `generateDebugID` method for debug IDs on the tree nodes
* The column headers in grids all got static IDs (with use from StaticIDHelper class)

* I also patched up a random NPE that happens if you have saved sessions enabled, save a session with no windows open, and then log in.


ash3rz:
Anyone mind if I close this?  It's been open for some time now.

psarando:
I haven't looked at this since the latest updates.
I didn't catch any issues :shipit:

ash3rz:
Cool - thanks!

PR #14, opened 2016-09-19
CORE-5888 Refresh file viewers after save + refactor + tests

ash3rz:
When a user opened up a tabular file, saving changes in one file viewer did not cause the other file viewer to update.  The user had to manually refresh the file by hitting the Refresh button for that viewer.  This PR adds logic for cycling through the list of file viewers and updates all other views with the latest changes after a file is saved.

I also refactored out some JSON object manipulations and got the tests for FileViewerPresenterImpl started (though they're not finished, but can be finished later).


sri2k1us:
:shipit:

ash3rz:
Cool - thanks for the review!

PR #15, opened 2016-09-20
CORE-6530 Remove file from favorites when de-favoriting in Favorites folder

ash3rz:
In the Data window, if you try to de-favorite a file/folder while viewing the Favorites folder, the view does not refresh and the file/folder remains in the Favorites folder (visually) until you refresh.

I also added some tests since we didn't have anything for GridViewPresenterImpl.


sri2k1us:
:shipit:

ash3rz:
Thanks for the review!

PR #17, opened 2016-09-21
Core 3443

sri2k1us:
This a PR for improvements done to the Notification View and window. Following updates are done

- Mark a notification as read when a message is clicked. 
- New / unread notifications will be highlighted using blue color
- Notification window has new button to mark notifications as read
- Notification window will highlighted using blue color.
- Notification count is updated when messages are marked as seen from the Notification window.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/notifications/client/presenter/NotificationPresenterImpl.java
+++ b/de-lib/src/main/java/org/iplantc/de/notifications/client/presenter/NotificationPresenterImpl.java
@@ -9,19 +9,22 @@ import org.iplantc.de.client.services.MessageServiceFacade;
 import org.iplantc.de.client.services.callbacks.NotificationCallback;
 import org.iplantc.de.client.util.JsonUtil;
 import org.iplantc.de.commons.client.ErrorHandler;
+import org.iplantc.de.commons.client.info.ErrorAnnouncementConfig;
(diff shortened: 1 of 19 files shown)


ash3rz:
I believe you'll want to close and reopen this against the master branch.

de-lib/src/main/java/org/iplantc/de/notifications/client/presenter/NotificationPresenterImpl.java
```diff
@@ -160,6 +146,7 @@ private void addEventHandlers() {
     public void onNotificationGridRefresh(NotificationGridRefreshEvent event) {
         if (listStore.size() > 0) {
             toolbar.setDeleteAllButtonEnabled(true);
+            toolbar.setMarkAsSeenButtonEnabled(true);
         } else {
             toolbar.setDeleteAllButtonEnabled(false);
```
sri2k1us: Good catch but I think line ` toolbar.setMarkAsSeenButtonEnabled(true);` should be removed since we don't want 'mark as seen' button enabled on Grid refresh event. We want them enabled only when an unread notification is selected. Please see line 201. I will commit a change to remove this code from `onNotificationGridRefresh()` method and re-open this PR against master. Thanks for the review.

sri2k1us:
I will re-open this PR against master.

PR #20, opened 2016-09-23
CORE-8148 Add Refresh button to Apps Window

ash3rz:
Since all the other DE windows have Refresh buttons, the Apps window should probably have a Refresh button as well.

This enables the user to be able to refresh the categories, hierarchies, and app listing at any time, and the user should be brought back to whatever category/hierarchy they had selected before hitting Refresh.

Given the nature of ontologies and how they can change at any time by the admins, there's also a lot of logic added to cover the scenario where the user has hit Refresh and the hierarchy they were viewing no longer exists.  This can happen if the user has had the apps window open for a long time, the admin has published a new ontology that no longer has the hierarchy they were viewing, and then the user hits Refresh.  In this case, the user will be brought back to the Apps Under Development category.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/hierarchies/OntologyHierarchiesPresenterImpl.java
+++ b/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/hierarchies/OntologyHierarchiesPresenterImpl.java
@@ -105,6 +105,42 @@ public class OntologyHierarchiesPresenterImpl implements OntologyHierarchiesView
         }
     }
 
+    class FilteredHierarchyCallback implements AsyncCallback<OntologyHierarchy> {
+        private final Tree<OntologyHierarchy, String> tree;
+        private final OntologyHierarchy root;
+        private final OntologyHierarchy selectedHierarchy;
+
+        public FilteredHierarchyCallback(Tree<OntologyHierarchy, String> tree,
+                                         OntologyHierarchy root,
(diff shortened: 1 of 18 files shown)


de-lib/src/main/java/org/iplantc/de/apps/client/presenter/hierarchies/OntologyHierarchiesPresenterImpl.java
```diff
@@ -105,6 +106,58 @@ public void onSuccess(final App appDetails) {
...
+            // which will allow the children to know the full path from its parent to node
+            ontologyUtil.getOrCreateHierarchyPathTag(result);
+            addHierarchies(tree.getStore(), null, result.getSubclasses());
+            checkForDesiredHierarchy();
+
+            tree.unmask();
+        }
+
+        void checkForDesiredHierarchy() {
+            callbackCount++;
```
ash3rz: I'd like some eyes on this portion.  I'm not sure if this is the best way to keep track of whether the hierarchy the user was previously viewing is still in the ontology after they've Refreshed.
I'm basically keeping a counter that increments per callback. Once I've reached the last callback's return, I check to see if the hierarchy was ever found. If not, then I fire an event telling `AppCategoryPresenterImpl` to go ahead and select/load the `Apps under Development` category instead.  Since callbacks are asynchronous I'm not sure if I'm missing anything.

ash3rz:
I have some refactoring to do for this.  I'll close and reopen when it's ready.

PR #21, opened 2016-09-26
CORE-8331 Add static ids for "Select Metadata Template" dialog.

sri2k1us:


de-lib/src/main/java/org/iplantc/de/diskResource/client/views/metadata/dialogs/SelectMetadataTemplateDialog.java
```diff
@@ -52,8 +52,15 @@
 
     private MetadataView.Presenter.Appearance appearance;
 
+    private String BASE_ID = "select_metadata_template";
```
ash3rz: I think for consistency these IDs should probably be defined with the rest of the [Metadata ids](https://github.com/cyverse-de/ui/blob/f69465f48b2a62b5a4a179e0fc5c1006f884eec4/de-lib/src/main/java/org/iplantc/de/diskResource/share/DiskResourceModule.java) and should be camel case.

sri2k1us: Also found close tool button is not getting a static id plus throws a NPE 👎

ash3rz: I've found that certain elements can't get static IDs until after they've been rendered (after the show() method has been called) which might be what you're noticing?

sri2k1us: 👍

sri2k1us: Commited changes according to review.

ash3rz:
LGTM 🚢

sri2k1us:
👍

PR #26, opened 2016-10-11
CORE-8148 Add Refresh button to Apps Window

ash3rz:
Since all the other DE windows have Refresh buttons, the Apps window should probably have a Refresh button as well.

This enables the user to be able to refresh the categories, hierarchies, search results, and app listing at any time, and the user should be brought back to whatever category/hierarchy or search they had before hitting Refresh.

Given the nature of ontologies and how they can change at any time by the admins, there's also logic added to cover the scenario where the user has hit Refresh and the hierarchy they were viewing no longer exists. This can happen if the user has had the apps window open for a long time, the admin has published a new ontology that no longer has the hierarchy they were viewing, and then the user hits Refresh. In this case, the user will be brought back to the Apps Under Development category.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/commons/client/widgets/DETabPanel.java
+++ b/de-lib/src/main/java/org/iplantc/de/commons/client/widgets/DETabPanel.java
@@ -36,4 +36,9 @@ public class DETabPanel extends TabPanel {
         XElement item = findItem(getWidgetIndex(widget));
         item.setId(debugId);
     }
+
+    @Override
+    public void close(Widget item) {
+        super.close(item);
+    }
 }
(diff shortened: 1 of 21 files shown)


de-lib/src/main/java/org/iplantc/de/commons/client/widgets/DETabPanel.java
```diff
@@ -36,4 +36,9 @@ public void setTabDebugId(Widget widget, String debugId) {
         XElement item = findItem(getWidgetIndex(widget));
         item.setId(debugId);
     }
+
+    @Override
+    public void close(Widget item) {
+        super.close(item);
+    }
```
psarando: Maybe I'm missing something, but this seems redundant?

ash3rz: This method in `TabPanel` is protected.  I think the only other way to get access to this method is to have those little close buttons per tab, which I didn't like.

ash3rz:
@psarando I know you reviewed this a while back - were there any other questions or is this ok to merge?

ash3rz:
Agreed - I'd like to add Promises/Futures, but this doesn't feel like the right turning point.  Thanks for the review!

PR #27, opened 2016-10-11
CORE-8148 Refresh button w/ GwtQuery Promises

ash3rz:
This is basically a dupe of PR#26, but instead of keeping track of the multiple callbacks for the filtered hierarchies with a "parent" callback, GwtQuery's Promises are taking care of that for me.

The code is pretty much identical, so I'm not entirely sure if it's worth adding another library, but I haven't investigated other areas where we could benefit from Promises.  I haven't fully flushed out how to test with Mockito and promises, but it seems a little bit trickier so far.

Not in this PR but as a note: there are some cleaner things you can do with DOM manipulation for setting hard-to-get static IDs as well.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/hierarchies/OntologyHierarchiesPresenterImpl.java
+++ b/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/hierarchies/OntologyHierarchiesPresenterImpl.java
@@ -39,6 +39,7 @@ import com.google.gwt.event.shared.HandlerRegistration;
 import com.google.gwt.inject.client.AsyncProvider;
 import com.google.gwt.query.client.Function;
 import com.google.gwt.query.client.GQuery;
+import com.google.gwt.query.client.Promise;
 import com.google.gwt.query.client.plugins.deferred.PromiseRPC;
 import com.google.gwt.user.client.rpc.AsyncCallback;
 import com.google.gwt.user.client.ui.Widget;
@@ -109,7 +110,7 @@ public class OntologyHierarchiesPresenterImpl implements OntologyHierarchiesView
         }
     }
 
-    class FilteredHierarchyCallback extends PromiseRPC<OntologyHierarchy> {
+    public class FilteredHierarchyCallback extends PromiseRPC<OntologyHierarchy> {
         private final Tree<OntologyHierarchy, String> tree;
         private final OntologyHierarchy root;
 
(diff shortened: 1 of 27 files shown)


de-lib/src/main/java/org/iplantc/de/apps/client/presenter/hierarchies/OntologyHierarchiesPresenterImpl.java
```diff
@@ -105,6 +110,42 @@ public void onSuccess(final App appDetails) {
         }
     }
 
+    public class FilteredHierarchyCallback extends PromiseRPC<OntologyHierarchy> {
```
ash3rz: `PromiseRPC` is a helper class from GwtQuery to basically mimic the same functionality as AsyncCallback but with Promises.
super gets called at the end of these methods to mark those Promises as complete in the appropriate fashion.  If you call super first, then the `done` method gets called before these methods which would put things out of order in this case.

de-lib/src/main/java/org/iplantc/de/apps/client/presenter/hierarchies/OntologyHierarchiesPresenterImpl.java
```diff
@@ -371,4 +454,12 @@ void fireEvent(GwtEvent<?> event) {
             handlerManager.fireEvent(event);
         }
     }
+
+    List<FilteredHierarchyCallback> createFilteredHierarchyList() {
+        return Lists.newArrayList();
+    }
+
+    Promise getCallbackPromise(List<FilteredHierarchyCallback> childCallbacks) {
```
psarando: I think the GQuery methods are meant to read a little more like natural language:
`GQuery.when(promises).done(then)`
Maybe we could rename this method to follow the same pattern? Perhaps we could rename it to something like `whenCallbackPromises`?

ash3rz: Oh sure - this was a last minute change to make the tests work with GQuery so I didn't put much thought into it.

de-lib/src/main/java/org/iplantc/de/apps/client/presenter/hierarchies/OntologyHierarchiesPresenterImpl.java
ash3rz: Done - I think the idea was to keep it open like this in case something else ended up being an event handler for the dialog, but perhaps this is specific enough that we can go ahead and simplify it.  I've already pushed the changes.

ash3rz:
I'm thinking we're not currently gaining enough from promises to warrant adding another library.  Any objections to me closing this PR?  PR #26 is still open for review (without promises).

ash3rz:
Since there weren't any objections, I'll go ahead and close this PR.  The other PR without promises will remain open for review.

PR #28, opened 2016-10-17
CORE-8362 quick-fix: ignore the referer's path when it's /, to avoid redirect loop

ianmcorvidae:
I'm also changing it to use `url.getFile()` because that includes the query string. But the main part is ignoring it when the path is `/`, so we don't produce a redirect loop.


ianmcorvidae:
Thanks for the review! We can always revert the query params part if there's issues.

PR #30, opened 2016-10-21
CORE-8364 Add an error page for when the properties or bootstrap services fail

ash3rz:
This PR is for getting the user out of a neverending loading spinner in the DE when the UI fails to retrieve the properties file or if one of the bootstrap services is down.  The user will be presented with an error page that gives them the option to email support and/or find other avenues for staying up-to-date on CyVerse status.

image

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/InitializationCallbacks.java
+++ b/de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/InitializationCallbacks.java
@@ -12,6 +12,7 @@ import org.iplantc.de.commons.client.info.IplantAnnouncer;
 import org.iplantc.de.commons.client.views.dialogs.AgaveAuthPrompt;
 import org.iplantc.de.desktop.client.DesktopView;
 import org.iplantc.de.shared.DEProperties;
+import org.iplantc.de.shared.exceptions.HttpException;
 
 import com.google.gwt.core.client.GWT;
 import com.google.gwt.safehtml.shared.SafeHtmlUtils;
@@ -27,6 +28,7 @@ import com.sencha.gxt.widget.core.client.event.DialogHideEvent;
 import com.sencha.gxt.widget.core.client.event.DialogHideEvent.DialogHideHandler;
 
 import java.util.HashMap;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 
 /**
@@ -60,7 +62,12 @@ class InitializationCallbacks {
 
         @Override
         public void onFailure(Throwable caught) {
-            errorHandlerProvider.get().post(appearance.systemInitializationError(), caught);
+            LOG.log(Level.SEVERE, caught.getMessage());
+            Integer statusCode = null;
+            if (caught instanceof HttpException) {
+                statusCode = ((HttpException)caught).getStatusCode();
+            }
+            presenter.onBootstrapError(statusCode);
         }
 
         @Override
@@ -134,6 +141,7 @@ class InitializationCallbacks {
         private final UserSessionServiceFacade userSessionService;
         private final UserSettings userSettings;
         private final IplantAnnouncer announcer;
+        Logger LOG = Logger.getLogger(PropertyServiceCallback.class.getName());
 
         public PropertyServiceCallback(DEProperties deProperties,
                                        UserInfo userInfo,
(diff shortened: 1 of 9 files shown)


de-webapp/src/main/webapp/WEB-INF/jsp/bootstrapError.jsp
```diff
@@ -0,0 +1,105 @@
...
+<body>
+
+<img src="../cyverse_logo.png" height="107px" width="500px"/>
+
+    <div class="container">
+
+    <c:if test="${not empty status_code}">
+        <h1 class="redHeader">${status_code} Error</h1>
+    </c:if>
+    <h1 class="blueHeader">Oops... That wasn't supposed to happen</h1>
```
ash3rz: I'm totally open for any different ideas for the wording or if we want to remove the email now or later.

Also, if there should be a different set of links instead of the status page, newsletter, and Twitter, I can change them before submitting the ansible changes.

de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/InitializationCallbacks.java
```diff
@@ -39,7 +41,7 @@
         private final DesktopPresenterImpl presenter;
         private final UserSessionServiceFacade userSessionService;
         private final UserPreferencesCallback userPreferencesCallback;
-        Logger LOG = Logger.getLogger(BootstrapCallback.class.getName());
+        Logger LOG = Logger.getLogger("BootstrapCallback");
```
psarando: I thought we were intentionally getting the class name here, in case the name is refactored, then the logger will use the new class name?

ash3rz: Stroot and I had a talk about this - we were doing that, but then the names become obfuscated and the obfuscated name is what ends up here which isn't helpful.

ash3rz: I could switch it back and try to just prioritize the deobfuscation ticket too... so either way.

ash3rz: The more I think about it, I should just prioritize that ticket.  So yeah, I'll change it back 😄

ash3rz:
Makes sense - I'll send this on over to Kathleen and Shelley

kkennedyua:
Like the friendliness of it and seems to offer helpful suggestions. Couple of things (of course!): 

- The most immediate need is to contact Support (as you have it), but I think the reader's eyes might jump to the bullets first and miss the suggestion to contact support, so you may want to consider putting all of the suggestions in a bulleted list. Something like: "We track errors automatically, but if you want to want to learn more about what's going on and when it may be resolved, you can:
- Contact Support (linked and also spell out the email address).
- Check Ask CyVerse (add the link to the DE, not just the general link).

The status page really doesn't give any information--unless changes are in the works, so I don't think it's a viable suggestion.

Is there any suggestion for the user?--refresh the page, try again in an hour, anything? If so, might want to add that.

As for links to the newsletter or Twitter, I guess that's helpful but it's not really pertinent to trying to get back to work. 

Re the 500 error example, I'm not sure that having the error is going to help. Whenever I get an error message with an error code, I wonder if I should make a note of the error if I'm contact Support. So I guess if it's necessary to have an error, maybe add that they can note the error code they received when (or if) they contact Support. 

Hope this helps,
Kathleen

ash3rz:
Thanks Kathleen! These are all really great points.  I like the idea of switching out the status page for the Ask forum and I agree that it's easy to miss the bit about contacting support.  Removing the error code should be easy as well.  Adding options for the user to try again sounds like a good idea too.

The newsletter and Twitter links are optional.  My thought was that it might be nice to direct users to other places where they can get information about what's going on with CyVerse.  For example, when power to the building went down a couple weeks ago, if they didn't see the email, they could see the tweet about what's going on.

kkennedyua:
That makes sense, Ashley! I guess if you preface the newsletter and Twitter with "You can learn more about goings-on at CyVerse, as well as scheduled maintenance events and upcoming workshops and more, by checking our CyVerse Facebook page (link), following us on CyVerse Twitter (link), and making sure you're subscribed to our newsletter, The Node (link)." (sorry, I forget the process on whether they're automatically subscribed)

ash3rz:
Hey Kathleen, I made some additional changes based on your feedback.  Could you please take another look?  I added a new screenshot to this PR so you can see the changes.

kkennedyua:
Hi Ashley, so much better! Only suggest making it (sorry, I should have stated it more clearly) "Contact Support at support@cyverse.org" (with the link as you have it). Thank you!

ash3rz:
Updated - thanks Kathleen!

sri2k1us:
LGTM :shipit:

ash3rz:
Cool - thanks for the review!  I'll push the config changes and then merge this in.

PR #32, opened 2016-10-26
CORE-8365 Use default values when /secured/preferences is down

ash3rz:
This PR enables the UI to be more resilient when /secured/preferences is down.  Previously, the UI would not load if this callback failed.  Now, a default set of preferences will be used, the user will be informed, and the UI will continue loading.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/client/models/UserSettings.java
+++ b/de-lib/src/main/java/org/iplantc/de/client/models/UserSettings.java
@@ -6,6 +6,7 @@ import org.iplantc.de.client.models.diskResources.Folder;
 
 import com.google.gwt.core.client.GWT;
 import com.google.gwt.json.client.JSONObject;
+import com.google.gwt.json.client.JSONString;
 import com.google.web.bindery.autobean.shared.AutoBean;
 import com.google.web.bindery.autobean.shared.AutoBeanCodex;
 import com.google.web.bindery.autobean.shared.AutoBeanUtils;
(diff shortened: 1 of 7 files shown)


de-lib/src/main/java/org/iplantc/de/client/models/UserSettings.java
```diff
@@ -271,6 +272,20 @@ public Splittable asSplittable() {
         return ret;
     }
 
+    public void useDefaultValues(UserInfo instance) {
+        JSONObject defaults = new JSONObject();
+        JSONObject defaultHomeDir = new JSONObject();
+
+        JSONString id = new JSONString(instance.getHomePath());
+        JSONString path = new JSONString(instance.getHomePath());
```
psarando: Keep in mind that this `getHomePath` method can currently return `null`, if something goes wrong parsing the user info JSON in the `BootstrapCallback`.

I'm guessing that the `getHomePath` method will still be updated in another PR to return a default in this case? We will also be updating the bootstrap endpoint to return an error message instead of the user's home path when `data-info` is down.

psarando: For this PR, I think it's a good idea to use `getHomePath` here.

If the bootstrap endpoint doesn't return the user's real home path because `data-info` is down, then I think this `getHomePath` method can return `/iplant/home/<username>` for now (it can be updated in another PR).

That static part of the path (`/iplant/home/`) should be configurable if it isn't already.

ash3rz:
Thanks for the review!  I'll merge this one now and create a separate ticket and PR for the scenario when data-info is down.

PR #33, opened 2016-10-26
CORE-8400 Disable saved sessions when /secured/sessions fails and add "Try Again" btn to preferences

ash3rz:
This PR adds to the UI's resiliency (and mostly fixes a bug) where the UI will still try to save a user's session even when the /secured/sessions callback has failed.

The user will be presented with a message indicating that their saved session could not be loaded and that their current session is not being saved.  When users open the Preferences dialog, they'll have a message indicating that their session couldn't be saved and there's a button to retry hitting the /secured/sessions endpoint.

(most of the changes here are a refactor of Preferences to have a dialog + view + presenter instead of just a dialog)


ash3rz:
Thanks for the review!  I'll fix the merge conflicts then merge.

PR #35, opened 2016-10-28
CORE-8401: use the location sent in the redirect when a redirect is r…

dennisroberts71:
…eturned from an app listing request


de-lib/src/main/java/org/iplantc/de/apps/client/presenter/list/AppsListPresenterImpl.java
```diff
@@ -77,7 +77,8 @@
         @Override
         public void onFailure(Throwable caught) {
             if (caught instanceof HttpRedirectException) {
-                AgaveAuthPrompt prompt = AgaveAuthPrompt.getInstance();
+                final String uri = ((HttpRedirectException)caught).getLocation();
```
ash3rz: Just for my own knowledge - is this redirect still related to Agave?  Is it different from what gets loaded in `userInfo.getAgaveRedirect()`?

ash3rz:
LGTM 🚢

dennisroberts71:
Thanks for the reviews!

PR #36, opened 2016-10-31
Core 8358

sri2k1us:
-Add Stats panel to belphegor UI. Admins and view and sort on
  - App name
  - total no.of times used
 - total no.of  app successfully completed
 - total no.of app failed
 - Last successfully completed date
 - Last Used date
 - Rating

Also Modify Admin app search to use A new `GET /admin/apps`

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/services/impl/AppAdminServiceFacadeImpl.java
+++ b/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/services/impl/AppAdminServiceFacadeImpl.java
@@ -13,6 +13,7 @@ import org.iplantc.de.client.models.apps.App;
 import org.iplantc.de.client.models.apps.AppCategory;
 import org.iplantc.de.client.models.apps.AppDoc;
 import org.iplantc.de.client.models.apps.AppList;
+import org.iplantc.de.client.models.apps.proxy.AppListLoadResult;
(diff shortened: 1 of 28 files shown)


de-lib/src/main/java/org/iplantc/de/admin/desktop/client/services/impl/AppAdminServiceFacadeImpl.java
```diff
@@ -250,4 +261,27 @@ protected AppDoc convertFrom(String object) {
...
+        StringBuilder address = new StringBuilder(APPS_ADMIN);
+        if(!Strings.isNullOrEmpty(term)) {
+            address.append("?search=" + URL.encodeQueryString(term));
+        }
+
+        ServiceCallWrapper wrapper = new ServiceCallWrapper(GET, address.toString());
+        deService.getServiceData(wrapper,  new AsyncCallbackConverter<String, AppListLoadResult>(callback) {
+            @Override
+            protected AppListLoadResult convertFrom(String object) {
+                List<App> apps = AutoBeanCodex.decode(svcFactory, AppList.class, object).as().getApps();
```
ash3rz: I think you should be able to add `AppListLoadResult` to `AppServiceAutoBeanFactory` so you can do a direct conversion from String to `AppListLoadResult` instead of separating it into `List<Apps>` and then building `AppListLoadResult` manually.

PR #38, opened 2016-10-31
WIP CORE-8367 first pass

ash3rz:
This is a really rough first pass at one example of the UI resiliency changes.  I wanted to get some eyes on it as soon as possible to see if the direction is sane or if folks have different ideas or if things can be simplified/generalized better.

The basics are:
* There is a `DECallback` which has onSuccess(T) and onFailure(Integer statusCode, Throwable caught).
* For any call that should disable the Apps window on a 502 or 503 failure, those calls will use `AppsCallback` which extends `DECallback` (I just have one callback using it right now)
* `DECallbackWrapper` processes the status codes, logs them, and fires off a `ServiceDown` event
* When a service is down, any instances of the affected windows will have their contents replaced with a "This service is currently down" message and a "Try Again" button
* If the "Try Again" button's callback comes back successful, a `ServiceRestored` event gets fired and the affected windows will switch back to their original contents


psarando:
So far so good 👍

ash3rz:
Sweet - thanks for the review.  I wanted at least one other person to look it over before I started trickling this pattern everywhere.

ash3rz:
I'll go ahead and close this just to keep the number of open PRs clean, but anybody else feel free to review.

PR #40, opened 2016-11-03
CORE-8366 Add retry btn to notification dropdown on /secured/notifications failure

ash3rz:
This PR gives users the chance to retry fetching their recent notifications for viewing in the notification dropdown if the /secured/notifications callback fails during login.  Currently, if this callback fails, there's no other chance in the UI to get the list of recent notifications (though users could still attempt to view them via the Notification window).


ash3rz:
Thanks for the review!

PR #41, opened 2016-11-03
CORE-8358 add path to columnconfig so that sort direction arrow displa…

sri2k1us:
…yed in column header.


sri2k1us:
Merging branch to master. This change does not affect any logic.

PR #42, opened 2016-11-03
CORE-8420 Construct user home dir from configs if data-info is down

ash3rz:
If the UI is unable to get a user's preferences, a default set of user preferences is used.  One of those default settings is using the user's home directory, retrieved from the bootstrap endpoint, as the default output folder for analyses.  If the data-info service is down while the UI is calling bootstrap, the path to the user's home directory stored in `UserInfo` will be empty.  

This PR attempts to use the irods home path from the configs plus the user's username to construct their home directory path as a default value in the case that data-info is down.  If this effort fails, then no default output folder is set.  The user will have to pick some directory they have write access to at app launch or change their preferences.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/client/models/UserSettings.java
+++ b/de-lib/src/main/java/org/iplantc/de/client/models/UserSettings.java
@@ -3,9 +3,7 @@ package org.iplantc.de.client.models;
 import org.iplantc.de.client.KeyBoardShortcutConstants;
 import org.iplantc.de.client.models.diskResources.DiskResourceAutoBeanFactory;
 import org.iplantc.de.client.models.diskResources.Folder;
-import org.iplantc.de.shared.DEProperties;
 
-import com.google.common.base.Strings;
 import com.google.gwt.core.client.GWT;
 import com.google.gwt.json.client.JSONObject;
 import com.google.gwt.json.client.JSONString;
@@ -278,7 +276,7 @@ public class UserSettings {
         JSONObject defaults = new JSONObject();
         JSONObject defaultHomeDir = new JSONObject();
 
-        String homeDir = getHomeDir(userInfo);
+        String homeDir = userInfo.getHomePath();
 
         JSONString id = new JSONString(homeDir);
         JSONString path = new JSONString(homeDir);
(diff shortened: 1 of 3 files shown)


de-lib/src/main/java/org/iplantc/de/client/models/UserSettings.java
```diff
@@ -286,6 +290,23 @@ public void useDefaultValues(UserInfo instance) {
...
+        }
+
+        String irodsHome = DEProperties.getInstance().getIrodsHomePath();
+        String username = userInfo.getUsername();
+
+        if (Strings.isNullOrEmpty(irodsHome) || Strings.isNullOrEmpty(username)) {
+            return "";
+        }
+
+        return irodsHome + "/" + username;
```
psarando: I'm thinking this logic should be moved into `UserInfo#getHomePath` itself.

Notice how many other places this method is called besides the `DEFAULT_OUTPUT_FOLDER` default setting, so those parts of the DE will need this default path as well if the bootstrap endpoint couldn't return the real home path.

Same goes for `getTrashPath` and `getBaseTrashPath`. Then when the work for CORE-8367 is done for the data-info serivce, we'll need a way to fetch these real paths once the service comes back up (I'm thinking of adding a "data-info bootstrap" endpoint to do this for CORE-7813).

ash3rz: That's a really good point - I'll refactor.

de-lib/src/main/java/org/iplantc/de/client/models/UserSettings.java
```diff
@@ -272,12 +274,14 @@ public Splittable asSplittable() {
         return ret;
     }
 
-    public void useDefaultValues(UserInfo instance) {
+    public void useDefaultValues(UserInfo userInfo) {
         JSONObject defaults = new JSONObject();
```
psarando: I noticed that too, and that `Splittable` is already used a lot in `UserSettings`, but why not use an `AutoBean` instead?

This seems like something for another PR.

ash3rz: Sure a refactor seems valid.  I'll create a ticket for this.

ash3rz:
Thanks for the review!

PR #43, opened 2016-11-09
CORE-8400 Improvements to saved sessions resiliency

ash3rz:
This PR should fix:
- Saved Session checkbox getting unchecked when the saved sessions service comes back up
- When the saved sessions service comes back up, the Try Again button and error message in the Preferences window are removed
- If saved sessions goes down while the user is logged in, the periodic session save will continue to attempt to save their session for a certain number of retries (currently 8 retries is 2 minutes) before disabling itself and letting the user know their session was not saved


ash3rz:
Thanks for the review!

PR #44, opened 2016-11-10
CORE-8432 Refactor UserSettings to AutoBean

ash3rz:
This PR helps the UI to be more consistent by moving away from using JSONObjects and Splittables to AutoBeans for user settings.  The singleton `UserSettings` is still maintained to facilitate setting default values.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/client/models/UserSettings.java
+++ b/de-lib/src/main/java/org/iplantc/de/client/models/UserSettings.java
@@ -25,6 +25,7 @@ public class UserSettings {
     private UserSetting userSetting;
     private static UserSettings instance;
     private boolean userSessionConnection;
+    private static String ANALYSES_DIR = "analyses";
 
     public UserSettings(final UserSetting userSetting) {
         setUserSettings(userSetting);
@@ -77,7 +78,8 @@ public class UserSettings {
 
         if (userSetting.getDefaultOutputFolder() == null) {
             String homePath = userInfo.getHomePath();
-            setDefaultOutputFolder(buildFolder(homePath));
+            String outputDir = homePath + "/" + ANALYSES_DIR;
+            setDefaultOutputFolder(buildFolder(outputDir));
         }
 
         if (userSetting.getSystemDefaultOutputDir() == null) {
--- a/de-lib/src/main/java/org/iplantc/de/client/models/UserSettings.java
+++ b/de-lib/src/main/java/org/iplantc/de/client/models/UserSettings.java
@@ -3,68 +3,35 @@ package org.iplantc.de.client.models;
 import org.iplantc.de.client.KeyBoardShortcutConstants;
 import org.iplantc.de.client.models.diskResources.DiskResourceAutoBeanFactory;
 import org.iplantc.de.client.models.diskResources.Folder;
+import org.iplantc.de.client.models.userSettings.UserSetting;
+import org.iplantc.de.client.models.userSettings.UserSettingAutoBeanFactory;
 
+import com.google.common.base.Strings;
 import com.google.gwt.core.client.GWT;
-import com.google.gwt.json.client.JSONObject;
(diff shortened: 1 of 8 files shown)


de-lib/src/main/java/org/iplantc/de/client/models/userSettings/UserSetting.java
```diff
@@ -0,0 +1,53 @@
...
+public interface UserSetting {
+
+    Boolean isEnableAnalysisEmailNotification();
+    void setEnableAnalysisEmailNotification(Boolean enable);
+
+    Boolean isEnableImportEmailNotification();
+    void setEnableImportEmailNotification(Boolean enable);
+
+    String getDefaultFileSelectorPath();
+	void setDefaultFileSelectorPath(String value);
```
psarando: I like how some of these getters and setters are aligned with each other, but this one and a few others are a little off (the 2 above are not aligned this way either, but they are still indented normally).

de-lib/src/main/java/org/iplantc/de/client/models/userSettings/UserSetting.java
```diff
@@ -0,0 +1,53 @@
...
+	void setAnalysisKBShortcut(String value);
+
+    String getNotificationKBShortcut();
+	void setNotificationKBShortcut(String value);
+
+    String getCloseKBShortcut();
+	void setCloseKBShortcut(String value);
+
+    Folder getSystemDefaultOutputDir();
+	void setSystemDefaultOutputDir(Folder value);
```
psarando: These default output setting keys were renamed by cyverse-de/terrain#8 to `default_output_folder` and `system_default_output_dir`. I think these fields will need `@Property` annotations for the new names?
If there are objections to the new name, these can be renamed back to the camel-case versions in terrain.

de-lib/src/main/java/org/iplantc/de/client/models/UserSettings.java
```diff
@@ -76,326 +42,252 @@ public static UserSettings getInstance() {
...
-        } else {
+        if (userSetting.isSaveSession() == null) {
             setSaveSession(true);
         }
 
-        if (split.get(DEFAULT_OUTPUT_FOLDER) != null) {
-            setDefaultOutputFolder(buildFolder(split.get(DEFAULT_OUTPUT_FOLDER)));
+        if (userSetting.getDefaultOutputFolder() == null) {
+            String homePath = userInfo.getHomePath();
+            setDefaultOutputFolder(buildFolder(homePath));
```
psarando: I'm not sure about using just the home folder as the default output folder. I think this will just dump the `log` and file outputs into the user's home folder. I think it would be safer to use `homePath + "/analyses"`.

ash3rz:
Thanks for the review!  I'll update the default output folder to include `"/analyses"` and probably wait to merge this till after PR #47 so it'll have all the new bootstrap changes.

de-lib/src/main/java/org/iplantc/de/client/models/UserSettings.java
```diff
@@ -3,68 +3,36 @@
...
-
+    private final DiskResourceAutoBeanFactory diskResourceFactory =
+            GWT.create(DiskResourceAutoBeanFactory.class);
+    private final UserSettingAutoBeanFactory factory = GWT.create(UserSettingAutoBeanFactory.class);
+    private UserInfo userInfo;
 
+    private UserSetting userSetting;
     private static UserSettings instance;
+    private boolean userSessionConnection;
+    private static String ANALYSES_DIR = "analyses";
```
psarando: This should probably come from the configs, but that doesn't need to happen in this PR, so we can get this merged this for the next QA drop 😁

ash3rz: Ah, good point - will make a note to add it to a config after the drop 😄

ash3rz:
Thanks for the review!

PR #46, opened 2016-11-16
CORE-8434 Clear Apps grid when HPC app load fails

ash3rz:
If HPC apps fail to load, the Apps grid will mistakenly appear as if HPC apps did load because apps are listed.  These apps are actually the same apps listed from whatever previous category/hierarchy the user had selected.  Now, if HPC is down, the grid will clear and the grid's heading will inform the user of the app load error.

Also, discovered that when the user hits Refresh, duplicate handlers were getting added to the view - this is now fixed.


ash3rz:
Thanks for the review!

PR #47, opened 2016-11-21
CORE-8431 Update UI to accept bootstrap changes

ash3rz:


de-lib/src/main/java/org/iplantc/de/client/models/bootstrap/WorkspaceError.java
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * @author aramsey
+ */
+public interface WorkspaceError {
+
+    @PropertyName("error_code")
+    String getErrorCode();
+
+    //TODO more fields here?
```
psarando: We can't assume that a `session.error` will always be a `String` or that a `workspace.error` will always be a JSON object with an `error_code`. I mentioned something similar in https://github.com/cyverse-de/terrain/pull/8#issuecomment-261057568.

The errors in the cyverse-de/terrain#8 description were examples of the different kinds of errors that can show up from our services, so any of those errors can show up in the `session`, `workspace`, `data_info`, or `preferences` error fields. This is why the UI has the [org.iplantc.de.commons.client.ErrorHandler#parseExceptionJson method](https://github.com/cyverse-de/ui/blob/2.7.0-RELEASE/de-lib/src/main/java/org/iplantc/de/commons/client/ErrorHandler.java#L154-L174).

I'm not sure how to handle this variability here, but maybe all these AutoBeans should use `String` as the `error` field type, then we use that `parseExceptionJson` method if we need to parse out more info?

ash3rz:
Ah, sorry, I know you mentioned that but it didn't quite get that all the services could have that variability.  I'll refactor - thanks for catching!

ash3rz:
Thanks for the review and catching all the corner cases and whatnot!

PR #49, opened 2016-11-28
CORE-8367 UI Resiliency - Apps Window

ash3rz:
This PR refactors the UI such that it has access to the status codes returned from the apps service which affect the base Apps window.  This is the fully flushed out version of PR #38.  The ServiceDownPanel could still use some work, since it's pretty boring looking, but it'll do for now.

(Copied from previous PR)
The basics:
* There is a `DECallback` which has onSuccess(T) and onFailure(Integer statusCode, Throwable caught).
* For any call that should disable the Apps window on a 502 or 503 failure, those calls will use `AppsCallback` which extends `DECallback` (I just have one callback using it right now)
* `DECallbackWrapper` processes the status codes, logs them, and fires off a `ServiceDown` event
* When a service is down, any instances of the affected windows will have their contents replaced with a "This service is currently down" message and a "Try Again" button
* If the "Try Again" button's callback comes back successful, a `ServiceRestored` event gets fired and the affected windows will switch back to their original contents


ianmcorvidae:
As best I can understand it, this looks good (getting a bit of eyes glazing over at the barrage of function signature changes that don't change functionality, so I could have missed something)! But I mostly looked at it to learn a bit about the UI; I'm sure someone with a better handle on it should also review.

ash3rz:
Heh, yeah, a lot changing `AsyncCallback` to `DECallback` which makes my eyes glaze over too. Thanks for taking a look over it!

de-lib/src/main/java/org/iplantc/de/client/services/converters/AppTemplateDECallbackConverter.java
```diff
@@ -0,0 +1,165 @@
...
+import com.google.web.bindery.autobean.shared.AutoBeanUtils;
+import com.google.web.bindery.autobean.shared.Splittable;
+import com.google.web.bindery.autobean.shared.impl.StringQuoter;
+
+import java.util.List;
+
+/**
+ * @author jstroot
+ */
+public class AppTemplateDECallbackConverter extends DECallbackConverter<String, AppTemplate> {
```
psarando: It looks like this class is replacing `AppTemplateCallbackConverter` in most places, but not everywhere?  (see `AppTemplateServicesImpl`)

As far as I can tell, this class is the same as `AppTemplateCallbackConverter`, just renamed and it now extends `DECallbackConverter`?

ash3rz: Yes, you're right.  Now that I'm looking at it, I think I might have missed a few spots, but the reason why I created a copy was not all the calls were going to a `/apps/*` endpoint and I was meaning to ask about those.  One of the calls goes to `/preview-template` for example and I wasn't sure if that was still going to the apps service.  If not, then I don't think it makes sense to disable the entire App window's functionality for that.

psarando: I'm not finding that terrain `/preview-template` endpoint.
We have [POST /apps/arg-preview](https://cyverse-de.github.io/api/endpoints/app-metadata.html#previewing-command-line-arguments) that is an apps service endpoint.

psarando: Ah, good catch. And yes, `/analyses` endpoints are forwarded to the apps service (see the docs for [GET /analyses](https://cyverse-de.github.io/api/endpoints/app-execution.html#listing-jobs) for example).
When in doubt, you can check the docs to see if the endpoint is a passthrough to the apps service: https://cyverse-de.github.io/api/endpoint-index.html

ash3rz:
I added some changes based on PR comments, but I also decided to go ahead and change `AppsCallback` to handle not only the Apps window, but also App Editor, App Launch, and Pipeline builder.  I think this makes the most sense so if one window discovers that the Apps service has gone down or has come back up, the other windows will be disabled/enabled appropriately.  

I think I'll keep the Analysis window separate with its own `AnalysisCallback` (in a different PR for a separate ticket) for now though.

ash3rz:
Cool - thanks for the review!

PR #50, opened 2016-11-30
CORE-8443 Improve UI's handling of individual bootstrap service errors

ash3rz:
* If the `user_info` key is missing the username, name, or email information, the user will now be redirected to the DE's error page so they can try logging in again.
* If the `session` key contains an error, the loginTime will not be populated (which is acceptable) and the UI will recheck for any redirects from the oauth service when the Apps window is opened (for example, the Agave redirect).
* If the `workspace` key contains an error, the UI will by default assume the user is a new user and show the prompt for the introduction/tutorial upon login.  The workspace ID isn't used in the UI, so that is ignored.
* If the `data_info` key contains an error, default home and trash paths will be used.  If the user opens the Data window and gets a successful callback from `/filesystem/root` the home and trash paths from that callback will be used instead.
* If the `preferences` key contains an error, default preferences (specifically their job output folder) will be used.

(requires ansible and terrain changes as well)

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/categories/AppCategoriesPresenterImpl.java
+++ b/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/categories/AppCategoriesPresenterImpl.java
@@ -16,15 +16,19 @@ import org.iplantc.de.apps.client.views.details.dialogs.AppDetailsDialog;
 import org.iplantc.de.apps.shared.AppsModule;
 import org.iplantc.de.client.events.EventBus;
 import org.iplantc.de.client.models.HasId;
+import org.iplantc.de.client.models.UserInfo;
 import org.iplantc.de.client.models.apps.App;
 import org.iplantc.de.client.models.apps.AppCategory;
 import org.iplantc.de.client.models.apps.integration.AppTemplate;
 import org.iplantc.de.client.services.AppServiceFacade;
 import org.iplantc.de.client.services.AppUserServiceFacade;
+import org.iplantc.de.client.services.OauthServiceFacade;
 import org.iplantc.de.client.util.JsonUtil;
 import org.iplantc.de.commons.client.ErrorHandler;
+import org.iplantc.de.commons.client.info.ErrorAnnouncementConfig;
 import org.iplantc.de.commons.client.info.IplantAnnouncer;
 import org.iplantc.de.commons.client.info.SuccessAnnouncementConfig;
+import org.iplantc.de.commons.client.views.dialogs.AgaveAuthPrompt;
 import org.iplantc.de.commons.client.widgets.DETabPanel;
 import org.iplantc.de.shared.AppsCallback;
 import org.iplantc.de.shared.AsyncProviderWrapper;
@@ -55,6 +59,7 @@ import com.sencha.gxt.widget.core.client.tree.Tree;
 import java.util.Arrays;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Map;
 
 /**
  * @author jstroot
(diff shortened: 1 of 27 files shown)


de-lib/src/main/java/org/iplantc/de/client/models/UserInfo.java
```diff
@@ -249,8 +253,19 @@ public void setSavedOrderedWindowStates(List<WindowState> savedOrderedWindowStat
         this.savedOrderedWindowStates = savedOrderedWindowStates;
     }
 
+    public void setAuthRedirects(Map<String, String> redirects) {
+        if (session == null) {
+            Session newSession = factory.session().as();
+            newSession.setAuthRedirects(redirects);
+            session = newSession;
+        } else {
+            session.setAuthRedirects(redirects);
```
psarando: Can `session.setAuthRedirects(redirects)` be moved after the if/else block?

de-lib/src/main/java/org/iplantc/de/apps/client/presenter/categories/AppCategoriesPresenterImpl.java
```diff
@@ -189,6 +196,30 @@ public void onSuccess(List<AppCategory> result) {
 
     }
 
+    @Override
+    public void checkForAgaveRedirect() {
+        if (!userInfo.hasSessionError()) {
+            return;
+        }
+
+        oauthServiceFacade.getRedirectUris(new AsyncCallback<Map<String, String>>() {
```
psarando: It looks like this `checkForAgaveRedirect` will only display the `AgaveAuthPrompt` if the `bootstrap` endpoint had an error checking for this redirect URI at login.

The `DEAppsWindow#afterShow` method used to show this prompt if `userInfo` had an Agave redirect URI at login from the `bootstrap` response (now this method just calls `checkForAgaveRedirect`, and it doesn't look like that's called from anywhere else).

If `userInfo.hasSessionError()` is `false` above, then we should probably still display the prompt if `userInfo` has an Agave redirect URI from `bootstrap`.

de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/navigation/proxy/FolderRpcProxyImpl.java
```diff
@@ -65,15 +66,17 @@ public void onSuccess(List<DiskResourceQueryTemplate> result) {
         final IsMaskable maskable;
         private final NavigationView.Presenter.Appearance appearance;
         final SearchServiceFacade searchSvc;
+        UserInfo userInfo;
 
         public RootFolderCallback(final SearchServiceFacade searchService,
                                   final AsyncCallback<List<Folder>> callback,
-                                  final IplantAnnouncer announcer,
+                                  final IplantAnnouncer announcer, UserInfo userInfo,
```
psarando: All other params in this constructor are on their own line, so this `UserInfo` param should also be on its own line.

de-lib/src/main/java/org/iplantc/de/apps/client/presenter/categories/AppCategoriesPresenterImpl.java
ash3rz: Ah, good catch.  I'll fix.

psarando:
I have one more suggestion for these changes (not a priority, so maybe for another PR?):

If the preferences can't be loaded at login (and we use defaults for the user at that time), could we try to reload their preferences later when the user opens the preferences dialog? That way the user can continue to work after login, but if the preferences service comes back up later, then if they open the dialog to change/save a setting, their original preferences will be loaded and they won't have to re-customize all their settings?

This isn't a major issue, since the only setting that may require resetting is their output folder, and all other settings can be loaded from the service by refreshing or logging in again :) But if we could reload preferences when the dialog opens, it would be a convenience to the user (especially if we add more settings in the future).

psarando:
These changes look good! 👍 
My low priority suggestion shouldn't hold up this PR from getting merged.

ash3rz:
That's a really good idea and I think it's part of the goal of the ticket (using defaults while providing opportunities to get the user's original values/settings), so I think it's worth putting into this PR before merging for completeness. 👍 
I'll get started on those changes shortly.

ash3rz:
Also, now that I think about it - since bootstrap is returning all of the user's preferences and not just the output directory, is there any reason to keep the separate, secondary call to fetch preferences during initialization?  Probably not, so I'll refactor that as well.

psarando:
The `Preferences` changes look good! 👍 
:shipit:

ash3rz:
Sweet - thanks for the review!

PR #52, opened 2016-12-05
CORE-8424 Remove text field validation. Instead add validation when 

sri2k1us:
`save` is clicked.


de-lib/src/main/java/org/iplantc/de/diskResource/client/views/metadata/DiskResourceMetadataViewImpl.java
```diff
@@ -462,8 +457,14 @@ private void initUserMdGridEditor() {
 
             @Override
             public void onCompleteEdit(CompleteEditEvent<Avu> event) {
-                dirty = true;
-                userMdListStore.commitChanges();
+              dirty = true;
```
ash3rz: Not important, but I think the spacing got screwed up here.

PR #53, opened 2016-12-05
CORE-8460 UI Resiliency - Data Window

ash3rz:
This PR builds on top of PR #49 but with changes for the Data window.

Really the only tricky thing with this PR is some of the endpoints in the Disk Resource service facades go from terrain to the metadata service, not data-info.  I tried to check them all to make sure only the data-info endpoints got updated (don't want to disable the entire Data window if only the metadata service is down).

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/client/services/DiskResourceServiceFacade.java
+++ b/de-lib/src/main/java/org/iplantc/de/client/services/DiskResourceServiceFacade.java
@@ -1,7 +1,5 @@
 package org.iplantc.de.client.services;
 
-import java.util.List;
-
 import org.iplantc.de.client.models.HasPaths;
 import org.iplantc.de.client.models.dataLink.DataLink;
 import org.iplantc.de.client.models.diskResources.DiskResource;
@@ -16,12 +14,16 @@ import org.iplantc.de.client.models.diskResources.RootFolders;
 import org.iplantc.de.client.models.diskResources.TYPE;
 import org.iplantc.de.client.models.services.DiskResourceMove;
 import org.iplantc.de.client.models.viewer.InfoType;
+import org.iplantc.de.shared.DECallback;
 
 import com.google.gwt.json.client.JSONObject;
 import com.google.gwt.user.client.rpc.AsyncCallback;
 import com.google.web.bindery.autobean.shared.Splittable;
+
 import com.sencha.gxt.core.shared.FastMap;
 import com.sencha.gxt.data.shared.loader.FilterPagingLoadConfigBean;
+
+import java.util.List;
 
 /**
  * @author jstroot
@@ -36,7 +38,7 @@ public interface DiskResourceServiceFacade {
      * @param folder The parent of the subfolders to refresh from the service.
      * @param callback executed when RPC call completes.
      */
-    void refreshFolder(Folder folder, final AsyncCallback<List<Folder>> callback);
+    void refreshFolder(Folder folder, final DECallback<List<Folder>> callback);
 
     DiskResource combineDiskResources(DiskResource from, DiskResource into);
 
@@ -47,7 +49,7 @@ public interface DiskResourceServiceFacade {
      * 
      * @param callback executed when RPC call completes.
      */
-    void getRootFolders(AsyncCallback<RootFolders> callback);
+    void getRootFolders(DECallback<RootFolders> callback);
 
     /**
      * Called to retrieve the entire contents of a folder.
@@ -64,7 +66,7 @@ public interface DiskResourceServiceFacade {
                            List<InfoType> infoTypeFilterList,
                            TYPE entityType,
                            FilterPagingLoadConfigBean loadConfig,
-                           AsyncCallback<Folder> callback);
+                           DECallback<Folder> callback);
 
     /**
      * Called to retrieve the contents of a folder without its file contents.
@@ -72,7 +74,7 @@ public interface DiskResourceServiceFacade {
      * @param parent requested folder.
      * @param callback executed when RPC call completes.
      */
-    void getSubFolders(final Folder parent, final AsyncCallback<List<Folder>> callback);
+    void getSubFolders(final Folder parent, final DECallback<List<Folder>> callback);
 
     /**
      * Call service to create a new folder
@@ -80,7 +82,7 @@ public interface DiskResourceServiceFacade {
      * @param parentFolder parent folder where the new folder will be created
      * @param callback executed when RPC call completes.
      */
-    void createFolder(Folder parentFolder, final String newFolderName, AsyncCallback<Folder> callback);
+    void createFolder(Folder parentFolder, final String newFolderName, DECallback<Folder> callback);
 
     /**
      * Check if a list of files or folders exist.
@@ -89,7 +91,7 @@ public interface DiskResourceServiceFacade {
      * @param callback callback executed when RPC call completes. On success, a map that maps resource
      *            paths to whether or not they exist.
      */
-    void diskResourcesExist(HasPaths diskResourcePaths, AsyncCallback<DiskResourceExistMap> callback);
+    void diskResourcesExist(HasPaths diskResourcePaths, DECallback<DiskResourceExistMap> callback);
 
     /**
      * Calls the move folder and move file services for the list of given disk resource ids.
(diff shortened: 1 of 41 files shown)


de-lib/src/main/java/org/iplantc/de/client/services/DiskResourceServiceFacade.java
```diff
@@ -36,7 +38,7 @@
      * @param folder The parent of the subfolders to refresh from the service.
      * @param callback executed when RPC call completes.
      */
-    void refreshFolder(Folder folder, final AsyncCallback<List<Folder>> callback);
+    void refreshFolder(Folder folder, final DECallback<List<Folder>> callback);
```
ianmcorvidae: Why is it these ones are `DECallback` rather than `DataCallback`? Couldn't quite tell what the difference was supposed to be (mostly looking through the PR to see if there's any tricky multi-service-endpoint edge cases I could rustle up)

de-lib/src/main/java/org/iplantc/de/client/services/DiskResourceServiceFacade.java
```diff
@@ -129,7 +131,7 @@ void moveContents(final Folder sourceFolder,
      * @param dest id of the destination folder.
      * @param callback service success/failure callback
      */
-    void importFromUrl(String url, DiskResource dest, AsyncCallback<String> callback);
+    void importFromUrl(String url, DiskResource dest, DECallback<String> callback);
```
ianmcorvidae: This one is + may turn into even more of a weird case; URL import right now goes from terrain to apps and doesn't use data-info. But, the plan is to have it go through terrain into data-info, which then calls out to apps (CORE-6214). I'm not sure how things go down in these cases.

de-lib/src/main/java/org/iplantc/de/client/services/DiskResourceServiceFacade.java
ash3rz: Mostly just to keep the service facades generic.  There probably won't be any cross-over with the disk resource / data window related callbacks, but for example, there could be a case where the Analysis window and the Apps window share the same callback and I want the service to be generic enough to handle either an `AppsCallback` or `AnalysisCallback` (in another PR coming up).

de-lib/src/main/java/org/iplantc/de/client/services/DiskResourceServiceFacade.java
```diff
@@ -144,29 +146,28 @@ void moveContents(final Folder sourceFolder,
...
-    void deleteDiskResources(HasPaths diskResources, AsyncCallback<HasPaths> callback);
+    void deleteDiskResources(HasPaths diskResources, DECallback<HasPaths> callback);
 
     /**
      * @param resource the <code>DiskResource</code> for which metadata will be retrieved.
      * @param callback callback executed when service call completes.
      */
-    void getDiskResourceMetaData(DiskResource resource,
-                                 AsyncCallback<DiskResourceMetadataList> callback);
+    void getDiskResourceMetaData(DiskResource resource, DECallback<DiskResourceMetadataList> callback);
```
ianmcorvidae: This and the set-metadata endpoint both go through terrain to data-info and then to metadata, which it sounds like you're aware of (but in any case, it's those two particularly)

de-lib/src/main/java/org/iplantc/de/client/services/DiskResourceServiceFacade.java
ianmcorvidae: Ah, I see. Since this is just a function parameter list it's better to use the more generic callback. With you now!

de-lib/src/main/java/org/iplantc/de/client/services/DiskResourceServiceFacade.java
```diff
@@ -311,10 +302,10 @@ void setDiskResourceMetaData(DiskResource resource,
      * @param paths destination DR's path to which metadata will be copied.
      * @param callback callback object
      */
-            void
-            copyMetadata(final String srcUUID,
-                         final Splittable paths,
-                         final AsyncCallback<String> callback);
+
+    void copyMetadata(final String srcUUID,
```
ianmcorvidae: (copy and save metadata both use the metadata service too, of course)

de-lib/src/main/java/org/iplantc/de/client/services/DiskResourceServiceFacade.java
ash3rz: Good catch - I'll update the presenter that calls this endpoint to `AppsCallback` since that makes the most sense right now.  We can switch it to `DataCallback` when CORE-6214 goes through.

de-lib/src/main/java/org/iplantc/de/client/services/DiskResourceServiceFacade.java
ash3rz: Yeah, I might need to do some more testing to make sure that if the metadata service is down it doesn't disable the Data window in some of these endpoints.  Hopefully that won't be the case.

psarando: I'm not sure if the UI can tell which service is down if it gets a 503 from this call.
In my previous comment I suggest that maybe the UI should assume data-info is down if this call returns a 503, and if it's just metadata, then the `retry` button will re-enable the data window right away?

de-lib/src/main/java/org/iplantc/de/client/services/DiskResourceServiceFacade.java
ianmcorvidae: That's a good idea; we should be able to make the nginx return that automatically if we add a JSON body to the 502/503/504 response, and then we just need to make sure our error-handling code passes that info along. That'll probably need more investigation, and it's possible some of these endpoints should more gracefully respond when stuff is down anyway (e.g. the manifest endpoint)

de-lib/src/main/java/org/iplantc/de/client/services/FileEditorServiceFacade.java
```diff
@@ -17,7 +18,7 @@
      *  @param file desired manifest's file ID (path).
      * @param callback executes when RPC call is complete.
      */
-    void getManifest(File file, AsyncCallback<Manifest> callback);
+    void getManifest(File file, DECallback<Manifest> callback);
```
ianmcorvidae: Here's another one to test an edge case: the manifest is through data-info, but it uses the tree-urls service in the background. Most likely we should make a fallback in data-info that causes this to not throw the 503 the whole way upward (and tree-urls is rarely down, it's so tiny), but for now it's something to check.

de-lib/src/main/java/org/iplantc/de/client/services/DiskResourceServiceFacade.java
ash3rz: Hm... right now the UI is assuming there's a link between windows and services.  So the presenters decide which type of callback, in this case `DataCallback`, to use.  If that callback fails, then DECallback knows which windows to disable.  So if we have a callback that can return a 502/503 but it shouldn't disable the Data window, then it shouldn't be a `DataCallback` type. 

The `retry` button will only re-enable a window if the call that initially failed begins to succeed, because it only resends that first failing call.

de-lib/src/main/java/org/iplantc/de/client/services/FileEditorServiceFacade.java
ash3rz: From what I can tell, if I enable maintenance on tree-urls, I can still view a file (which calls the manifest endpoint) and the Data window stays enabled.  So we're good here 👍

de-lib/src/main/java/org/iplantc/de/client/services/DiskResourceServiceFacade.java
ash3rz: So it seems like getting metadata currently won't disable the Data window, but trying to copy/save metadata will. 😖

ash3rz: (and you get a ton of 503 error pop ups from the Data window's Details panel, but that's a different story)

ash3rz: I think for now, I'll switch the get and set metadata endpoints back to the old `AsyncCallback` so they can't disable the Data window.

ash3rz:
Thanks for the review!

PR #54, opened 2016-12-06
Update ServiceDownPanel appearance

ash3rz:
This PR is just to update the wording and appearance of the panel that gets swapped into the Apps/Data/Analysis window when the UI discovers that the backend services those windows require are down. 

image


ash3rz:
I got the green light from Kathleen via Slack on the wording for this PR, so I'll go ahead and merge since that was really what I wanted feedback on.  Anybody else can feel free to comment though as well.

PR #56, opened 2016-12-09
Update the UI to include the system ID in several service endpoint requests.

dennisroberts71:
This is a fairly significant change, and I wouldn't be surprised if there were issues. A thorough review would be appreciated.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/client/models/pipelines/PipelineTask.java
+++ b/de-lib/src/main/java/org/iplantc/de/client/models/pipelines/PipelineTask.java
@@ -1,5 +1,7 @@
 package org.iplantc.de.client.models.pipelines;
 
+import org.iplantc.de.client.models.HasSystemId;
+
 import com.google.gwt.user.client.ui.HasName;
 import com.google.web.bindery.autobean.shared.AutoBean.PropertyName;
 
@@ -11,13 +13,7 @@ import java.util.List;
  * @author psarando
  *
  */
-public interface PipelineTask extends HasName {
-
-    @PropertyName("system_id")
-    String getSystemId();
-
-    @PropertyName("system_id")
-    void setSystemId(String systemId);
+public interface PipelineTask extends HasName, HasSystemId {
 
     @PropertyName("id")
     String getTaskId();
--- a/de-lib/src/main/java/org/iplantc/de/client/models/pipelines/PipelineTask.java
+++ b/de-lib/src/main/java/org/iplantc/de/client/models/pipelines/PipelineTask.java
@@ -13,35 +13,42 @@ import java.util.List;
  */
 public interface PipelineTask extends HasName {
 
-    @PropertyName("id")
-    public String getTaskId();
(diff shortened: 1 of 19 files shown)


de-lib/src/main/java/org/iplantc/de/client/models/pipelines/PipelineTask.java
```diff
@@ -13,35 +13,42 @@
  */
 public interface PipelineTask extends HasName {
 
+    @PropertyName("system_id")
```
ash3rz: I think we can remove these two additions and extend `HasSystemId` instead here like you did in `JobExecution.java`

de-lib/src/main/java/org/iplantc/de/client/models/pipelines/ServicePipelineStep.java
```diff
@@ -13,16 +13,22 @@
  */
 public interface ServicePipelineStep extends HasName, HasDescription {
 
+    @PropertyName("system_id")
```
ash3rz: Same here, I think we can just extend `HasSystemId`

de-lib/src/main/java/org/iplantc/de/client/services/AppTemplateServices.java
```diff
@@ -34,9 +34,17 @@
...
-    void getAppTemplateForEdit(HasId appId, DECallback<AppTemplate> callback);
+    void getAppTemplateForEdit(HasQualifiedId appId, DECallback<AppTemplate> callback);
     
     /**
+     * Retrieves a UI preview of the given <code>AppTemplate</code>.
+     * 
+     * @param at the <code>AppTemplate</code> for which the preview should be produced.
+     * @param callback
+     */
+    void getAppTemplatePreview(AppTemplate at, DECallback<AppTemplate> callback);
```
ash3rz: Minor thing, but I recently removed this and its counterpart in `AppTemplateServicesImpl` as it was dead code.  It might've accidentally gotten added back in a rebase.

de-lib/src/main/java/org/iplantc/de/pipelineBuilder/client/json/App.java
```diff
@@ -24,6 +24,7 @@
     private List<PipelineAppData> outputAppData;
 
 	private int publicId;
+	private String systemId;
 	private String ID;
     private String templateId;
```
ash3rz: Minor thing that's probably been there for a while, but looks like the spacing is a little wacky in this file?

dennisroberts71: The invading spaces are at it again! 👾  Thanks for catching this.

dennisroberts71: Looks like several lines were indented with tabs. I reformatted the code using IntelliJ Idea. That should fix it.

PR #57, opened 2016-12-12
CORE-8490 UI Resiliency - App Launch Window

ash3rz:
This turned out to be a bit of a refactor since we were making service calls directly from `AppLaunchWindow` that I moved to `AppLaunchPresenterImpl`.  Otherwise, the goal of this PR was to refactor any calls related to the App Launch window to use `AppLaunchCallback` instead of `AppsCallback` since it is possible for users to be able to launch apps even if the metadata service is down.  (Currently the metadata service being down is capable of causing most `AppsCallbacks` to fail which would disable the Apps, App Editor, and Workflow builder windows.)


ash3rz:
Cool, thanks for the review!

PR #58, opened 2016-12-15
CORE-8481 UI Resiliency - Notification Window

ash3rz:
This PR builds on top of PR #49 but with changes for the Notification window.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/client/services/impl/MessageServiceFacadeImpl.java
+++ b/de-lib/src/main/java/org/iplantc/de/client/services/impl/MessageServiceFacadeImpl.java
@@ -4,10 +4,6 @@ import static org.iplantc.de.shared.services.BaseServiceCallWrapper.Type.DELETE;
 import static org.iplantc.de.shared.services.BaseServiceCallWrapper.Type.GET;
 import static org.iplantc.de.shared.services.BaseServiceCallWrapper.Type.POST;
 
-import java.util.Collections;
-import java.util.List;
-
-import org.iplantc.de.shared.DEProperties;
 import org.iplantc.de.client.models.HasId;
 import org.iplantc.de.client.models.UserInfo;
 import org.iplantc.de.client.models.notifications.Counts;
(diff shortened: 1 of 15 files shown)


de-lib/src/main/java/org/iplantc/de/client/services/impl/MessageServiceFacadeImpl.java
```diff
@@ -56,27 +55,25 @@ protected Counts convertFrom(final String json) {
     }
 
     private final NotificationAutoBeanFactory notesFactory;
-    private final DEProperties deProperties;
     private final DiscEnvApiService deServiceFacade;
     private final UserInfo userInfo;
+    private final String NOTIFICATIONS = "org.iplantc.services.notifications";
```
psarando: Will this require a change to the properties file in our ansible repo?

de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/DesktopPresenterImpl.java
```diff
@@ -669,23 +670,27 @@ public void doPeriodicSessionSave() {
...
-   }
+        getSystemMessages();
+    }
 
-   @Override
-   public void getNotifications() {
-       messageServiceFacade.getRecentMessages(new InitializationCallbacks.GetInitialNotificationsCallback(view, appearance, announcer));
-       messageServiceFacade.getMessageCounts(new NewSysMessageCountCallback());
-   }
+    private void getSystemMessages() {
```
psarando: This method name was a little confusing at first. Maybe it could be `getSystemMessageCounts`?

de-lib/src/main/java/org/iplantc/de/client/services/impl/MessageServiceFacadeImpl.java
ash3rz: Ah, yes, I forgot to mention that in my PR.  After the long break, I'm really glad you caught that because I would've forgotten 😄

de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/DesktopPresenterImpl.java
ash3rz: Makes sense - I'll update

ash3rz:
Thanks for the review!

PR #59, opened 2016-12-16
CORE-8066: added the system ID path to several admin apps endpoint calls

dennisroberts71:


dennisroberts71:
Thanks for the review!

de-lib/src/main/java/org/iplantc/de/client/services/impl/AppUserServiceFacadeImpl.java
```diff
@@ -130,6 +128,9 @@ public void getApps(String id, DECallback<List<App>> callback) {
         });
     }
 
+    public void getApps(String id, DECallback<List<App>> callback) {
```
psarando: It doesn't look like this empty method is used anymore, so it should probably be removed.

de-lib/src/main/java/org/iplantc/de/client/models/sharing/Sharing.java
```diff
@@ -39,25 +45,31 @@ public void setId(String id) {
     }
 
     public boolean isReadable() {
-        return permission != null && (permission.equals(PermissionValue.read) || permission.equals(
-                PermissionValue.write) || permission.equals(PermissionValue.own));
+        return isWritable() || PermissionValue.read.equals(permission);
```
ash3rz: This is much more clear 👍

de-lib/src/main/java/org/iplantc/de/client/models/apps/sharing/AppPermission.java
```diff
@@ -1,18 +1,24 @@
...
-import com.google.web.bindery.autobean.shared.AutoBean;
+import com.google.web.bindery.autobean.shared.AutoBean.PropertyName;
 
 /**
  * Created by sriram on 2/3/16.
  */
 public interface AppPermission {
 
-    @AutoBean.PropertyName("app_id")
+    @PropertyName("app_id")
```
ash3rz: I believe we can shorten this to extend `QualifiedAppId`?

de-lib/src/main/java/org/iplantc/de/client/models/apps/sharing/AppUserPermissions.java
```diff
@@ -1,18 +1,24 @@
...
 import java.util.List;
 
 /**
  * Created by sriram on 2/3/16.
  */
-public interface AppUserPermissions extends HasId, HasName {
+public interface AppUserPermissions extends HasSystemId, HasName {
 
-    @AutoBean.PropertyName("permissions")
+    @PropertyName("app_id")
```
ash3rz: This could also be `QualifiedAppId`

de-lib/src/main/java/org/iplantc/de/client/models/sharing/SharedResource.java
```diff
@@ -8,11 +8,21 @@
...
+    /**
      * name of the shared resource
      */
     private String name;
 
     public SharedResource(String id, String name) {
+        this(null, id, name);
+    }
+
+    public SharedResource(String systemId, String id, String name) {
```
ash3rz: In one of Dennis's previous commits to this PR, he introduced the model for `QualifiedId` which has an ID and system ID.  I'm not sure if it is or isn't worth putting here, just mentioning it.

de-lib/src/main/java/org/iplantc/de/client/models/apps/sharing/AppPermission.java
psarando: Good catch! I only saw `HasQualifiedId` before and hadn't noticed that `QualifiedAppId` interface. 👍

de-lib/src/main/java/org/iplantc/de/client/models/QualifiedId.java
```diff
@@ -0,0 +1,44 @@
...
+    public QualifiedId(final String systemId, final String id) {
+        this.systemId = systemId;
+        this.id = id;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj instanceof QualifiedId) {
+            QualifiedId other = (QualifiedId) obj;
+            return other.getSystemId() == systemId && other.getId() == id;
```
psarando: Since these are strings, they should be compared with `.equals()` instead of `==`.

dennisroberts71: Heh. D'oh!

PR #65, opened 2017-01-20
CORE-7291 Update to GWT 2.8 and GXT 4.0.2

ash3rz:
This is a huge PR!  I don't expect everyone to look over this or to even look over the entire thing due to the number of changes, but I wanted to at least put it there for folks to see if they're interested.

A basic summary of the changes include:

- Some method name changes coming from GXT
- Some method signature changes regarding strings and/or SafeHtml
- Instances where strings with HTML in them were getting printed as regular strings
- GWT 2.8 has [additional checks](https://github.com/gwtproject/gwt/blob/master/user/super/com/google/gwt/emul/javaemul/internal/InternalPreconditions.java) for things like array sizes, null values, etc. that caused new NPEs to surface 
- All CSS converted to [GSS](http://www.gwtproject.org/doc/latest/DevGuideGssVsCss.html)


de-lib/src/main/java/org/iplantc/de/client/services/impl/AppUserServiceFacadeImpl.java
```diff
@@ -175,10 +175,7 @@ public void getAppDetails(final App app, DECallback<App> callback) {
         deServiceFacade.getServiceData(wrapper, new DECallbackConverter<String, App>(callback) {
             @Override
             protected App convertFrom(String object) {
-                Splittable split = StringQuoter.split(object);
-                AutoBean<App> appAutoBean = AutoBeanUtils.getAutoBean(app);
-                AutoBeanCodex.decodeInto(split, appAutoBean);
-                return appAutoBean.as();
+                return AutoBeanCodex.decode(factory, App.class, object).as();
```
psarando: I wonder if this change introduced the bug that the "Fix NPEs in AppDetailsDialog" commit fixed?
It's too bad that this technique won't work anymore. Unless there's another way of fixing this date parsing error? Maybe changing the AutoBean's field type from a Date (or changing the value returned by the service)?

ash3rz: Really good catch!  I think I skimmed over this code too quickly and missed the intent.  What's strange is that if I remove this commit and the NPE fix commit, that doesn't seem to reintroduce the date parsing error, which is suspicious to me.  Either way, I'll go ahead and remove these two commits for now.

ash3rz: Side note, I remembered the date parsing error.  It's that silly `NumberFormatException` that always comes up if you have dev tools open with `Pause on Caught Exceptions` enabled.  It doesn't show up in the console or the logs, but while I was trying to trace down all the various bugs, this was getting in my way enough that I thought I'd try to fix it.  So, it's really just an annoyance.

ash3rz:
Thanks for the reviews!

PR #68, opened 2017-01-25
Migrate PreferencesDialog and DEFeedbackDialog to using AsyncProviderWrapper

ianmcorvidae:
It works!

I have the feeling I'm supposed to be using the presenter here (no other views seem to import ErrorHandler, particularly, but the presenter which corresponds to this does seem to use it via a provider), but I would need advising on how to move this stuff around. This does successfully reduce the size of the initial download, while increasing the leftovers fragment a little bit and creating one extra split (at least in my testing, it bundled the PreferencesDialog with the preexisting AppLaunchWindow split, but DEFeedbackDialog got its own).

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/DesktopPresenterImpl.java
+++ b/de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/DesktopPresenterImpl.java
@@ -53,11 +53,13 @@ import org.iplantc.de.desktop.client.presenter.util.NotificationUtil;
 import org.iplantc.de.desktop.client.presenter.util.NotificationWebSocketManager;
 import org.iplantc.de.desktop.client.presenter.util.SystemMessageWebSocketManager;
 import org.iplantc.de.desktop.client.views.windows.IPlantWindowInterface;
+import org.iplantc.de.desktop.client.views.widgets.PreferencesDialog;
 import org.iplantc.de.desktop.shared.DeModule;
 import org.iplantc.de.fileViewers.client.callbacks.LoadGenomeInCoGeCallback;
 import org.iplantc.de.notifications.client.events.WindowShowRequestEvent;
 import org.iplantc.de.notifications.client.utils.NotifyInfo;
 import org.iplantc.de.notifications.client.views.dialogs.RequestHistoryDialog;
+import org.iplantc.de.shared.AsyncProviderWrapper;
 import org.iplantc.de.shared.DEProperties;
 import org.iplantc.de.shared.NotificationCallback;
 import org.iplantc.de.shared.events.ServiceDown;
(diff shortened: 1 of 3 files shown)


ash3rz:
If I remember correctly, the ErrorHandler is there as mostly a filler.  If the `AsyncProviderWrapper` returns back as a failure, that class has its own `onFailure` method that will run which basically logs the user out.  The only time it _should_ fail is if their session has expired and they've not gotten logged out by making any service calls but instead hitting a codesplit.

Anyway, this looks straightforward to me 🚢

ianmcorvidae:
It [looks](https://github.com/cyverse-de/ui/blob/8ae0c1a/de-lib/src/main/java/org/iplantc/de/commons/client/ErrorHandler.java#L110) like ErrorHandler is supposed to bring up a dialog, at least, which is hopefully correct. Looks like `AsyncProviderWrapper` has code to log people out in the case of a failure. So I guess what might actually be the case is that, when `AsyncProviderWrapper` is in use, it doesn't matter what you put for `onFailure` in the callback, since the wrapper creates its own callback which doesn't call the `onFailure` of the passed-in callback.

ianmcorvidae:
(I'm still leaving this open because I'm curious about my question re: putting this stuff into a presenter, and it's not pressing. If there's no other comments I'll aim to merge it after the release.)

ash3rz:
Ah, on that note - re: the presenter - yeah, if you want to move them to the presenter, it would be more consistent with the rest of the code.  On a much broader note, there's a lot of refactoring that could be done in the Desktop view and presenter to make that code more consistent with the rest of the code, but we should table that for another time/PR.  

But yes, putting this stuff in the presenter would help in the eventual, overall refactor so I think that's a good idea 👍

ianmcorvidae:
I moved the PreferencesDialog stuff to the presenter, as best I thought I knew how to. I'm not sure if there's a better way of doing things than the `final DesktopView.Presenter presenter = this;` I'm using; might be some sort of reorganization of the PreferencesDialog class that's needed to not require that? In any case as far as I know it's fine for now.

The DEFeedbackDialog has some more complex code (particularly, binding SelectEvents for buttons) that I'm not sure the right way to migrate, and in light of current discussions around feedback might be changing in a way involving someone more UI-savvy than me anyway. I'm guessing a lot more of it should be in the DEFeedbackDialog class itself, but...

de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/DesktopPresenterImpl.java
```diff
@@ -456,6 +459,22 @@ public void onCollaboratorsClick() {
...
+	final DesktopView.Presenter presenter = this;
+        preferencesDialogProvider.get(new AsyncCallback<PreferencesDialog>() {
+            @Override
+            public void onFailure(Throwable caught) {
+                ErrorHandler.post(caught);
+            }
+
+            @Override
+            public void onSuccess(final PreferencesDialog dialog) {
+                dialog.show(presenter, userSettings);
```
ash3rz: Ah, right, in this context `this` is the AsyncCallback<PreferencesDialog> type.  So here, you could do `DesktopPresenterImpl.this` to get the correct "this" you want.

ianmcorvidae: Ah! That's definitely better. Thanks!

ianmcorvidae:
Merging, now that the release is done!

PR #69, opened 2017-02-03
CORE-7014 Fix Metadata Template bug during DND

ash3rz:
If you edit a Metadata Template attribute in Belphegor, then try to drag and drop that attribute to a different location, your changes are lost.
This change immediately commits changes to the liststore so changes are preserved.


ash3rz:
Thanks for the review!

PR #72, opened 2017-02-06
CORE-8158 Update admin wording from “Delete” app to “Deprecate” app

ash3rz:
This is to match documentation which prefers one term over the other.


de-lib/src/main/java/org/iplantc/de/theme/base/client/admin/BelphegorDisplayStrings.properties
```diff
@@ -17,7 +17,7 @@ addCategoryPrompt = New Category
 appCategorizeSuccess = Successfully added {0} to the following Categories: {1,list}.
 appCategorizeSuccess[\=1] = Successfully added {0} to the following Category: {1,list}.
 confirmDeleteAppCategory = Do you want to delete the category \"{0}\"?
-confirmDeleteAppTitle = Delete App
+confirmDeprecateAppTitle = Delete App
```
psarando: I was expecting the opposite change for this PR: internally we can still name methods and properties `*deleteApp*`, since we're still marking apps as deleted in our services, but we only need to change labels presented to the user from `delete` to `deprecate`.

de-lib/src/main/java/org/iplantc/de/theme/base/client/admin/ontologies/OntologyDisplayStrings.properties
```diff
@@ -31,8 +31,8 @@ ontologyDeleted = Deleted {0} successfully
 rootColumnLabel = Iri(s)
 hierarchyColumnLabel = Root(s)
 confirmDeleteHierarchy = Are you sure you want to delete the `{0}` hierarchy?<BR><BR>NOTE: You can always add it back by re-saving any root hierarchy to which it is a child.
-confirmDeleteAppTitle = Delete App
-confirmDeleteAppWarning = Are you sure you want to delete the app `{0}`?
+confirmDeprecateAppTitle = Deprecate App
+confirmDeprecateAppWarning = Are you sure you want to deprecate the app `{0}`?
```
psarando: I see this label updated from `delete` to `deprecate` 👍 but I still don't think we need to rename the internal property name.

de-lib/src/main/java/org/iplantc/de/theme/base/client/admin/BelphegorDisplayStrings.properties
ash3rz: Hm... that's an interesting point.  I went through and updated the methods and properties for consistency between what's seen in the UI, but if it's clearer to leave method names as they were to match the service verbiage, then I'm ok with reverting.

psarando: That's a good point. Since you put it that way, it does make sense to rename the properties and method names for the display strings to match the term used in the strings.

For non-display-string-methods and fields (like in the Views), I would still stick with `delete` over `deprecate`, especially since the term `delete` is still used in a lot of places.

ash3rz:
I updated the view and appearance methods to go back to using the "delete" wording.  Thanks for the review!

PR #73, opened 2017-02-08
CORE-6604 make error details panel render html.

sri2k1us:
While addressing the original issue, I also found UI is swallowing some of the errors. I found if the error message cannot be parsed into json (example: Gateway timeout), it results in JS exception. So I added try/catch to few callbacks in DiskResource Module. This problem may be present in other modules as well. Probably a separate PR is needed for the rest.

Also found UI was throwing an JS exception when service returned `ERR_TOO_MANY_RESULTS` as error code.


de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/callbacks/NcbiSraSetupCompleteCallback.java
```diff
@@ -40,12 +40,17 @@ public void onSuccess(String result) {
     @Override
     public void onFailure(Integer statusCode, Throwable caught) {
         unmaskCaller();
+        try {
         DiskResourceErrorAutoBeanFactory factory = GWT.create(DiskResourceErrorAutoBeanFactory.class);
```
ash3rz: Small spacing issue here, not a huge deal.

de-lib/src/main/java/org/iplantc/de/commons/client/views/dialogs/ErrorDialog.ui.xml
```diff
@@ -55,8 +55,7 @@
                                      collapsible="true">
                         <container:VerticalLayoutContainer ui:field="detailsContainer">
                             <container:child layoutData="{topData}">
-                                <gxtForm:TextArea ui:field="descriptionArea"
-                                                  readOnly="true"/>
+                                <g:HTML ui:field="descriptionArea" />
```
ianmcorvidae: this might be overwhelmingly minor, but how does this change from text area to HTML affect our users' ability to copy-paste from this area into bug reports? A lot of the time this box's contents make it in, so we should make sure it's still reasonably copyable

de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/callbacks/DiskResourceMoveCallback.java
```diff
@@ -34,10 +34,15 @@ public void onSuccess(DiskResourceMove result) {
...
-        ErrorHandler.post(errorBean.as(), caught);
+        try {
+            DiskResourceErrorAutoBeanFactory factory = GWT.create(DiskResourceErrorAutoBeanFactory.class);
+            AutoBean<ErrorDiskResourceMove> errorBean =
+                    AutoBeanCodex.decode(factory, ErrorDiskResourceMove.class, caught.getMessage());
+
+            ErrorHandler.post(errorBean.as(), caught);
+        } catch (Exception e) {
+            ErrorHandler.post(caught);
+        }
```
ianmcorvidae: The structure of this try-catch block seems repeated a lot; I don't know if it can be put somewhere to avoid this, but it could be kinder to our future selves to do so.

de-lib/src/main/java/org/iplantc/de/commons/client/views/dialogs/ErrorDialog.ui.xml
sri2k1us: I double checked. The message in details panel is still copyable!

de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/callbacks/DiskResourceMoveCallback.java
sri2k1us: @ash3rz @ianmcorvidae I have seen this case happen only for Gateway timeout only happening in Data window.  I am not sure if there are any cases / services that will return non-json response. So for this PR, I just decided to add it places i think we need this logic.

PR #74, opened 2017-02-10
Core 8438 Allow users request support for their analysis from DE analysis window

sri2k1us:


ash3rz:
I haven't had a chance to review yet, but at a glance I don't see any tests 😉

de-lib/src/main/java/org/iplantc/de/analysis/client/views/AnalysisColumnModel.java
```diff
@@ -129,6 +114,10 @@ public String getPath() {
         endDate.setHeader(appearance.endDate());
 
         status.setHeader(appearance.status());
+        status.setMenuDisabled(true);
+        status.setCell(new AnalysisUserSupportCell());
+        status.setSortable(false);
+        status.setHideable(false);
```
psarando: Does sorting not work for the status column anymore? If sorting still works on the status value, I don't think we should remove it (the name and date columns use custom cells and they are still sortable).

de-lib/src/main/java/org/iplantc/de/client/services/impl/DEUserSupportServiceFacadeImpl.java
```diff
@@ -12,29 +13,25 @@
...
 /**
- * Provides access to remote services for submitting user feedback.
+ * Created by sriram on 12/1/16.
  */
-@SuppressWarnings("nls")
-public class DEFeedbackServiceFacadeImpl implements DEFeedbackServiceFacade {
+public class DEUserSupportServiceFacadeImpl implements DEUserSupportServiceFacade {
 
-    private static String FEEDBACK_SERVICE_PATH = "feedback";
+    private static String SUPPORT_SERVICE_PATH = "support-email";
```
psarando: I also think that would be better. Then if the path changes, we only have to change the config and not the code.

de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/DesktopPresenterImpl.java
```diff
@@ -648,9 +648,7 @@ public void show(final WindowConfig config,
 
     @Override
     public void submitUserFeedback(Splittable splittable, final IsHideable isHideable) {
-        StringQuoter.create(userInfo.getUsername()).assign(splittable, "username");
-        StringQuoter.create(Window.Navigator.getUserAgent()).assign(splittable, "User-agent");
-        feedbackServiceProvider.get().submitFeedback(splittable, new AsyncCallback<Void>() {
+       userSupportServiceProvider.get().submitSupportRequest(splittable, new AsyncCallback<Void>() {
```
psarando: If we're renaming the service call from `submitFeedback` to `submitSupportRequest`, should we rename this presenter method too? Also, what about the `DesktopView`: are we still letting users submit feedback from the desktop help menu?

de-lib/src/main/java/org/iplantc/de/theme/base/client/analyses/AnalysesMessages.properties
```diff
@@ -38,3 +38,14 @@ shareSupportMi = Share with support...
 shareSupportConfirm = Your output folder(s) and file(s) will be shared with CyVerse support team. Do you want to continue ?
 shareWithInput = Yes. Share my input files as well!
 shareOutputOnly = Share output folder(s) only!
+outputUnexpected = Analysis completed but I got unexpected output.
+noOutput = Analysis completed but I got no output.
+selectCondition = Select the condition of your analysis:
+comments = Comments
+needHelp = I still need help!
+submit = Submit
+agreeToShare = <span style="overflow:wrap;">I agree to <a href="https://wiki.cyverse.org/wiki/x/KBfHAQ" target="_blank">share my analysis</a> details, input file(s), output file(s) and logs with CyVerse Support. <br/>If you want any of the <a href="#" target="_blank">Science Informaticians</a> to recuse themselves from assisting with this problem please list their name(s) in comments.</span>
```
psarando: Is the phrase `Science Informaticians` supposed to be a link? Also, there should probably be a comma in `with this problem, please list`.

de-lib/src/main/java/org/iplantc/de/theme/base/client/analyses/cells/AnalysisUserSupportCellDefaultAppearance.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    public interface AnalysisUserSupportCellResources extends ClientBundle {
+        @Source("AnalysisUserSupportCell.css")
+        AnalysisUserCellStyles css();
+
+    }
+
+    interface Templates extends SafeHtmlTemplates {
+        @SafeHtmlTemplates.Template("<span name=\"{0}\" title=\" {3}\" class='{2}'>{1}</span>")
```
psarando: Looks like there's an extra space in the `title` here.

de-lib/src/main/java/org/iplantc/de/analysis/client/presenter/AnalysesPresenterImpl.java
```diff
@@ -448,7 +475,83 @@ public void onSuccess(AnalysisStepsInfo result) {
...
+            public void onSuccess(String result) {
+                emailSupport(parent);
+            }
+        });
+    }
+
+    @Override
+    public void onUserSupportRequested(AnalysisUserSupportRequestedEvent event) {
+        final Analysis value = event.getValue();
+        final AnalysisUserSupportDialog ausd = new AnalysisUserSupportDialog(value);
```
ianmcorvidae: Just because I've been looking at code splits...

It'd be nice if this class could have a `@Inject AsyncProviderWrapper<AnalysisUserSupportDialog>` and then use that provider to get a dialog and call something to pass in the value then. That way, the code for the dialog doesn't go into the initial fragment.

ianmcorvidae: (the dialog for viewing analysis parameters seems to do this approach reasonably well, at least with my understanding of the code)

de-lib/src/main/java/org/iplantc/de/analysis/client/views/AnalysisColumnModel.java
psarando: Also, if we can enable sorting for this column, then we should also enable the menu.
For grids that take a while to load after sorting, the menu is a time-saver when the user knows they want to sort descending in the first place (such as in the data window).
I've encountered columns that are sortable, but the menu is disabled, so I have to sort ascending, wait a second or more for the sort to complete, then click the column header again to get the descending sort I want (and also wait a second or more for the sort to complete in the correct order).

de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/DesktopPresenterImpl.java
```diff
@@ -150,7 +150,7 @@ public void onSuccess(Counts result) {
     @Inject DEClientConstants deClientConstants;
     @Inject DEProperties deProperties;
     @Inject Provider<ErrorHandler> errorHandlerProvider;
-    @Inject Provider<DEFeedbackServiceFacade> feedbackServiceProvider;
+    @Inject Provider<DEUserSupportServiceFacade> userSupportServiceProvider;
```
ianmcorvidae: Another code-split potential -- if this is `AsyncProviderWrapper` instead of `Provider`, and the `.get()` below becomes a: 

```java
userSupportServiceProvider.get(new AsyncCallback<DEUserSupportServiceFacade>(){
    @Override
    public void onFailure (Throwable caught) {} // no-op, AsyncProviderWrapper doesn't care

    @Override
    public void onSuccess(DEUserSupportServiceFacade supportService) {
        supportService.submitSupportRequest(...)
    }
});
```

... then all the code of the service facade doesn't need to be there on initial load. This is used one other place, if I'm not mistaken, so that would need the same treatment or only the injection code would actually end up in the split.

de-lib/src/main/java/org/iplantc/de/analysis/client/presenter/AnalysesPresenterImpl.java
```diff
@@ -448,7 +475,83 @@ public void onSuccess(AnalysisStepsInfo result) {
...
+                shareWithSupport(value, parent);
+            }
+        });
+
+        ausd.show();
+
+    }
+
+    protected void emailSupport(Splittable parent) {
+        supportService.submitSupportRequest(parent, new AsyncCallback<Void>() {
```
ianmcorvidae: (here's the other spot it'd need to be changed to use an `AsyncProviderWrapper`, I think)

sri2k1us:
Closing this PR and opening a new one!

PR #76, opened 2017-02-13
CORE-8597 Fix issue where private apps do not show lock icon

ash3rz:
I accidentally removed this code to fix a bug in Belphegor where this was causing NPE, not realizing this also affected the DE.  Now there's a null check so Belphegor can do its thing and the DE will show the lock.


ash3rz:
Thanks for the review!

PR #78, opened 2017-02-16
CORE-8599 and CORE-8441

ash3rz:
CORE-8441 - Private apps were listing the Unclassified hierarchies in the App Details view, which incorrectly conveyed that those apps were public.

CORE-8599 - Hitting the Refresh button in the Apps window causes the go() method to be rerun.  The unclassifiedHierarchies list was never reset in the go() method which would cause dupes to get added every time the Refresh button was pressed.  This led to an incorrect number of Unclassified hierarchies listed in the App Details view.


ash3rz:
Thanks for the review!

PR #79, opened 2017-02-17
CORE-4232 Rename Copy Link button in Data Link window for clarity

ash3rz:
When creating public data links, users thought the verbiage and buttons were trying to convey that opening the data link window automatically copied the data link into the user's clipboard.  This is changing the wording so that it's clearer that users still have to manually copy the link.

From "Copy Link" to "Show Link"
image

From "Copy" title to "Public Data Link" title.  Cancel button has been removed so as to not indicate there is an action to cancel.
image


ash3rz:
Thanks for the review!

PR #80, opened 2017-02-17
CORE-7539 When relaunching an analysis, the Retain Inputs flag does not persist

ash3rz:
This PR fixes a bug where the UI was not checking the Retain Inputs flag (debug key) during analysis relaunch.


ash3rz:
Thanks for the review!

PR #83, opened 2017-02-22
CORE-7582 GWT Stack Trace Deobfuscation

ash3rz:
This branch is based off of the GWT 2.8 branch, so only the last commit is relevant.

The one thing I'm not sure of is how to get the module name (DE or Belphy) from the server side code in a clean way. Right now I'm doing a regex and I get the feeling there's an easier way but I don't know the server side stuff well enough.

### DIFF
--- a/de-webapp/src/main/java/org/iplantc/de/server/rpc/GwtRpcController.java
+++ b/de-webapp/src/main/java/org/iplantc/de/server/rpc/GwtRpcController.java
@@ -67,9 +67,6 @@ public class GwtRpcController extends RemoteServiceServlet implements Controller
 
             RPCRequest rpcRequest = RPC.decodeRequest(payload,
                                                       this.remoteServiceClass);
-            if (remoteService instanceof DeGwtRemoteLoggingServiceImpl) {
-                ((DeGwtRemoteLoggingServiceImpl)remoteService).setRequest(getThreadLocalRequest());
-            }
 
             // delegate work to the spring injected service
             return RPC.invokeAndEncodeResponse(this.remoteService, rpcRequest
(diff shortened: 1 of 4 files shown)


de-webapp/src/main/java/org/iplantc/de/server/rpc/GwtRpcController.java
```diff
@@ -67,6 +67,9 @@ public String processCall(String payload) throws SerializationException {
 
             RPCRequest rpcRequest = RPC.decodeRequest(payload,
                                                       this.remoteServiceClass);
+            if (remoteService instanceof DeGwtRemoteLoggingServiceImpl) {
+                ((DeGwtRemoteLoggingServiceImpl)remoteService).setRequest(getThreadLocalRequest());
```
ianmcorvidae: Just an "educating Ian more" comment: Everything makes sense to me other than this block -- what does it do, and where does `getThreadLocalRequest()` come from? I can't find other references to it in the UI, at least.

ash3rz: This is mostly beyond my knowledge - I followed some examples of other people implementing GWT's stack trace deobfuscator.   But anyway, this method is coming from extending `RemoteServiceServlet`.

ash3rz: I was looking at this again and realized we're already passing the request during `handleRequest` as long as the RemoteService implements HasHttpServletRequest.  I went ahead and had `DeGwtRemoteLoggingServiceImpl` do that which removed the need to have this separate bit of logic.

ash3rz:
If nobody minds, I think I'll go ahead and merge this so it can be part of the drop tomorrow.

PR #84, opened 2017-02-23
Core 8438 Allow users request support for their analysis from DE analysis window

sri2k1us:


de-lib/src/main/java/org/iplantc/de/analysis/client/views/dialogs/AnalysisUserSupportDialog.java
```diff
@@ -0,0 +1,237 @@
...
+        vlc.setScrollMode(ScrollSupport.ScrollMode.AUTO);
+    }
+
+    public void renderHelp(Analysis analysis) {
+        this.selectedAnalysis = analysis;
+        HTML text = null;
+        switch (AnalysisExecutionStatus.fromTypeString(selectedAnalysis.getStatus().toLowerCase())) {
+            case SUBMITTED:
+                if (selectedAnalysis.getSystemId().equalsIgnoreCase(appearance.deSystemId())) {
+                    text = new HTML(renderer.renderCondorSubmitted(selectedAnalysis));
```
ash3rz: I would still move the renderer to the appearance class instead of having it here.

de-lib/src/main/java/org/iplantc/de/analysis/client/views/dialogs/AnalysisUserSupportDialog.java
```diff
@@ -0,0 +1,237 @@
...
+
+        String supportRequestSuccess();
+
+        String deSystemId();
+    }
+
+
+    private Analysis selectedAnalysis;
+    private HelpRendererTemplates renderer = GWT.create(HelpRendererTemplates.class);
+    private AnalysisUserSupportAppearance appearance = GWT.create(AnalysisUserSupportAppearance.class);
```
ash3rz: The appearance can be injected either with @Inject or in the constructor.

de-lib/src/main/java/org/iplantc/de/theme/base/client/analyses/AnalysesMessages.properties
```diff
@@ -38,3 +38,15 @@ shareSupportMi = Share with support...
...
+selectCondition = Select the condition of your analysis:
+comments = Comments
+needHelp = I still need help!
+submit = Submit
+agreeToShare = <span style="overflow:wrap;">I agree to <a href="https://wiki.cyverse.org/wiki/x/KBfHAQ" target="_blank">share my analysis</a> details, output file(s) and logs with CyVerse Support. <br/>If you want any of the <a href="https://wiki.cyverse.org/wiki/x/XRfHAQ" target="_blank">Science Informaticians</a> to recuse themselves from assisting with this problem, please list their name(s) in comments.</span>
+disclaimer = Disclaimer
+termsOfSupport = You must agree to terms of support!
+supportRequestFailed = Unable to submit your support request. Please email support at support@cyverse.org.
+supportRequestSuccess = Your support request is submitted. A support staff will contact you shortly.
+requestHelp = requesting help with Analysis
```
ash3rz: "requesting" should probably capitalized.  And actually, Kathleen should probably review these strings and the information in those HTML files that were added as well.

de-lib/src/main/java/org/iplantc/de/analysis/client/views/dialogs/AnalysisUserSupportDialog.java
sri2k1us: 👍

de-lib/src/main/java/org/iplantc/de/theme/base/client/analyses/AnalysesMessages.properties
sri2k1us: This part of a string that starts with username. For e.g: In the email to Science team the subject will be `sriram requesting help with Analysis`. All the HTML text has been reviewed by Katheleen and Science team already. The plan is for Katheleen to review them once again in QA Env.

de-lib/src/main/java/org/iplantc/de/analysis/client/views/dialogs/AnalysisUserSupportDialog.java
sri2k1us: Am not sure what exactly you mean. Maybe we can discuss this in person.

de-lib/src/main/java/org/iplantc/de/theme/base/client/analyses/AnalysesMessages.properties
ash3rz: Ah I didn't catch that - great!

psarando: Ah, in that case, I would change the name of this property to be a little more descriptive. Even just `userRequestingHelpSubject` might be enough?

ianmcorvidae:
The fixes to the code split stuff I mentioned look good to me. I've run a compile report and there's a little bit that I'm not sure why it's ending up in the leftovers fragment (part of the service facade, but it looks like it should always be behind the async provider...), but things still look like they're behaving essentially right.

de-lib/src/main/java/org/iplantc/de/analysis/client/views/dialogs/AnalysisUserSupportDialog.java
ash3rz: I mean instead of having `HelpRendererTemplates` referenced in this class since it affects the appearance and uses a template, have it referenced in `AnalysisUserSupportDefaultAppearance`.  Then you would add those methods to `AnalysisUserSupportAppearance`.
This line would then become:
text = new HTML(appearance.renderCondorSubmitted(selectedAnalysis));
or even
text = appearance.renderCondorSubmitted(selectedAnalysis);
This way `AnalysisUserSupportDefaultAppearance` has complete control over all of the text and also which templates are used.

de-lib/src/main/java/org/iplantc/de/analysis/client/views/dialogs/AnalysisUserSupportDialog.java
```diff
@@ -0,0 +1,237 @@
...
+        vlc.getElement().getStyle().setBackgroundColor(appearance.backgroudColor());
+        vlc.setScrollMode(ScrollSupport.ScrollMode.AUTO);
+    }
+
+    public void renderHelp(Analysis analysis) {
+        this.selectedAnalysis = analysis;
+        HTML text = null;
+        switch (AnalysisExecutionStatus.fromTypeString(selectedAnalysis.getStatus().toLowerCase())) {
+            case SUBMITTED:
+                if (selectedAnalysis.getSystemId().equalsIgnoreCase(appearance.deSystemId())) {
```
ash3rz: Also, instead of checking the analysis system ID against a string in the appearance, I think it might need to be against `DEClientConstants`?

dennisroberts71: Yeah, it should be compared to the value from `DEClientConstants`. It's not a huge deal, but having one spot in the UI that is the source of the DE system ID is probably best.

sri2k1us: Now that I think about it, It should be defined in DE.properties since the property affects both service and the UI. It will also eliminate the need to UI rebuild when it changes.

de-lib/src/main/java/org/iplantc/de/theme/base/client/analyses/AnalysesMessages.properties
sri2k1us: :ok_man:

de-lib/src/main/java/org/iplantc/de/analysis/client/views/dialogs/AnalysisUserSupportDialog.java
sri2k1us: Ahh! Now I remember why I did not have that injected. I am using appearance in the constructor to initialize button text and background color. If I inject the appearance, it will be `null` because it will be available only after the constructor is complete.

de-lib/src/main/java/org/iplantc/de/analysis/client/views/dialogs/AnalysisUserSupportDialog.java
sri2k1us: Got it!

de-lib/src/main/java/org/iplantc/de/analysis/client/views/dialogs/AnalysisUserSupportDialog.java
dennisroberts71: Yeah, that would work, although it's defined as a constant in the apps service at the moment. I suppose we could have the service return the DE system ID and the UI code could keep track of it. That would make the apps service the definitive source of the DE system ID and avoid having to have it in the config file.

sri2k1us: Will that be part of bootstrap service?

dennisroberts71: Yeah, I think that would be the easiest place to put it.

sri2k1us: Cool! I will wait for that change.

de-lib/src/main/java/org/iplantc/de/client/models/UserInfo.java
```diff
@@ -73,15 +77,19 @@ public void init(String userInfoJson) {
...
 
     public boolean hasErrors() {
         return hasUserProfileError() ||
                hasDataInfoError() ||
                hasPreferencesError() ||
                hasSessionError() ||
-               hasWorkspaceError();
+               hasWorkspaceError() ||
+               hasSystemsError() ||
+               hasAppsInfoError();
```
psarando: Actually, there wouldn't be separate errors for the `workspace` or `system_ids` keys. These come from the apps service, so if apps is down, then only the parent `apps_info` key will contain an error message.
So `hasAppsInfoError` should cover all 3 of these.

sri2k1us: Ohh! Didn't know that! Will fix it!

ash3rz:
Latest changes look good to me 👍  (Not sure if there are still more pending changes though)

sri2k1us:
We still have a change related to private apps / permissions. I am waiting for @psarando 's change on services side for that!

sri2k1us:
All necessary service changes are done and PR comments are addressed.

PR #87, opened 2017-03-01
CORE-8438 fix null start / end date issue

sri2k1us:


de-lib/src/main/java/org/iplantc/de/theme/base/client/analyses/AnalysisUserSupportDefaultAppearance.java
```diff
@@ -118,8 +118,12 @@ public SafeHtml renderCompletedNoOutput(Analysis selectedAnalysis) {
...
     public SafeHtml renderSubmitToSupport(Analysis selectedAnalysis, UserProfile userProfile) {
         return renderer.renderSubmitToSupport(selectedAnalysis,
-                                              new Date(selectedAnalysis.getStartDate()),
-                                              new Date(selectedAnalysis.getEndDate()),
+                                              selectedAnalysis.getStartDate() != 0 ?
+                                              new Date(selectedAnalysis.getStartDate()).toString() :
+                                              "",
+                                              selectedAnalysis.getEndDate() != 0 ?
+                                              new Date(selectedAnalysis.getEndDate()).toString() :
+                                              "",
```
psarando: It's too bad you had to switch from Dates to Strings.
Does this renderer output `null` in the email for null Dates?

sri2k1us: Yes! It prints null in the email / Share screen when date is 0L. Also Date object cannot be initialized with null.

PR #88, opened 2017-03-02
CORE-8447 fix display position of intro text and change links.

sri2k1us:


ash3rz:
Looks pretty straightforward :shipit:

PR #90, opened 2017-03-02
CORE-8602 Part 2 - Refactor the Analysis module for code consistency

ash3rz:
These are refactoring changes that probably merit a second pair of eyes.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/analysis/client/presenter/AnalysesPresenterImpl.java
+++ b/de-lib/src/main/java/org/iplantc/de/analysis/client/presenter/AnalysesPresenterImpl.java
@@ -436,8 +436,7 @@ public class AnalysesPresenterImpl implements AnalysesView.Presenter,
         final AnalysisSharingPresenter sharingPresenter = new AnalysisSharingPresenter(analysisService,
                                                                                  selected,
                                                                                  sharingView,
-                                                                                 collaboratorsUtil,
-                                                                                 jsonUtil);
+                                                                                 collaboratorsUtil);
        aSharingDialogProvider.get(new AsyncCallback<AnalysisSharingDialog>() {
             @Override
             public void onFailure(Throwable caught) {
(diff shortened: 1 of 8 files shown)


de-lib/src/main/java/org/iplantc/de/analysis/client/presenter/sharing/AnalysisSharingPresenter.java
```diff
@@ -46,10 +43,10 @@
 
     private final class LoadPermissionsCallback extends AnalysisCallback<String> {
         private final class GetUserInfoCallback implements AsyncCallback<FastMap<Collaborator>> {
-            private final List<String> usernames;
+            private final AnalysisUserPermissionsList analysisPermsList;
```
ash3rz: The changes in this file mirror the changes from a refactor done in the [AppSharingPresenter](https://github.com/cyverse-de/ui/pull/59/commits/ff4511f5e19224ab8251d83804d980d51dce4933)

de-lib/src/main/java/org/iplantc/de/client/services/UserSessionServiceFacade.java
```diff
@@ -21,8 +22,10 @@
 
     void postClientNotification(JSONObject notification, AsyncCallback<String> callback);
```
ash3rz: The Data window is using this callback with the JSONObject and I didn't want to do that refactor in this PR.

de-lib/src/main/java/org/iplantc/de/analysis/client/presenter/AnalysesPresenterImpl.java
```diff
@@ -394,8 +394,7 @@ public void onShareSelected(List<Analysis> selected) {
         final AnalysisSharingPresenter sharingPresenter = new AnalysisSharingPresenter(analysisService,
```
ash3rz: There's more refactoring that can be done so that we're injecting the sharing presenter, but since the App,Analysis, and Data sharing presenters all share the same interface, I think I'll save those changes for one PR together.

ash3rz:
I think I'll go ahead and merge this so the changes for this ticket can be complete before the next drop.  Still open for any reviews as always.

PR #92, opened 2017-03-03
CORE-8613 Log errors from the UI during bootstrap

ash3rz:
We discovered recently that if there are any kind of errors during bootstrap (for example, if the backend has updated the bootstrap JSON), the UI will use default values as expected, but will hide the cause of the use of defaults.

This change will log the bootstrap errors that occur within the UI prior to using defaults for more transparency.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/client/models/UserInfo.java
+++ b/de-lib/src/main/java/org/iplantc/de/client/models/UserInfo.java
@@ -78,8 +78,10 @@ public class UserInfo {
         preferences = userBootstrap.getPreferences();
         session = userBootstrap.getSession();
         appsInfo = userBootstrap.getAppsInfo();
-        workspace = appsInfo.getWorkspace();
-        systemIds = appsInfo.getSystemsIds();
+        if (appsInfo != null) {
+            workspace = appsInfo.getWorkspace();
+            systemIds = appsInfo.getSystemsIds();
+        }
     }
 
     public boolean hasErrors() {
--- a/de-lib/src/main/java/org/iplantc/de/client/models/UserInfo.java
+++ b/de-lib/src/main/java/org/iplantc/de/client/models/UserInfo.java
@@ -18,6 +18,7 @@ import com.google.web.bindery.autobean.shared.Splittable;
 
 import java.util.List;
 import java.util.Map;
+import java.util.logging.Logger;
 
 /**
  * Holds all the information about an user.
@@ -56,6 +57,7 @@ public class UserInfo {
     private AppsInfo appsInfo;
     private List<WindowState> savedOrderedWindowStates;
     private static String AGAVE_AUTH_KEY = "agave";
+    private Logger LOG = Logger.getLogger(UserInfo.class.getName());
 
     /**
      * Constructs a default instance of the object with all fields being set to null.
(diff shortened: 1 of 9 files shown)


de-lib/src/main/java/org/iplantc/de/client/models/UserInfo.java
```diff
@@ -78,8 +80,41 @@ public void init(String userInfoJson) {
         preferences = userBootstrap.getPreferences();
         session = userBootstrap.getSession();
         appsInfo = userBootstrap.getAppsInfo();
-        workspace = appsInfo.getWorkspace();
-        systemIds = appsInfo.getSystemsIds();
+        if (appsInfo != null) {
```
ash3rz: Well in the case where maybe the backend updates the key from `apps_info` to `app_info`, for example, we can avoid getting an NPE that would cause the UI to be stuck with an infinite loading spinner.

de-lib/src/main/java/org/iplantc/de/client/models/UserInfo.java
```diff
@@ -78,8 +80,41 @@ public void init(String userInfoJson) {
...
+
+    private void logBootStrapError(String key, HasSplittableError obj) {
+        String errorMsg;
+        errorMsg = "Bootstrap error within '" + key + "' key: ";
+        if (obj != null) {
+            errorMsg += obj.getError().getPayload();
+        } else {
+            errorMsg += "null value or key parsing error";
+        }
+        LOG.severe(errorMsg);
```
ash3rz: Server side logs.

de-lib/src/main/java/org/iplantc/de/client/models/UserInfo.java
sri2k1us: 👌

de-lib/src/main/java/org/iplantc/de/client/models/UserInfo.java
sri2k1us: Do we have do any kind of set-up for server side logging? I am referring to http://www.gwtproject.org/doc/latest/DevGuideLogging.html#Client_vs_Server_side_Logging

de-lib/src/main/java/org/iplantc/de/client/models/UserInfo.java
```diff
@@ -78,8 +80,41 @@ public void init(String userInfoJson) {
...
+            workspace = appsInfo.getWorkspace();
+            systemIds = appsInfo.getSystemsIds();
+        }
+
+        checkForErrors();
+    }
+
+    private void checkForErrors() {
+        if (userProfile == null || hasUserProfileError()) {
+            logBootStrapError(UserBootstrap.USER_INFO_KEY, null);
```
ianmcorvidae: (this is just an OCD thing on my part, feel free to ignore)

One of these says `BootStrap` and one says `Bootstrap`, which maybe isn't likely to confuse but is inconsistent (I'd guess `Bootstrap` is the more common usage here, probably?)

de-lib/src/main/java/org/iplantc/de/client/models/UserInfo.java
ash3rz: Nope, that's all already set up.  So when an error happens during bootstrap, the error message will be found in the logs on the UI box.  If you search for `Logger.getLogger` in the repo, you can see we're using this in a few spots like AsyncCallbackWrapper, DiskResources, etc.

de-lib/src/main/java/org/iplantc/de/client/models/UserInfo.java
ash3rz: Oops, yeah, I'll update.

de-lib/src/main/java/org/iplantc/de/client/models/UserInfo.java
sri2k1us: Cool! I didn't see any GWT specific setup that why I was curious but I think it taken care of by `logback.xml`

sri2k1us: My bad! I do see the set-up under `discoveryenvironment.gwt.xml`

ash3rz:
If there are no other comments, I think I'll go ahead and merge this one.  Thanks for the reviews!

PR #93, opened 2017-03-09
CORE-8483 Add `save metadata to file` functionality to metadata dialog.

sri2k1us:
This functionality already exists under `Metadata` menu in Data window. Eric Lyons suggested, that this functionality should be available from the metadata dialog as well.


de-lib/src/main/java/org/iplantc/de/diskResource/client/ToolbarView.java
```diff
@@ -18,7 +18,6 @@
 import org.iplantc.de.diskResource.client.events.selection.RefreshFolderSelected.HasRefreshFolderSelectedHandlers;
 import org.iplantc.de.diskResource.client.events.selection.RenameDiskResourceSelected.HasRenameDiskResourceSelectedHandlers;
 import org.iplantc.de.diskResource.client.events.selection.RestoreDiskResourcesSelected.HasRestoreDiskResourceSelectedHandlers;
-import org.iplantc.de.diskResource.client.events.selection.SaveMetadataSelected.HasSaveMetadataSelectedEventHandlers;
```
ash3rz: Since you already had the Handlers set up for this, was there a reason you preferred to switch to the EventBus now?

de-lib/src/main/java/org/iplantc/de/diskResource/client/views/metadata/DiskResourceMetadataViewImpl.java
```diff
@@ -295,6 +297,11 @@ void onImportSelected(SelectEvent event) {
         presenter.onImport(additionalMdgrid.getSelectionModel().getSelectedItems());
     }
 
+    @UiHandler("saveToFileButton")
+    void onSaveToFileButtonSelected(SelectEvent event) {
+         presenter.onSaveToFile();
```
ash3rz: It'd be nice if the view fired off its own event instead of calling the presenter, but since this whole class is calling the presenter, that might be a refactor saved for another time.

de-lib/src/main/java/org/iplantc/de/diskResource/client/ToolbarView.java
sri2k1us: Handlers were wired up for items under `Metadata` menu. So `GridViewPresenterImpl` will handle events only from those menu items and not from the new button in `DiskResourceMetadataViewImpl`. By changing the event to fire/listen on `EventBus`, `GridViewPresenterImpl` will handle both the cases. I haven't figured out if there is a way to pass these Handlers into Dialogs.

ash3rz: Gotcha. I took a quick peek at it and I think to do it without the EventBus would require some other refactors that would be outside the scope of this PR.  I think this is an appropriate solution.

PR #94, opened 2017-03-09
CORE-8619 Add loading mask to Data window's Details panel

ash3rz:


de-lib/src/main/java/org/iplantc/de/diskResource/client/views/details/DetailsViewImpl.java
```diff
@@ -178,12 +179,14 @@ public HandlerRegistration addSubmitDiskResourceQueryEventHandler(SubmitDiskReso
...
+        mask(appearance.loadingMask());
         if (event.getSelection().isEmpty()
                 || event.getSelection().size() != 1 || event.getSelection().get(0).isFilter()) {
             bind(null);
             // Hide table
             table.addClassName(appearance.css().hidden());
             emptyDetails.removeClassName(appearance.css().hidden());
+            unmask();
             return;
         }
```
psarando: Could the `mask` be moved after this `if` check? Then it wouldn't need the `unmask` inside?

ash3rz: Ah yeah, that's cleaner.  I'll add that before merging 👍

ash3rz:
Thanks for the review!

PR #96, opened 2017-03-14
CORE-8625 Update UI logging during logout

ash3rz:
The logging wasn't as bad as I initially thought - mostly had to switch some client side logging to server side.


ash3rz:
Cool, thanks for the review!

PR #98, opened 2017-03-16
CORE-8643 Refactor Apps.Integration module for code consistency Part 1

ash3rz:
Here's the first set of easy changes for the app editor.  If no one has time to review, I feel like these are still ok to merge.


ash3rz:
Since the release is done, I think I'll go ahead and merge in these changes so I can open the PR based on top of this one (sorry, an even bigger PR).

de-lib/src/main/java/org/iplantc/de/apps/integration/client/gin/AppEditorGinModule.java
```diff
@@ -8,10 +8,33 @@
 import org.iplantc.de.apps.integration.client.view.AppIntegrationPalette;
 import org.iplantc.de.apps.integration.client.view.AppsEditorView;
 import org.iplantc.de.apps.integration.client.view.AppsEditorViewImpl;
-import org.iplantc.de.apps.integration.client.view.propertyEditors.*;
+import org.iplantc.de.apps.integration.client.view.deployedComponents.cells.DCNameHyperlinkCell;
```
psarando: This is a general comment about the UI code (not something I think should have been done in this PR), but as we make changes, we should all feel free to rename anything `*DeployedComponent*` (or `DC`) to `*Tool*`.

"Deployed Component" is our obsolete term for "Tools" 🗝

ash3rz: I totally agree - I left it in case there was some preference for "Deployed Component" that I wasn't aware of.  I'll change in the future.

PR #99, opened 2017-03-17
Core 8591 Allow repeating attributes to be specified in metadata templates

sri2k1us:
* Allow repeating of fields except checkbox and Text Area
* Use `+` and `-` buttons to add and remove fields
* Attribute duplication in grid view should be rendered as repeating fields in template
* Save duplicated fields


de-lib/src/main/java/org/iplantc/de/diskResource/client/views/metadata/dialogs/MetadataTemplateViewDialog.java
```diff
@@ -49,45 +51,54 @@
  * Created by sriram on 5/9/16.
  */
 public class MetadataTemplateViewDialog extends IPlantDialog {
-   final MetadataView.Appearance appearance = GWT.create(MetadataView.Appearance.class);
+    final MetadataView.Appearance appearance = GWT.create(MetadataView.Appearance.class);
```
ash3rz: This dialog is pretty huge, it can probably be split into a separate dialog and view class.

de-lib/src/main/java/org/iplantc/de/diskResource/client/views/metadata/dialogs/MetadataTemplateViewDialog.java
```diff
@@ -49,45 +51,54 @@
...
-        loadTemplateAttributes();
+        initTemplate();
         add(widget);
     }
 
     public ArrayList<Avu> getMetadataFromTemplate() {
-        ArrayList<Avu> avus =new ArrayList<>();
-        for (String attr : templateAttrFieldMap.keySet()) {
-            Avu avu = templateAttrAvuMap.get(attr);
+        ArrayList<Avu> avus = new ArrayList<>();
```
ash3rz: I think ideally this logic should be in the presenter and tested?  That would eliminate the need to call `MetadataPresenterImpl.newMetadata()` and it would "dumb down" the view.

de-lib/src/main/java/org/iplantc/de/diskResource/client/views/metadata/dialogs/MetadataTemplateViewDialog.java
```diff
@@ -100,45 +111,63 @@ public MetadataTemplateViewDialog(MetadataView.Presenter presenter,List<Avu> tem
...
+        return valid;
     }
-    
+
     private void buildAvuMap() {
-    	for (MetadataTemplateAttribute attribute: attributes) {
-    		Avu md = findMetadataForAttribute(attribute.getName());
-    		templateAttrAvuMap.put(attribute.getName(),md);
-    	}
+        for (MetadataTemplateAttribute attribute : attributes) {
```
ash3rz: Same for this method, it has enough logic that should probably go in the presenter and be tested.

de-lib/src/main/java/org/iplantc/de/diskResource/client/views/metadata/dialogs/MetadataTemplateViewDialog.java
```diff
@@ -100,45 +111,63 @@ public MetadataTemplateViewDialog(MetadataView.Presenter presenter,List<Avu> tem
...
-    	if(templateMd != null) {
-	    	for(Avu md: templateMd) {
-	    		if(md.getAttribute().equals(attribute)) {
-	    			return md;
-	    		}
-	    	}
-    	}
-		return null;
+
+    private List<Avu> findMetadataForAttribute(String attribute) {
```
ash3rz: Another method that might be better suited in the presenter.

de-lib/src/main/java/org/iplantc/de/diskResource/client/views/metadata/dialogs/MetadataTemplateViewDialog.java
```diff
@@ -49,45 +51,54 @@
...
     private final MetadataView.Presenter presenter;
     private VerticalLayoutContainer widget;
     private final DateTimeFormat timestampFormat;
     private boolean writable;
-    private final FastMap<Avu> templateAttrAvuMap;
-    private final FastMap<Field<?>> templateAttrFieldMap;
-    private List<MetadataTemplateAttribute> attributes;
-    private List<Avu> templateMd;
+    FastMap<Avu> templateTagAvuMap;
+    FastMap<MetadataTemplateAttribute> templateTagAtrrMap;
```
ash3rz: Small thing - this should probably be templateTag**Attr**Map

de-lib/src/main/java/org/iplantc/de/diskResource/client/views/metadata/dialogs/MetadataTemplateViewDialog.java
```diff
@@ -338,28 +465,17 @@ public String getLabel(TemplateAttributeSelectionItem item) {
...
-                                        new VerticalLayoutContainer.VerticalLayoutData(.90, -1));
-                        }
-                        Dialog w = new Dialog();
-                        w.setHideOnButtonClick(true);
-                        w.setSize("350", "400");
-                        w.setPredefinedButtons(PredefinedButton.OK);
-                        w.setHeading(MetadataTemplateViewDialog.this.getHeader().getText());
-                        w.setBodyStyle("background: #fff;");
-                        w.setWidget(helpVlc);
+                        MetadataTermGuideDialog w = new MetadataTermGuideDialog(attributes,
```
ash3rz: Since we're keeping split points in mind, it might be good to refactor this sort of like [AppDocMarkdownDialog](https://github.com/cyverse-de/ui/commit/4152726f27ed38aa26f5b27c8298ab8421ccff16)

de-lib/src/main/java/org/iplantc/de/diskResource/client/views/metadata/dialogs/MetadataTermGuideDialog.java
```diff
@@ -0,0 +1,42 @@
...
+        setWidget(buildGuide(attributes));
+        show();
+    }
+
+    private VerticalLayoutContainer buildGuide(List<MetadataTemplateAttribute> attributes) {
+        VerticalLayoutContainer helpVlc = new VerticalLayoutContainer();
+        helpVlc.setScrollMode(ScrollSupport.ScrollMode.AUTOY);
+        for (MetadataTemplateAttribute mta : attributes) {
+            HTML l = new HTML("<b>" + mta.getName() + ":</b> <br/>");
+            HTML helpText = new HTML("<p>" + mta.getDescription() + "</p><br/>");
```
ash3rz: I would probably put `l` and `helpText` values in the appearance class.

de-lib/src/main/java/org/iplantc/de/theme/base/client/diskResource/metadata/MetadataViewDefaultAppearance.java
```diff
@@ -254,4 +255,23 @@ public ImageResource saveToFileIcon() {
...
         return toolbarMessages.saveMetadataMenuItem();
     }
+    
+    public String loading() {
+        return iplantDisplayStrings.loadingMask();
+    }
+
+    @Override
+    public String backgroudStyle() {
+        return "background:#fff;";
```
ash3rz: I didn't realize you could do this!  Not sure if it's simpler this way or if it'd be better to put in a GSS file?

de-lib/src/test/java/org/iplantc/de/diskResource/client/views/metadata/dialogs/MetadataTemplateViewDialogTest.java
```diff
@@ -0,0 +1,115 @@
...
+        Field f1 = mock(Field.class);
+        Field f2 = mock(Field.class);
+        Field f3 = mock(Field.class);
+
+        HashSet<String> tagSet = new HashSet<>();
+        tagSet.add(tag1);
+        tagSet.add(tag2);
+        tagSet.add(tag3);
+
+        when(templateTagAtrrMapMock.keySet()).thenReturn(tagSet);
```
ash3rz: Yay tests!  I think I'd prefer the logic to be in the presenter instead of the dialog, but I like that we at least have tests even though it's in a dialog.

de-lib/src/main/java/org/iplantc/de/diskResource/client/views/metadata/dialogs/MetadataTemplateViewDialog.java
sri2k1us: I will summarize my response here for above comments about moving some code to Presenter.  I feel they should stay here because all these logic is about managing / manipulating the template view. The `FastMap`s in this class is just to manage how attributes, fields and values are linked. Also, all these values does not get saved to the server from here. Rather they get loaded into the grid view through the presenter.

de-lib/src/main/java/org/iplantc/de/diskResource/client/views/metadata/dialogs/MetadataTemplateViewDialog.java
sri2k1us: 👍

de-lib/src/main/java/org/iplantc/de/theme/base/client/diskResource/metadata/MetadataViewDefaultAppearance.java
sri2k1us: Am not sure either. But I will check.

de-lib/src/main/java/org/iplantc/de/diskResource/client/views/metadata/dialogs/MetadataTemplateViewDialog.java
sri2k1us: 👍

de-lib/src/main/java/org/iplantc/de/theme/base/client/diskResource/metadata/MetadataViewDefaultAppearance.java
```diff
@@ -254,4 +255,23 @@ public ImageResource saveToFileIcon() {
     public String saveMetadataToFile() {
         return toolbarMessages.saveMetadataMenuItem();
     }
+    
+    public String loading() {
+        return iplantDisplayStrings.loadingMask();
+    }
+
+    @Override
+    public String backgroudStyle() {
```
ianmcorvidae: Not sure if this is an issue (probably not if everything uses the same name), but should this be `backgroundStyle`?

sri2k1us: Sigh! Another typo. I will fix it.

de-lib/src/main/java/org/iplantc/de/diskResource/client/views/metadata/dialogs/MetadataTemplateViewDialog.java
ash3rz: That's fair.  It's not at all critical for this PR, but I think I was leaning towards kind of what we do with the liststores. The presenters (for the most part) manipulate the data in the liststores so the view only has to worry about displaying the information, not manipulating it.  So in this case, I was envisioning the presenter would create and then pass in the FastMaps to the view.  Then for `getMetadataFromTemplate`, instead of the presenter calling that method which then has to call the `newMetadata()` method from the presenter and then returning the value back, it could all be in the presenter for example.  
Again, not critical for this PR, and I do see your point about wanting to keep view-related things in the view.

sri2k1us:
Merging this PR for QA drop tomorrow. Will submit a fresh PR with refactoring based on comments.

de-lib/src/main/java/org/iplantc/de/diskResource/client/views/metadata/dialogs/MetadataTemplateViewDialog.java
psarando: These are both good points. One thing to keep in mind when trying to decide if something should go into a view vs. a presenter is that if we switch the view to something else, ideally we wouldn't have to change anything in the main Metadata presenter. So in that respect, these FastMaps may belong to the view, since if we port the view to something like React, then maybe FastMaps and ListStores would no longer be required.

On the other hand, it may make porting the view to something else easier if the view was simple and dumb. So in that case, maybe it would help if this view had its own dialog-preseneter for manipulating the FastMaps and creating new AVUs, then as @ash3rz said, the view would only have to be concerned with displaying the models its given and calling events back to its presenter (which could in turn coordinate with the main presenter). Then if this view is ported to something else, its dialog-preseneter may or may not still be required.

These are just some general design ideas I wanted to add to this discussion, and not specific to the changes for this PR.

PR #104, opened 2017-03-27
CORE-6049 Fix contextual hep for `Send to Genome Browser` dialog.

sri2k1us:
- And Some code formatting.


de-lib/src/main/java/org/iplantc/de/theme/base/client/fileViewers/callbacks/ShareAnonymousCallbackDefaultAppearance.java
```diff
@@ -9,15 +9,14 @@
...
-import com.google.gwt.safehtml.shared.SafeUri;
 
 /**
  * @author jstroot
  */
 public class ShareAnonymousCallbackDefaultAppearance implements ShareAnonymousCallback.ShareAnonymousCallbackAppearance {
     interface EnsemblPopupTemplate extends SafeHtmlTemplates {
-        @SafeHtmlTemplates.Template("{0}<img src='{1}' qtip='{2}'></img>")
-        SafeHtml notificationWithContextHelp(SafeHtml label, SafeUri img, String toolTip);
+        @SafeHtmlTemplates.Template("{0}")
```
ash3rz: Since this doesn't have HTML anymore, does it still need to be a template?

sri2k1us: Good point. Although it doesn't hurt anything we can get away with just a text. I will push that change.

PR #108, opened 2017-03-30
Pipeline fix

sri2k1us:
- CORE-4419 fix button state.
- CORE-6741 fix step number when removing an app.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/pipelines/client/presenter/PipelineViewPresenter.java
+++ b/de-lib/src/main/java/org/iplantc/de/pipelines/client/presenter/PipelineViewPresenter.java
@@ -464,6 +464,7 @@ public class PipelineViewPresenter implements Presenter,
 
         if (selectedApp != null) {
             ListStore<PipelineTask> store = view.getPipelineAppStore();
+            store.remove(selectedApp);
             for (int step = store.size(); step > 0; step--) {
                 PipelineTask app = store.get(step - 1);
                 app.setStep(step - 1);
--- a/de-lib/src/main/java/org/iplantc/de/pipelines/client/presenter/PipelineViewPresenter.java
+++ b/de-lib/src/main/java/org/iplantc/de/pipelines/client/presenter/PipelineViewPresenter.java
@@ -464,7 +464,6 @@ public class PipelineViewPresenter implements Presenter,
 
         if (selectedApp != null) {
             ListStore<PipelineTask> store = view.getPipelineAppStore();
-            boolean removed = store.remove(selectedApp);
             for (int step = store.size(); step > 0; step--) {
                 PipelineTask app = store.get(step - 1);
                 app.setStep(step - 1);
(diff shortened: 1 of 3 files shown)


de-lib/src/main/java/org/iplantc/de/pipelines/client/views/PipelineAppOrderViewImpl.java
```diff
@@ -115,6 +141,21 @@ public PipelineTask getOrderGridSelectedApp() {
...
+            if(appOrderGrid.getStore().size() > 1) {
+                moveDownBtn.enable();
+                moveUpBtn.enable();
+            }
+        } else {
+            removeAppBtn.disable();
+            moveUpBtn.disable();
+            moveDownBtn.disable();
+        }
+    }
```
dennisroberts71: This may just be my own ignorance, but does this mean that `moveDownBtn` and `moveUpBtn` will be enabled any time more than one app is included in the pipeline? If that's the case, should the buttons be disabled unless an app is currently selected? Similarly, should only one of those buttons be enabled if the first or last app in the list is selected?

psarando: And that's why I just left those buttons always enabled in the first place :rofl:

sri2k1us: ` if(size > 0)` checks if any app is selected. Good point - I missed checks for first app / last app selection. Let me add that.

dennisroberts71: Ah, I missed that `size` referred to the size of the selection model. I assume that this is the number of items currently selected in the list (which is at most one because this is a single-select list).

sri2k1us: ✔️

de-lib/src/main/java/org/iplantc/de/pipelines/client/presenter/PipelineViewPresenter.java
```diff
@@ -450,15 +464,12 @@ public void onRemoveAppClicked() {
...
             ListStore<PipelineTask> store = view.getPipelineAppStore();
-
-            store.remove(selectedApp);
-
-            for (int step = 1; step <= store.size(); step++) {
+            boolean removed = store.remove(selectedApp);
+            for (int step = store.size(); step > 0; step--) {
                 PipelineTask app = store.get(step - 1);
-                app.setStep(step);
+                app.setStep(step - 1);
```
psarando: So I'm guessing that app steps used to be 1-based, but now they are 0-based? So this is the fix for `CORE-6741 fix step number when removing an app`?

This seems like the only effective change here, since the for-loop looks logically the same to me.
Also, it looks like the `store.remove` call is getting stored in a boolean now, but it's not used?

sri2k1us: Ahh! I left debugging code around. Will remove it. As far as `step` number is concerned, it still starts from `1`. See `AppColumnModel.java`->` return object.getStep() + 1;`

PR #109, opened 2017-04-04
CORE-8701 Add app ID to App Editor window config to allow multiple ap…

ash3rz:
…p editor windows to be open at the same time.

Previously each App Editor window had a window config of just "APP_INTEGRATION".  So when a user tried to open another App Editor window for a different app, the window configs would be the same, so the desktop manager assumed the user wanted to open a window that was already open.  Now the window config appends the system and app IDs so multiple windows can be open.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/commons/client/views/window/configs/ConfigFactory.java
+++ b/de-lib/src/main/java/org/iplantc/de/commons/client/views/window/configs/ConfigFactory.java
@@ -37,12 +37,15 @@ public class ConfigFactory {
     }
 
     public static AppsIntegrationWindowConfig appsIntegrationWindowConfig(HasQualifiedId app) {
-        AppsIntegrationWindowConfig aiwc = applyWindowType(WindowType.APP_INTEGRATION,
-                factory.appsIntegrationWindowConfig()).as();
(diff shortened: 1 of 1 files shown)


ash3rz:
I did some quick manual tests when I first created the PR, but I can go back and be a bit more thorough.  I'll post my findings!

ash3rz:
OK, I haven't totally nailed down exactly how to reproduce this, but if you have 2 App Editor windows open and then you start dragging and dropping items between the two, then it's possible that one app won't save because of a duplicate ID.  It doesn't happen all the time or every time, but it can happen.  If you don't DND between the two windows they seem to behave and save properly.

psarando:
OK, that's not so bad, but it also sounds like we should disable drag-n-drop between windows? Is there a simple way to disable that?

ash3rz:
I took a look and it seems like we're storing the original container/element where the DND originated and checking to make sure the target is within the same container.  Those checks should work to prevent DND between windows.  However, as soon as you drag an item to a secondary App Editor window, the container variable somehow updates to be the matching container in the new window.  The checks then always pass between windows.  So... I'll save that debug process for another day 😄 

Thanks for the review though!  I'll merge this now and open a ticket for the known issue.

PR #110, opened 2017-04-07
CORE-8709 Better error message handling when a user tries to save metadata

ash3rz:
When a user tried to save a metadata file with a name that already existed, the user was only informed that they should try again with no details as to why. Now if the file already exists, the user will be better informed, for any other errors they'll get the standard error handler message.  I also thought it would be better to unmask the Save dialog instead of hiding it so users don't have to type the file names in again.

I also added more key strings to the ErrorHandler class to parse for in the error JSON from what I've seen Agave use and examples in other tickets.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/grid/GridViewPresenterImpl.java
+++ b/de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/grid/GridViewPresenterImpl.java
@@ -111,8 +111,13 @@ public class GridViewPresenterImpl implements
 
         @Override
          public void onFailure(Throwable caught) {
-             save_dialog.hide();
-             announcer.schedule(new ErrorAnnouncementConfig("Unable to save your file. Please try again or contact support."));
+            save_dialog.unmask();
+             if (caught.getMessage().contains("ERR_EXISTS")) {
+                 announcer.schedule(new ErrorAnnouncementConfig(appearance.fileExistsError()));
+             } else {
+                 ErrorHandler.post("Unable to save your file. Please try again or contact support.",
+                                   caught);
+             }
          }
 
         @Override
--- a/de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/grid/GridViewPresenterImpl.java
+++ b/de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/grid/GridViewPresenterImpl.java
@@ -687,7 +687,7 @@ public class GridViewPresenterImpl implements
         saveAsDialogProvider.get(new AsyncCallback<SaveAsDialog>() {
             @Override
             public void onFailure(Throwable caught) {
-                announcer.schedule(new ErrorAnnouncementConfig("Unable to save your file. Please try again or contact support."));
+                ErrorHandler.post(caught);
             }
 
             @Override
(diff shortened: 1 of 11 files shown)


de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/grid/GridViewPresenterImpl.java
```diff
@@ -110,20 +110,23 @@ private SaveMetadataCallback(SaveAsDialog save_dialog) {
...
-             save_dialog.hide();
-             announcer.schedule(new ErrorAnnouncementConfig("Unable to save your file. Please try again or contact support."));
-         }
+        public void onFailure(Throwable caught) {
+            save_dialog.unmask();
+            if (caught.getMessage().contains("ERR_EXISTS")) {
+                announcer.schedule(new ErrorAnnouncementConfig(appearance.fileExistsError()));
+            } else {
+                ErrorHandler.post("Unable to save your file. Please try again or contact support.",
+                                  caught);
```
psarando: 👍

de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/grid/GridViewPresenterImpl.java
```diff
@@ -110,20 +110,23 @@ private SaveMetadataCallback(SaveAsDialog save_dialog) {
...
 
         @Override
-         public void onFailure(Throwable caught) {
-             save_dialog.hide();
-             announcer.schedule(new ErrorAnnouncementConfig("Unable to save your file. Please try again or contact support."));
-         }
+        public void onFailure(Throwable caught) {
+            save_dialog.unmask();
+            if (caught.getMessage().contains("ERR_EXISTS")) {
+                announcer.schedule(new ErrorAnnouncementConfig(appearance.fileExistsError()));
```
psarando: Is it possible to get the filename that the user entered from the `SaveAsDialog` here?

If so, then you could use the existing `org.iplantc.de.resources.client.messages.IplantErrorStrings#fileExists` error message, which takes a list of 1 or more file names to include with the message, instead of adding this new `fileExistsError` message.

Although I did notice that the `fileExists` message is inserting a `<br />` tag between the message and the filename(s), so I'm not sure how that would look with the `ErrorAnnouncementConfig`.

ash3rz: Ah, I saw `fileExists()`, but missed the other one with the file name in it.  I'll switch it over - good catch!

de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/grid/GridViewPresenterImpl.java
ash3rz: And I don't know why I spaced and left this string out of the appearance, I'll move that over too.

psarando: heh, I missed that too, I was just giving a 👍 for including the `Throwable` in the message 😁

ash3rz:
Sweet - ended up catching a few more things to refactor, so glad for the second pair of 👀 !

PR #111, opened 2017-04-11
CORE-8711 Groups UI - Listing Groups

ash3rz:
Here's the initial code just for listing Groups (nothing for creating/using/deleting yet). The button to view groups is currently not visible so it should be safe to merge.

I had to do a lot of refactoring for dependency injection in order to insert the GroupsListDialog into the collaborators view, so there are a few commits just for that.

NOTE: Will require an ansible config update prior to merging.

Here's where the button will be once visible:
image

The dialog itself will be pretty standard:
image


de-lib/src/main/java/org/iplantc/de/analysis/client/presenter/AnalysesPresenterImpl.java
```diff
@@ -431,22 +429,15 @@ public void onShowAllSelected() {
...
-        final AnalysisSharingPresenter sharingPresenter = new AnalysisSharingPresenter(analysisService,
-                                                                                 selected,
-                                                                                 sharingView,
-                                                                                 collaboratorsUtil);
-       aSharingDialogProvider.get(new AsyncCallback<AnalysisSharingDialog>() {
+        aSharingDialogProvider.get(new AsyncCallback<AnalysisSharingDialog>() {
             @Override
             public void onFailure(Throwable caught) {
-
+                ErrorHandler.post(caught);
```
ianmcorvidae: This probably isn't an issue per se, but this code can't really ever be run -- `AsyncProviderWrapper` doesn't use the `onFailure` at all. Not sure if this is one of the dialogs I switched over or one that did the same thing, but that's probably why it was initially an empty function.

de-lib/src/main/java/org/iplantc/de/desktop/client/gin/DEGinModule.java
```diff
@@ -2,10 +2,23 @@
 
 import org.iplantc.de.client.events.EventBus;
 import org.iplantc.de.client.gin.ServicesInjector;
-import org.iplantc.de.shared.DEProperties;
 import org.iplantc.de.client.models.UserInfo;
 import org.iplantc.de.client.models.UserSettings;
-import org.iplantc.de.client.services.*;
```
ianmcorvidae: Oh good, getting rid of `*`. I definitely approve this message! :smile:

de-lib/src/main/java/org/iplantc/de/analysis/client/presenter/AnalysesPresenterImpl.java
ash3rz: Yeah, it's more habit than anything else 😄   There are probably times where I leave it blank, and then circle back around thinking, "Oops, I missed this."

ash3rz:
Just to capture the discussion the team had in Slack -
It would probably be a better design to manage Collaborators and Groups within the same window.  Still on the fence about whether Collaborators and Groups should appear in the same list or in two separate lists that reside in the same window.
Also, some thought will need to be put into a feature where the list of all collaborators is also a group so it's easy for users to do something like "Share this with all my collaborators".

Regardless, there are some good dependency injection refactors in this PR that should still be merged in and will probably be required no matter the design direction.

ash3rz:
Thanks for the reviews!  I think I'll go ahead and merge this as-is, then make the changes as we discussed in our UI meetings.

PR #113, opened 2017-04-17
CORE-8701 Patch up NPE and allow multiple new app editor windows to b…

ash3rz:
…e open

I realized I introduced a NPE and also didn't factor in users who may want to open more than one **new** App Editor window instead of editing multiple current apps under development.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/commons/client/views/window/configs/ConfigFactory.java
+++ b/de-lib/src/main/java/org/iplantc/de/commons/client/views/window/configs/ConfigFactory.java
@@ -8,6 +8,7 @@ import org.iplantc.de.client.models.notifications.NotificationCategory;
 import org.iplantc.de.client.models.viewer.MimeType;
 import org.iplantc.de.client.services.FileEditorServiceFacade;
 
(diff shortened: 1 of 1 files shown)


de-lib/src/main/java/org/iplantc/de/commons/client/views/window/configs/ConfigFactory.java
```diff
@@ -39,11 +41,17 @@ public static AnalysisWindowConfig analysisWindowConfig() {
     public static AppsIntegrationWindowConfig appsIntegrationWindowConfig(HasQualifiedId app) {
         AutoBean<AppsIntegrationWindowConfig> aiwc = applyWindowType(WindowType.APP_INTEGRATION,
                 factory.appsIntegrationWindowConfig());
-        String systemId = app.getSystemId();
-        String appId = app.getId();
-        aiwc.as().setSystemId(app == null ? "" : systemId);
-        aiwc.as().setAppId(app == null ? "" : appId);
```
psarando: I totally overlooked these potential NPE's in my review too!

ash3rz:
Thanks for the review!

PR #116, opened 2017-04-18
CORE-8380 Configure UI for consul configs

ash3rz:


de.properties.tmpl
```diff
@@ -0,0 +1,95 @@
...
+{{ with $v := (key (printf "%s/terrain/base" $base)) }}org.iplantc.services.apps = {{ $v }}/apps{{ end }}
+{{ with $v := (key (printf "%s/terrain/base" $base)) }}org.iplantc.services.apps.ids = {{ $v }}/apps/ids{{ end }}
+{{ with $v := (key (printf "%s/terrain/base" $base)) }}org.iplantc.services.apps.categories = {{ $v }}/apps/categories{{ end }}
+{{ with $v := (key (printf "%s/terrain/base" $base)) }}org.iplantc.services.apps.hierarchies = {{ $v }}/apps/hierarchies{{ end }}
+{{ with $v := (key (printf "%s/terrain/base" $base)) }}org.iplantc.services.apps.elements = {{ $v }}/apps/elements{{ end }}
+{{ with $v := (key (printf "%s/terrain/base" $base)) }}org.iplantc.services.apps.elements.tools = {{ $v }}/apps/elements/tools{{ end }}
+{{ with $v := (key (printf "%s/terrain/base" $base)) }}org.iplantc.services.apps.elements.infoTypes = {{ $v }}/apps/elements/info-types{{ end }}
+{{ with $v := (key (printf "%s/terrain/base" $base)) }}org.iplantc.services.apps.elements.dataSources = {{ $v }}/apps/elements/data-sources{{ end }}
+{{ with $v := (key (printf "%s/terrain/base" $base)) }}org.iplantc.services.apps.argPreview = {{ $v }}/apps/arg-preview{{ end }}
```
ianmcorvidae: So just to clarify in text the bit I was poorly explaining in person: since `with` acts as an implicit `if` statement, if for some reason the terrain base wasn't in consul, this line would end up empty. Since from what you said the UI requires each thing to be there, it might make more sense to change:

```
{{ with $v := (key (printf "%s/terrain/base" $base)) }}org.iplantc.services.apps.argPreview = {{ $v }}/apps/arg-preview{{ end }}
```

to 

```
org.iplantc.services.apps.argPreview = {{ key (printf "%s/terrain/base" $base) }}/apps/arg-preview
```

Both will probably technically be wrong in the case of a missing variable, but the latter is probably actually easier to read in a case like this where there's no defaults in play to make it better to exclude the property entirely when it's not set.

ash3rz: Got it - let me see if I can push out a separate commit for these changes.

de.properties.tmpl
```diff
@@ -0,0 +1,95 @@
...
+org.iplantc.discoveryenvironment.local-ip-ranges = {{ key (printf "%s/de/local-ip-ranges" $base) }}
+org.iplantc.discoveryenvironment.environment.name = {{ key (printf "%s/de/environment-name" $base) }}
+org.iplantc.discoveryenvironment.environment.prod-deployment = {{ key (printf "%s/de/prod-deployment" $base) }}
+org.iplantc.discoveryenvironment.notifications.poll-interval = {{ key (printf "%s/de/poll-interval" $base) }}
+org.iplantc.discoveryenvironment.notification.amqp.uri = {{ key (printf "%s/de/amqp-de-uri" $base) }}
+org.iplantc.discoveryenvironment.notification.amqp.exchange.name = {{ key (printf "%s/de/amqp-de-name" $base) }}
+org.iplantc.discoveryenvironment.notification.amqp.exchange.durable = {{ key (printf "%s/de/amqp-de-durable" $base) }}
+org.iplantc.discoveryenvironment.notification.amqp.exchange.auto-delete = {{ key (printf "%s/de/amqp-de-auto-delete" $base) }}
+org.iplantc.discoveryenvironment.analysis.support.user = {{ key (printf "%s/de/support-user" $base) }}
+org.iplantc.discoveryenvironment.support = {{ key (printf "%s/terrain/base" $base) }}/support-email
```
psarando: Similar question as my review for cyverse-de/ansible#25: Will we need to update these new settings and the old `de.properties.j2` ansible template when we add or remove UI configs now?

de.properties.tmpl
```diff
@@ -0,0 +1,201 @@
...
+# is a tricky environment.
+###############################################################################
+org.iplantc.discoveryenvironment.about.defaultBuildNumber = {{ key (printf "%s/de/default-build-number" $base) }}
+org.iplantc.discoveryenvironment.about.releaseVersion = {{ key (printf "%s/de/release-version" $base) }}
+
+###############################################################################
+# CAS Authentication Settings
+###############################################################################
+org.iplantc.discoveryenvironment.cas.base-url = {{ key (printf "%s/de/base-url" $base) }}
+org.iplantc.discoveryenvironment.cas.server-name = {{ key (printf "%s/de/cas-server-name" $base) }}
```
ianmcorvidae: I think this one might be the one that could be changed to `/de/base` or something like that? we looked at it briefly while trying to fix the JSON stuff yesterday

ianmcorvidae: I guess those do actually come from separate ansible group_vars, but I don't know if we have different values for `de.base` and `de.app_base` anywhere

ianmcorvidae: Yeah, never mind, they're totally different. I could swear there was something that was still duplicated though...

ianmcorvidae:
LGTM, aside from my own confusion about non-duplicates!

de.properties.tmpl
ash3rz: Yeah... I could've sworn you showed me a dupe too, but now it seems to not be?

ash3rz:
Cool - thanks for all your help on this!  Time to see if it breaks stuff.

PR #119, opened 2017-04-26
CORE-8751 Allow users to create Collaborator Lists

ash3rz:
This is just for creating a Collaborator List (a group).  There's a separate endpoint for adding members to a group, so that will be in a separate PR to match.  I also went ahead and gutted the `CollaboratorUtil` class so that presenters are calling the service facade instead of using the util class to do it for them.  There's probably more refactoring that could be done there, but I'll save that for a different PR.

image

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/gin/BelphegorAppsGinModule.java
+++ b/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/gin/BelphegorAppsGinModule.java
@@ -34,6 +34,7 @@ import org.iplantc.de.client.events.EventBus;
 import org.iplantc.de.client.gin.ServicesInjector;
 import org.iplantc.de.client.models.UserInfo;
 import org.iplantc.de.client.models.UserSettings;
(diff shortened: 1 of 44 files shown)


cyverse-org:
Hi Ashely,
It would be good to allow users to designate a "manager" for the group.
This would allow someone else to manage the list.

I know Sriram is setting up a "meeting of minds" for Grouper. So we can
scope things.

Regards,
Nirav

ash3rz:
Hey Nirav, yes, we definitely need a "meeting of the minds".  We've had our own discussions and there are lots of different ideas about "groups".  I've captured most of those ideas/conversations in a [google doc](https://docs.google.com/document/d/1ofKeeKCM-RhHEgS3xvrNTpeVXzCASCSeZsOhHPZxnMY/edit#heading=h.y0nym1ajdr8b).  But basically this PR is the simplest idea to start with groups - you can think of it like creating a collection of contacts in GMail which is exclusive and private to you and only you.  This is useful if you regularly have data that you want to share with the same group of collaborators.  Instead of having to share data with each person one by one, you can share once with your custom list.

ash3rz:
Thanks for the review!

PR #120, opened 2017-04-26
CORE-8752 Allow users to delete Collaborator Lists

ash3rz:
These are the changes for deleting a Collaborator List (group).


tedgin:
The logic looks tight.  But since we talked about documenting in chat and everyone seemed to like the idea, do you mind documenting the newly created interface methods? :D

ash3rz:
I think it's a good idea to start the habit now - sure, will do!  And thanks for the review!

tedgin:
LGTM :shipit:

ash3rz:
Thanks for the review!

PR #121, opened 2017-05-02
CORE-8794 Allow users to list the members in a Collaborator List

ash3rz:
In case it's easier to review separately, the first set of commits are adding the logic for listing members in a Collaborator List.  The last commit is a bit of refactoring - I basically decided to add a presenter specifically for the `GroupDetailsView` so events don't have to trickle all the way back to `ManageCollaboratorsPresenter`.


ash3rz:
In case anybody is reviewing, I decided to make some refactors to this.  Should be ready by EOB.

PR #122, opened 2017-05-03
CORE-8801 Update DE logo and banner to be more consistent with atmo

ash3rz:
Changes towards a more consistent look between DE and Atmo.

image

@ianmcorvidae Maybe you can do the image compression magic you always do?


de-lib/src/main/java/org/iplantc/de/theme/base/client/desktop/BaseDesktopAppearance.java
```diff
@@ -50,7 +50,7 @@
         @Source("org/iplantc/de/theme/base/client/desktop/header_bg.png")
         ImageResource headerBg();
 
-        @Source("org/iplantc/de/theme/base/client/desktop/mini_logo.png")
```
ianmcorvidae: I guess this image is used elsewhere? Otherwise the source file should probably be removed, in any case.

de-lib/src/main/java/org/iplantc/de/theme/base/client/desktop/Desktop.gss
```diff
@@ -168,7 +172,7 @@
     gwt-sprite: "dataBtn";
     height: WIN_BTN_HEIGHT !important;
     width: WIN_BTN_WIDTH !important;
-    margin-top: WIN_BTN_MARGIN_TOP;
```
ianmcorvidae: What's this change? I guess the margin needed to be bigger given some sort of margin collapsing, or?

ianmcorvidae:
https://github.com/ash3rz/ui/pull/4 there's the recompression part :)

de-lib/src/main/java/org/iplantc/de/theme/base/client/desktop/BaseDesktopAppearance.java
ash3rz: Good question - let me check.  If not, I'll remove it.

de-lib/src/main/java/org/iplantc/de/theme/base/client/desktop/Desktop.gss
ash3rz: Yeah, making the banner bigger meant that I had to push the Data icon down further, otherwise it was only a few pixels below the banner which looked crowded.  The other icons then are spaced evenly away from the Data icon.

ash3rz:
Thanks for the review and image recompression!

PR #123, opened 2017-05-05
CORE-8792 Allow users to add members to a Collaborator List

ash3rz:
NOTE: This is rebased on top of CORE-8794 PR #121.  You only need to view the commits starting from `CORE-8792 Add endpoint for adding group members`

I had to add the concept of a mode to the GroupDetailsView for this to work - basically if you're creating a brand new Collaborator List, you're in the `ADD` mode which makes sure that the list and any members you add aren't saved until you complete the form - first the list gets saved, then the members get added.  If you're editing a list, you're in the `EDIT` mode and members get immediately added as you select them.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/collaborators/client/presenter/GroupDetailsPresenterImpl.java
+++ b/de-lib/src/main/java/org/iplantc/de/collaborators/client/presenter/GroupDetailsPresenterImpl.java
@@ -10,6 +10,7 @@ import org.iplantc.de.collaborators.client.events.AddGroupMemberSelected;
 import org.iplantc.de.collaborators.client.events.GroupSaved;
 import org.iplantc.de.commons.client.ErrorHandler;
 
+import com.google.common.base.Strings;
 import com.google.common.collect.Lists;
 import com.google.gwt.event.shared.HandlerManager;
 import com.google.gwt.event.shared.HandlerRegistration;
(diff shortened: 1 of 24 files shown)


de-lib/src/main/java/org/iplantc/de/collaborators/client/views/GroupViewImpl.java
```diff
@@ -122,10 +127,11 @@ public void onFailure(Throwable caught) {}
             @Override
             public void onSuccess(GroupDetailsDialog result) {
                 result.show();
```
sri2k1us: I was under the impression that launching new views/dialogs are done from presenter ?

de-lib/src/main/java/org/iplantc/de/collaborators/client/presenter/GroupDetailsPresenterImpl.java
```diff
@@ -0,0 +1,118 @@
...
+        this.factory = factory;
+        this.appearance = appearance;
+    }
+
+    @Override
+    public void go(HasOneWidget container, Group group) {
+        container.setWidget(view);
+
+        if (group != null) {
+            serviceFacade.getMembers(group, new AsyncCallback<List<Collaborator>>() {
```
ash3rz: 👍   makes sense - I'll refactor

de-lib/src/main/java/org/iplantc/de/collaborators/client/views/GroupViewImpl.java
ash3rz: Good question.  It looks like it's mixed right now whether they're launched from the view or the presenter, but also looks like the majority are from the presenter.  I can refactor for consistency.

ash3rz:
Thanks for the review!  I'll merge in the previous PR and then merge in this one.

PR #124, opened 2017-05-05
CORE-8812 Allow users to edit and update a Group

ash3rz:


ash3rz:
For my own sanity, I'm going to go ahead and merge this so I don't have to maintain as many branches on top of branches. Reviews still welcome.

PR #126, opened 2017-05-09
Minor fixes for Collaborator Lists

ash3rz:
These are minor fixes after some testing with the new terrain endpoints for Collaborator Lists.


ash3rz:
These are pretty small changes, I think I'll go ahead and merge this in.  Still welcome comments though.

PR #128, opened 2017-05-16
CORE-8835 Update sharing requests in the UI to match the updated endpoints

ash3rz:
This is a bit hacky and hard-coded for now just to get sharing working again for this release.  Things will be properly updated and/or refactored once the search and sharing logic are updated to use Subjects instead of Collaborators and the service calls get updated.


de-lib/src/main/java/org/iplantc/de/apps/client/presenter/sharing/AppSharingPresenter.java
```diff
@@ -222,14 +225,17 @@ private AppUnSharingRequestList buildUnSharingRequest() {
...
             for (String userName : unSharingMap.keySet()) {
                 List<Sharing> shareList = unSharingMap.get(userName);
 
-                AppUnsharingRequest unsharingRequest = appFactory.appUnSharingRequest().as();
-                unsharingRequest.setUser(userName);
-                unsharingRequest.setApps(buildUnshareAppPermissionList(shareList));
+                AppSharingRequest unsharingRequest = appFactory.appSharingRequest().as();
+                SharingSubject sharingSubject = shareFactory.getSharingSubject().as();
+                sharingSubject.setId(userName);
+                sharingSubject.setSourceId("ldap");
```
psarando: I see what you mean by this PR's description 😄 
I'm guessing we will eventually want a util method somewhere to convert a username to a `SharingSubject`?

ash3rz: Yeah, something like that.  I just wanted the smallest amount of code changes that will actually be used in this release since we're so close to the end. 😄

ash3rz:
Thanks for the review!

PR #129, opened 2017-05-17
CORE-8753 Allow users to search for Groups in the UI

ash3rz:
NOTE: Requires ansible changes before merging.  Also, since this has a lot of refactoring and uses the new `GET /subjects` endpoint, it might be best to save this for next release.

Basically, the UserSearchField now uses the iplant-groups `GET /subjects` endpoint which will return a list of Subjects - a subject can be a Group/Collaborator List or a user.  Lists are distinguished with an icon:

image

For this PR, I refactored UserSearchField for dependency injection and also to use an appearance class.  I also renamed the UI's model for `Collaborator` to `Subject` and renamed any other classes or method names that should change along with it (most of the file changes). There are probably other opportunities for refactoring with the switch from `Collaborator` to `Subject`, but I'll save those for a different PR.  

iplant-groups cannot currently supply a user's list of collaborators that do not belong to a group, so some of the old endpoints are still in use and where needed use an autobean called `OldCollaborator`.  Those endpoints will be updated and the `OldCollaborator` model will be removed once iplant-groups can supply the old information.


de-lib/src/main/java/org/iplantc/de/client/models/collaborators/OldCollaborator.java
```diff
@@ -9,9 +9,9 @@
 
 /**
  * @author sriram
- * 
+ *
  */
-public interface Collaborator extends HasId {
+public interface OldCollaborator extends HasId {
```
sri2k1us: I might be missing something but is there a reason to keep the `OldCollaborator` around ?

de-lib/src/main/java/org/iplantc/de/collaborators/client/views/GroupDetailsViewImpl.java
```diff
@@ -118,12 +100,20 @@ public void onSelectionChanged(SelectionChangedEvent<Collaborator> event) {
...
-        handlerRegistration =
-                eventBus.addHandler(UserSearchResultSelected.TYPE, new CollaboratorSelectedHandler());
-
         editorDriver.initialize(this);
     }
 
+    @Override
+    public void onUserSearchResultSelected(UserSearchResultSelected userSearchResultSelected) {
+        Collaborator collaborator = userSearchResultSelected.getCollaborator();
+        if (MODE.EDIT == mode) {
```
sri2k1us: On `enum` you can use `.equals()`

de-lib/src/main/java/org/iplantc/de/client/models/collaborators/OldCollaborator.java
ash3rz: Yes, we still have a few endpoints that the UI needs that is still returning the old `Collaborator` format which has different keys than a `Subject`.  For example, iplant-groups doesn't currently have an endpoint which will return a user's list of plain old Collaborators that do NOT belong to a list.  So we're still using the old apps endpoint which has the old format.

ash3rz:
Thanks for the review! 👍   Since this transitions some of the current UI (not just the hidden Collaborator List UI stuff) to using the new iplant-groups endpoints, I'll merge this after the release is done so it can have more time to be tested by QA.

ash3rz:
Now that the release and whatnot is done, I'll start merging my PRs in.

PR #130, opened 2017-05-20
CORE-8845 Add deprecated tool warning when editing private apps.

psarando:
If the user opens (or copies) an app for editing, and that app uses a tool that has been marked as `deprecated`, then the following warning message will display in a dialog pop-up:

> This app uses a deprecated tool. You can still edit and use this private app, but you will have to choose a newer version of the tool before it can be made public.

This message will not display when a user opens a public app to edit its labels.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/apps/integration/client/presenter/AppsEditorPresenterImpl.java
+++ b/de-lib/src/main/java/org/iplantc/de/apps/integration/client/presenter/AppsEditorPresenterImpl.java
@@ -430,10 +430,12 @@ public class AppsEditorPresenterImpl implements AppsEditorView.Presenter,
     public void go(HasOneWidget container) {
         clearRegisteredHandlers();
 
-        if(appTemplate.isPublic()!=null) {
-            setLabelOnlyEditMode(appTemplate.isPublic());
-        } else {
-            setLabelOnlyEditMode(false);
+        boolean isPublic = appTemplate.isPublic() != null ? appTemplate.isPublic() : false;
+
+        setLabelOnlyEditMode(isPublic);
+
+        if (!isPublic) {
+            checkForDeprecatedTools(appTemplate.getTools());
         }
 
         view.getEditorDriver().edit(appTemplate);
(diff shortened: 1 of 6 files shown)


de-lib/src/main/java/org/iplantc/de/apps/integration/client/presenter/AppsEditorPresenterImpl.java
```diff
@@ -430,10 +430,12 @@ public AppTemplate getAppTemplate() {
     public void go(HasOneWidget container) {
         clearRegisteredHandlers();
 
-        if(appTemplate.isPublic()!=null) {
-            setLabelOnlyEditMode(appTemplate.isPublic());
-        } else {
-            setLabelOnlyEditMode(false);
+        boolean isPublic = appTemplate.isPublic() != null ? appTemplate.isPublic() : false;
+
+        setLabelOnlyEditMode(isPublic);
```
dennisroberts71: ✨

de-lib/src/main/java/org/iplantc/de/apps/integration/client/presenter/AppsEditorPresenterImpl.java
```diff
@@ -814,4 +816,16 @@ void clearRegisteredHandlers(){
         }
     }
 
+    private void checkForDeprecatedTools(final List<Tool> tools) {
+        if (tools != null && !tools.isEmpty()) {
+            if (tools.stream().anyMatch(Tool::isDeprecated)) {
```
dennisroberts71: I still need to play around with streams. This code is nice and tidy. ✨

psarando:
Thanks for the quick review! 👍

ash3rz:
LGTM too 👍

de-lib/src/main/java/org/iplantc/de/apps/integration/client/presenter/AppsEditorPresenterImpl.java
sri2k1us: Neat! :1st_place_medal:

PR #131, opened 2017-05-23
CORE-8795 Allow users to share resources with a Group

ash3rz:
NOTE: This is a branch on top of another branch.  The commits start at `CORE-8795 Add concept of Mode to GroupView to manage ManageCollaboratorsView mode`. You can also use [this link](https://github.com/cyverse-de/ui/pull/131/files/2b53938c3e812007d6ec39607d15c68c1541be4f..1ffed07f5b3164b0259cf2136895910e2d3fa044) to only view those commits.

This PR should allow users to share apps/analyses/data with Collaborators and/or Collaborator Lists.  Places where "ldap" is hard-coded in the `source_id` field should be updated after the endpoints get updated to return `Subject` types instead of the old `Collaborator` types.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/client/models/collaborators/Subject.java
+++ b/de-lib/src/main/java/org/iplantc/de/client/models/collaborators/Subject.java
@@ -2,6 +2,8 @@ package org.iplantc.de.client.models.collaborators;
 
 import org.iplantc.de.client.models.HasSettableId;
 
+import com.google.gwt.regexp.shared.MatchResult;
+import com.google.gwt.regexp.shared.RegExp;
 import com.google.gwt.user.client.ui.HasName;
 import com.google.web.bindery.autobean.shared.AutoBean;
 
@@ -9,9 +11,14 @@ import java.util.List;
 
 /**
  * The autobean representation of a Grouper subject
+ * Also some helper methods to determine if the subject is a collaborator list and how
+ * to get the display name
  * @author aramsey
  */
 public interface Subject extends HasSettableId, HasName {
+
+    String GROUP_IDENTIFIER = "g:gsa";
+    String GROUP_LONG_NAME_REGEX = ".*collaborator-lists:(.+)";
 
     @AutoBean.PropertyName("first_name")
     String getFirstName();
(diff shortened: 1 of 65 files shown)


de-lib/src/main/java/org/iplantc/de/client/models/collaborators/Subject.java
```diff
@@ -0,0 +1,94 @@
...
+    String getSourceId();
+
+    @AutoBean.PropertyName("source_id")
+    void setSourceId(String sourceId);
+
+    default boolean isCollaboratorList() {
+        return getSourceId().equals(GROUP_IDENTIFIER);
+    }
+
+    default String getSubjectDisplayName() {
```
ash3rz: Default methods are a new thing in Java 8... not sure how I feel about having the logic here, but thought I'd try it out.

de-lib/src/main/java/org/iplantc/de/client/sharing/SharingPermissionsPanel.java
```diff
@@ -295,7 +266,8 @@ public void loadSharingData(FastMap<List<Sharing>> sharingMap) {
         }
     }
 
-    private ColumnModel<Sharing> buildColumnModel() {
+    @UiFactory
+    ColumnModel<Sharing> createColumnModel() {
         List<ColumnConfig<Sharing, ?>> configs = new ArrayList<>();
         DataSharingProperties props = GWT.create(DataSharingProperties.class);
```
sri2k1us: This can be injected!

de-lib/src/main/java/org/iplantc/de/client/models/collaborators/Subject.java
sri2k1us: 👍

de-lib/src/main/java/org/iplantc/de/client/sharing/SharingPermissionsPanel.java
ash3rz: Ah, good catch!

ash3rz:
Thanks for the review! 👍  I'll merge this PR after the release as well so QA can have more time with the changes.

ash3rz:
I'm going to merge this one now too - thanks for the reviews!

PR #133, opened 2017-05-23
Core 8737 Create tools listing user interface

sri2k1us:
This PR includes
- Events for various menu items
- Tool search
- Some appearance

PS. In the commit message, i accidentally put the wrong jira# (core-8437 instead of core-8737)


de-lib/src/main/java/org/iplantc/de/apps/client/events/tools/ToolSearchResultLoadEvent.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public static final Type<ToolSearchResultLoadEventHandler> TYPE = new Type<>();
+    private final String searchPattern;
+    private final List<Tool> results;
+    private final String searchText;
+
+    public ToolSearchResultLoadEvent(final String searchText,
+                                    final String searchPattern,
+                                    final List<Tool> results) {
+        Preconditions.checkNotNull(results);
```
ash3rz: I like the use of the Preconditions class 👍   If that condition fails though, it will throw an error which may or may not be what you want for some of the use cases in this PR.  Just throwing that out there.

de-lib/src/main/java/org/iplantc/de/apps/client/events/tools/ToolSelectionChangedEvent.java
```diff
@@ -0,0 +1,52 @@
...
+        void onToolSelectionChanged(ToolSelectionChangedEvent event);
+    }
+
+    public static interface HasToolSelectionChangedEventHandlers {
+        HandlerRegistration addToolSelectionChangedEventHandler(ToolSelectionChangedEventHandler handler);
+    }
+
+    public static final Type<ToolSelectionChangedEventHandler> TYPE = new Type<>();
+
+    private final List<Tool> ToolSelection;
```
ash3rz: Another variable that should have a lower case first letter.

de-lib/src/main/java/org/iplantc/de/apps/client/views/tools/ManageToolsViewImpl.java
```diff
@@ -108,7 +119,24 @@ public String getPath() {
...
+            @Override
+            public String getPath() {
+                return null;
+            }
+        },50, appearance.tag());
+
+        ColumnConfig<Tool, String> status = new ColumnConfig<Tool, String>(new ValueProvider<Tool, String>() {
+            @Override
+            public String getValue(Tool object) {
+                return (object.isPublic() == true)? "Public" : object.getPermission();
```
ash3rz: This should probably go into the appearance class. Probably the other strings in the ColumnConfig getValue() methods should go into the appearance as well.

de-lib/src/main/java/org/iplantc/de/apps/client/views/tools/ManageToolsViewToolbarImpl.java
```diff
@@ -16,42 +23,198 @@
...
+                    break;
+                default:
+                    addTool.setEnabled(true);
+                    requestTool.setEnabled(true);
+                    edit.setEnabled(false);
+                    delete.setEnabled(false);
+                    useInApp.setEnabled(false);
+                    shareMenuButton.setEnabled(false);
+                    shareCollab.setEnabled(false);
+                    sharePublic.setEnabled(false);
```
ash3rz: For the default case, if a user has multiple tools selected, I think it might be nice to enable them to delete multiple tools at once instead of one at a time.
Also, we enable multiple apps/data to be shared at a time, we might want to do the same with tools.

de-lib/src/main/java/org/iplantc/de/apps/client/views/tools/ManageToolsViewToolbarImpl.java
```diff
@@ -16,42 +23,198 @@
...
+        }
+
+    }
+
+    private boolean isEditable(Tool tool) {
+        return !tool.isPublic() && (isOwner(tool) || hasWritePermission(tool));
+    }
+
+    private boolean isOwner(Tool tool) {
+        return tool.getPermission().equalsIgnoreCase("OWNER");
```
ash3rz: There's also the `PermissionValue` enum in `org.iplantc.de.client.models.diskResources`.  Can that be used here instead of strings?  Although, that set has `own` instead of `owner`.

de-lib/src/main/java/org/iplantc/de/apps/client/presenter/tools/ManageToolsViewPresenter.java
```diff
@@ -23,32 +24,41 @@
...
     public ManageToolsViewPresenter(ManageToolsView.ManageToolsViewAppearance appearance) {
         this.appearance = appearance;
     }
 
     @Override
     public void go(HasOneWidget container) {
-       container.setWidget(toolsView.asWidget());
-       toolsView.mask(appearance.mask());
-       loadTools();
+        container.setWidget(toolsView.asWidget());
```
ash3rz: I'm pretty sure you mentioned you're writing tests last, so I'll hold off on those comments 😜

de-lib/src/main/java/org/iplantc/de/apps/client/views/tools/ManageToolsViewToolbarImpl.java
```diff
@@ -16,42 +23,198 @@
...
     public ManageToolsViewToolbarImpl(final ManageToolsToolbarView.ManageToolsToolbarApperance apperance) {
-        loader = new PagingLoader<FilterPagingLoadConfig, PagingLoadResult<Tool>>(new ToolSearchRPCProxy()) ;
+        loader = new PagingLoader<FilterPagingLoadConfig, PagingLoadResult<Tool>>(toolSearchRPCProxy);
         toolSearch = new ToolSearchField(loader);
         initWidget(uiBinder.createAndBindUi(this));
+        toolSearchRPCProxy.setHasHandlers(this);
+    }
+
+    @Override
+    protected void onEnsureDebugId(String baseID) {
```
ash3rz: Yay static IDs! 😄

de-lib/src/main/java/org/iplantc/de/apps/client/views/tools/ManageToolsViewToolbarImpl.java
psarando: Also, the `/tools/permission-lister` endpoint documentation says the permission would be `own` as well. So it looks like we can use that `PermissionValue` enum here.

sri2k1us: Yes. I was meaning to refactor `PermissionValue` out of `diskResources` module into `commons` module. But forgot to get to it. Will makes those changes.

de-lib/src/main/java/org/iplantc/de/apps/client/views/tools/ManageToolsViewImpl.java
psarando: `object.isPublic() == true` is redundant. You could just check `object.isPublic() ?`

sri2k1us: 👍

de-lib/src/main/java/org/iplantc/de/apps/client/events/tools/DeleteToolSelected.java
```diff
@@ -0,0 +1,46 @@
...
+
+import com.google.common.base.Preconditions;
+import com.google.gwt.event.shared.EventHandler;
+import com.google.gwt.event.shared.GwtEvent;
+import com.google.gwt.event.shared.HandlerRegistration;
+
+import java.util.List;
+
+/**
+ * Created by jstroot on 3/5/15.
```
sri2k1us: Ya I copied most of events for an App and changed it for Tools :). I will get that updated.

PR #140, opened 2017-05-23
Core 8744 Implement `Share with public` for private tools

sri2k1us:
- Refactor New Tool Request form to use it for `Share with public`


de-lib/src/main/java/org/iplantc/de/tools/client/views/dialogs/NewToolRequestDialog.java
```diff
@@ -60,6 +59,24 @@ public void onSelect(SelectEvent event) {
 
     }
 
+    public void show(NewToolRequestFormView.Mode mode) {
+        presenter.setMode(mode);
+         switch (mode) {
+             case NEWTOOL:
+                 setHeading(I18N.DISPLAY.requestNewTool());
```
ash3rz: This heading and the one below should go into an appearance class

de-lib/src/main/java/org/iplantc/de/tools/client/presenter/NewToolRequestFormPresenterImpl.java
```diff
@@ -56,16 +56,19 @@
...
+    private Tool tool;
 
     @Inject
     NewToolRequestFormPresenterImpl(final NewToolRequestFormView view,
                                     @Assisted final Command callbackCmd) {
         this.view = view;
         this.callback = callbackCmd;
         view.setPresenter(this);
-        setToolMode(NewToolRequestFormView.SELECTION_MODE.UPLOAD);
+        setToolMode(NewToolRequestFormView.SELECTION_MODE.LINK);
```
psarando: 👍

PR #141, opened 2017-05-25
CORE-8858 Add deprecated tool warnings to workflow editor.

psarando:
This PR will update the `PipelineViewPresenter` to check for deprecated tools, using the details added by cyverse-de/apps#61, in order to display a warning when opening a workflow for editing that uses apps with `deprecated` tools, or when adding apps that use `deprecated` tools to a workflow.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/pipelines/client/presenter/PipelineViewPresenter.java
+++ b/de-lib/src/main/java/org/iplantc/de/pipelines/client/presenter/PipelineViewPresenter.java
@@ -9,10 +9,12 @@ import org.iplantc.de.client.models.apps.App;
 import org.iplantc.de.client.models.pipelines.Pipeline;
 import org.iplantc.de.client.models.pipelines.PipelineAppMapping;
 import org.iplantc.de.client.models.pipelines.PipelineTask;
+import org.iplantc.de.client.models.tool.Tool;
 import org.iplantc.de.commons.client.ErrorHandler;
 import org.iplantc.de.commons.client.info.ErrorAnnouncementConfig;
 import org.iplantc.de.commons.client.info.IplantAnnouncer;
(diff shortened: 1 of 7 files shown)


de-lib/src/main/java/org/iplantc/de/pipelines/client/presenter/PipelineViewPresenter.java
```diff
@@ -179,6 +208,22 @@ public void setPipeline(Splittable serviceWorkflowJson) {
...
     }
 
+    private void checkForDeprecatedTools(final List<PipelineTask> apps) {
+        if (apps != null && !apps.isEmpty()) {
+            if (apps.stream().anyMatch(app -> ((app.getTool() != null) && app.getTool()
+                                                                             .getContainer()
+                                                                             .getImage()
+                                                                             .isDeprecated()))) {
+                Scheduler.get().scheduleDeferred(() -> {
+                    IplantInfoBox errorsInfo = new IplantInfoBox(I18N.DISPLAY.warning(),
```
sri2k1us: I think we are moving away from using `118N` and use appearance instead!

de-lib/src/main/java/org/iplantc/de/pipelines/client/presenter/PipelineViewPresenter.java
```diff
@@ -179,6 +208,22 @@ public void setPipeline(Splittable serviceWorkflowJson) {
         setPipeline(pipeline);
     }
 
+    private void checkForDeprecatedTools(final List<PipelineTask> apps) {
+        if (apps != null && !apps.isEmpty()) {
+            if (apps.stream().anyMatch(app -> ((app.getTool() != null) && app.getTool()
```
ash3rz: ❇️  Yay Java 8!

psarando:
Thanks for the reviews 👍

PR #142, opened 2017-05-31
Core 8850 Display list of apps used by the selected tool in the tool info dialog

sri2k1us:
For this PR, it is enough to review commits with `core-8850` in the message.


de-lib/src/main/java/org/iplantc/de/tools/client/views/AppUsedByAToolView.java
```diff
@@ -0,0 +1,16 @@
+package org.iplantc.de.tools.client.views;
```
ash3rz: Was this class created for some later changes or is there something missing here?

de-lib/src/main/java/org/iplantc/de/tools/client/views/manage/ToolInfoViewImpl.java
```diff
@@ -0,0 +1,80 @@
...
+import com.sencha.gxt.widget.core.client.container.AbstractHtmlLayoutContainer;
+import com.sencha.gxt.widget.core.client.container.HtmlLayoutContainer;
+import com.sencha.gxt.widget.core.client.container.SimpleContainer;
+import com.sencha.gxt.widget.core.client.container.VerticalLayoutContainer;
+
+/**
+ *  A tab based view of tool information
+ * Created by sriram on 5/30/17.
+ */
+public class ToolInfoViewImpl extends Composite implements ToolInfoView {
```
ash3rz: 👍  Nice refactor!

de-lib/src/main/java/org/iplantc/de/tools/client/views/AppUsedByAToolView.java
sri2k1us: Ahh! This can be deleted!

de-lib/src/main/java/org/iplantc/de/tools/client/presenter/ManageToolsViewPresenter.java
```diff
@@ -308,16 +310,28 @@ public HandlerRegistration addSelectionChangedHandler(SelectionChangedEvent.Sele
 
     @Override
     public void onShowToolInfo(ShowToolInfoEvent event) {
-       toolInfoDialogProvider.get(new AsyncCallback<ToolInfoDialog>() {
-           @Override
-           public void onFailure(Throwable throwable) {
+        dcService.getAppsForTool(event.getTool().getId(), new AppsCallback<List<App>>() {
+
+            @Override
+            public void onFailure(Integer statusCode, Throwable exception) {
```
psarando: So it looks like if the tool's apps can't be loaded, then none of the tool details will display?
Maybe we can display an error message and display tool details with an empty apps list instead?

sri2k1us: What do you mean ? - An error with the service ? or there are no apps using the selected tool? The later case is handled.

sri2k1us: Ah! There is no error message in case the service call fails.

psarando: Error from the service. This is the `onFailure` for `dcService.getAppsForTool`, so it could be anything, like a 404 if the service can't find this tool-id for some reason, or if the response doesn't pass the format validation, etc.

sri2k1us: 👍

de-lib/src/main/java/org/iplantc/de/theme/base/client/apps/ToolMessages.properties
```diff
@@ -26,4 +26,5 @@ timeLimit = Time Limit
 create = Add Tool
 restrictions = Restrictions
 toolInfo = Tool Information
-confirmDelete = Do you really want to delete this tool?
\ No newline at end of file
+confirmDelete = Do you really want to delete this tool?
+appsLoadError = Unable to get the list of apps used this tool!
```
psarando: Should probably say "apps used `by` this tool"

PR #143, opened 2017-05-31
CORE-8872 Fix Class Cast Exception being thrown during DND within HT …

ash3rz:
…Files

The DND logic within PathListViewer assumed the user would only ever DND files from the Data window (with type DiskResource) to the HT File viewer (with type Splittable).  In the code, there was logic casting the DND source to DiskResource. This is not always the case as users may want to rearrange files within the HT file viewer, which was unaccounted for.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/fileViewers/client/views/PathListViewer.java
+++ b/de-lib/src/main/java/org/iplantc/de/fileViewers/client/views/PathListViewer.java
@@ -138,33 +138,48 @@ public class PathListViewer extends AbstractStructuredTextViewer implements Stor
                               CancellableEvent cancellableEvent,
                               StatusProxy statusProxy) {
 
-            if(!hasCorrectData(o)){
+            if(!isNonEmptyCollection(o)){
                 cancellableEvent.setCancelled(true);
                 statusProxy.setStatus(false);
+                return;
             }
-            // TODO Check to see if any items are pathlists. If they are, prevent drop
-            Iterable<DiskResource> iterable = (Iterable<DiskResource>) o;
-            for(DiskResource dr : iterable){
-                InfoType infoType1 = InfoType.fromTypeString(dr.getInfoType());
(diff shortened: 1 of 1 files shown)


de-lib/src/main/java/org/iplantc/de/fileViewers/client/views/PathListViewer.java
```diff
@@ -138,33 +138,48 @@ void handleDropStatus(Object o,
...
+                       && hasDiskResources(data);
+        }
+
+        boolean hasDiskResources (Object data) {
+            return ((Collection<?>) data).iterator().next() instanceof DiskResource;
+        }
+
+        boolean isNonEmptyCollection(Object data) {
+            boolean isCollection = data instanceof Collection<?>;
             boolean isEmpty = ((Collection<?>) data).isEmpty();
```
sri2k1us: Should this check be done only when `isCollection` is `true`? Otherwise, this can can cause a `ClassCastException`

ash3rz: Sure, let me change it

psarando:
This PR is going to merge to `master`.
Are we going to push this fix into prod?
Prod is currently at [2.11.0](https://github.com/cyverse-de/ui/tree/2.11.0)

ash3rz:
I was opening the PR just for review.  I think last time we did a hotfix we created a hotfix branch in the repo so we could test in de-2, then QA, then push to prod?

psarando:
OK, got it.
So we could have created the branch first and opened this PR and merged it against that branch as well.

ash3rz:
Ah right, sorry, habit.  OK, so I can close this PR, push my branch to origin, then re-open the PR... does that sound right?

psarando:
Well, we've already reviewed this PR, so I don't think another PR is needed.
We can just close this PR, then you can push your branch+commit directly to origin.

PR #146, opened 2017-06-07
CORE-8878 Add quick-add option to Collaborator Lists

ash3rz:
*NOTE: This is currently a branch on top of another branch, you only have to look at the last 2 commits*
Currently, to add a user to a collaborator list, you have to first open the list details, then search for the user to add them.
This PR allows users to simply have the list selected in the Collaborators window and whomever they search for will then get added to the list(s).  If the Collaborators tab is selected, the selected collaborator will be added to their default set of collaborators like normal.


de-lib/src/main/java/org/iplantc/de/collaborators/client/presenter/ManageCollaboratorsPresenter.java
```diff
@@ -112,14 +113,49 @@ public void onUserSearchResultSelected(UserSearchResultSelected userSearchResult
...
+            } else {
+                List<Group> selectedCollaboratorLists = view.getSelectedCollaboratorLists();
+                if (selectedCollaboratorLists != null && !selectedCollaboratorLists.isEmpty()) {
+                    addMemberToGroups(subject, selectedCollaboratorLists);
+                } else {
+                    announcer.schedule(new ErrorAnnouncementConfig(groupAppearance.noCollabListSelected()));
+                }
             }
         } else {
             announcer.schedule(new ErrorAnnouncementConfig(I18N.DISPLAY.collaboratorSelfAdd()));
```
sri2k1us: Appearance can be used instead of `I18N`

sri2k1us: This code could use some unit tests give the nested `if-else` structure.

de-lib/src/main/java/org/iplantc/de/collaborators/client/presenter/ManageCollaboratorsPresenter.java
```diff
@@ -112,14 +113,49 @@ public void onUserSearchResultSelected(UserSearchResultSelected userSearchResult
...
         }
     }
 
+    void addMemberToGroups(Subject subject, List<Group> selectedCollaboratorLists) {
+        selectedCollaboratorLists.forEach(new Consumer<Group>() {
+            @Override
+            public void accept(Group group) {
+                groupServiceFacade.addMembers(group, Lists.newArrayList(subject), new AsyncCallback<List<UpdateMemberResult>>() {
```
sri2k1us: Is there way to avoid `RPC` calls from a loop? or Somehow combine the results of all the calls and finally display one error message ? Throwing multiple errors may lead to unpleasant user experience.

de-lib/src/main/java/org/iplantc/de/collaborators/client/presenter/ManageCollaboratorsPresenter.java
ash3rz: Ah, not sure how I missed that I18N.  Good catch.  I'll also add tests.

de-lib/src/main/java/org/iplantc/de/collaborators/client/presenter/ManageCollaboratorsPresenter.java
ash3rz: Sure, I can do something similar to what I did in Apps for the hierarchies.

ash3rz:
Thanks for the review!

PR #149, opened 2017-06-08
CORE-8844 Allow users to de-select analyses in the analysis window

ash3rz:
This took me longer to find than I'd have liked, but basically we were injecting a ColumnModel that had a separate CheckBoxSelectionModel than the one created and used in the view.


sri2k1us:
LGTM as well. 
@psarando - Yes Metadata dialog has some special logic for checkbox selection since that grid needs to support editing as well.

ash3rz:
Cool - thanks for the reviews!

PR #150, opened 2017-06-08
CORE-8883 Prevent new users from seeing an error banner regarding preferences

ash3rz:
I also changed some wording.  If preferences failed to load during bootstrap, we retry fetching preferences when opening the Preferences window.  In the case of a new user, "restored" doesn't make much sense, but "loaded" should work for all scenarios.


ash3rz:
Thanks for the review!

PR #151, opened 2017-06-09
Core 8714 Add image name validator and set default image name prefix for admin

sri2k1us:
Note - It is enough to review just the last commit (core-8714 in message)


de-lib/src/main/java/org/iplantc/de/commons/client/validators/ImageNameValidator.java
```diff
@@ -0,0 +1,44 @@
...
+        this.validationMessages = validationMessages;
+    }
+
+    @Override
+    public List<EditorError> validate(Editor<String> editor, String value) {
+        if (Strings.isNullOrEmpty(value)) {
+            return createError(new DefaultEditorError(editor, validationMessages.imgNameReqd(), value));
+        }
+
+        if (value.matches("^[\\s/:].*|.*[\\s/:]$")) {
```
ash3rz: I believe in ToolAdminDetailsDialog we're adding `docker.cyverse.org/` to the tool image value.  Is that because we only ever want the images to start with `docker.cyverse.org`? If so, do we need some sort of validation to make sure they didn't remove `docker.cyverse.org` and put in a different URL?

psarando: We just want to put the default to assist the admins and remind them that they should be adding tools that use the `docker.cyverse.org` Docker registry proxy (and not our internal Docker registry name).

Although they should only be adding tools from our `docker.cyverse.org` registry, we don't want to strictly enforce that and lock the admins into that for now.

psarando: For example, if the admins really require it, they may want to add a tool that uses a BioConda Docker image, but at least this default will encourage them to copy it to our registry and use that instead ;)

ash3rz: Cool, makes sense 👍

psarando: Could this regex be shortened to `^[\\s/:].*[\\s/:]$`, without the `|.*` in the middle?

sri2k1us: May be. Let me give that a shot!

sri2k1us: Hmm that didn't work. I think we need the `|` for `OR` operation ? Test if the string begins `OR` ends with`space /  :`

psarando: Ah, I see now. I was misreading this as "does `not` begin and end with `space / :`".
So you're right, we need the `OR` for the opposite (test if it does begin or end with...).

PR #152, opened 2017-06-13
CORE-8888 Set maxInactiveInterval on the user's session to configure …

ash3rz:
…timeout length

NOTE: Requires ansible changes before merging.


ash3rz:
😄  Thanks for the review!  I'll merge in the ansible/consul changes and then merge this in.

ash3rz:
Oh, before I do that I was assuming 60 minutes for the timeout was appropriate - I thought there was some other auth restriction set to 60 minutes (could be remembering wrong).  Does that make sense or should we use a different time?

psarando:
If our CAS timeout is 8 hours, maybe this Spring session timeout should match?

ash3rz:
Sounds good to me!  I'll make sure that's reflected in the ansible change then.

PR #154, opened 2017-06-15
Core 8416 ref

sri2k1us:
- Contains a bug fix. I was using a non-admin end-point in belphegor. 
- Java 8 refactors.
- Introduced appearances in few places i missed.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolRequest/presenter/ToolRequestPresenterImpl.java
+++ b/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolRequest/presenter/ToolRequestPresenterImpl.java
@@ -23,7 +23,6 @@ import com.google.gwt.user.client.ui.HasOneWidget;
 import com.google.inject.Inject;
 
 import com.sencha.gxt.data.shared.ListStore;
-import com.sencha.gxt.data.shared.ModelKeyProvider;
 
 import java.util.List;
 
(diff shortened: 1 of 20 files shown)


de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolRequest/presenter/ToolRequestPresenterImpl.java
```diff
@@ -122,12 +121,7 @@ public void setViewDebugId(String baseId) {
 
     @Override
     public void onAdminMakeToolPublicSelected(AdminMakeToolPublicSelectedEvent event) {
-        adminView = adminFactory.create(new ListStore<>(new ModelKeyProvider<Tool>() {
-            @Override
-            public String getKey(Tool item) {
-                return item.getId();
-            }
-        }));
+        adminView = adminFactory.create(new ListStore<>(item -> item.getId()));
```
ianmcorvidae: This is really clean!

Mostly because I don't know how Java works (lol): how does this syntax know to override the "getKey" method specifically, or does it just do some sort of different handling when it's passed a function rather than a class, or what?

sri2k1us: This syntax (works only for interfaces with one method only. So it knows which method to override. See https://docs.oracle.com/javase/tutorial/java/javaOO/lambdaexpressions.html#approach6
also See http://docs.sencha.com/gxt/4.x/javadoc/gxt-4.0.2/index.html

ianmcorvidae: Ah, that makes sense! Thanks.

sri2k1us:
Thanks for reviews!

PR #155, opened 2017-06-15
More collaborator list fixes

ash3rz:
* Validation checks on group names
* Removed old code relating to the old Collaborators model
* Some other bug fixes
* Disabling collaborator lists in the UI until the view gets updated per the design discussion


ash3rz:
Thanks for the review!

PR #156, opened 2017-06-16
Patch up NPEs and fix dummy subject creation for anonymous user

ash3rz:
When the UI looks up sharing info, it first gets a list of usernames that have permissions on the resource.  Once that's returned, then a secondary call is made to look up the user information to be able to display the user's name.  In the case of the anonymous user, the secondary call returns no information because that user doesn't exist in grouper, so the UI creates a "dummy" collaborator to display.  The UI has been updated to create a dummy collaborator where the username is both the id and the name.  The username is then what is displayed in the UI for that particular case.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/client/models/collaborators/Subject.java
+++ b/de-lib/src/main/java/org/iplantc/de/client/models/collaborators/Subject.java
@@ -58,7 +58,7 @@ public interface Subject extends HasSettableId, HasName {
     void setSourceId(String sourceId);
 
     default boolean isCollaboratorList() {
-        return getSourceId().equals(GROUP_IDENTIFIER);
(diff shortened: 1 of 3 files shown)


ash3rz:
Thanks for the review! 😄

de-lib/src/main/java/org/iplantc/de/client/models/collaborators/Subject.java
```diff
@@ -58,7 +58,7 @@
     void setSourceId(String sourceId);
 
     default boolean isCollaboratorList() {
-        return getSourceId().equals(GROUP_IDENTIFIER);
+        return GROUP_IDENTIFIER.equals(getSourceId());
```
psarando: 👍

PR #157, opened 2017-06-16
CORE-8893 Enable "View in Genome Browser" for multiple genome files.

psarando:
The "View in Genome Browser" menu item is now enabled when multiple genome files or their index files are selected. At least 1 genome file must be selected, but if an index file isn't selected, its existence is still checked and it's still automatically shared with the anon user. The file viewer will only open if only 1 genome file is selected.

The "View in Genome Browser" dialog will now display the URLs for all selected genome files (but not their index files), and its help text and labels have been updated.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/client/services/impl/DiskResourceServiceFacadeImpl.java
+++ b/de-lib/src/main/java/org/iplantc/de/client/services/impl/DiskResourceServiceFacadeImpl.java
@@ -969,11 +969,31 @@ public class DiskResourceServiceFacadeImpl extends TreeStore<Folder> implements
     }
 
     @Override
-    public void shareWithAnonymous(HasPaths diskResourcePaths, DECallback<String> callback) {
+    public void shareWithAnonymous(final HasPaths diskResourcePaths, DECallback<List<String>> callback) {
         String address = deProperties.getDataMgmtBaseUrl() + "anon-files"; //$NON-NLS-1$
         final String body = encode(diskResourcePaths);
         ServiceCallWrapper wrapper = new ServiceCallWrapper(POST, address, body);
-        callService(wrapper, callback);
+
+        callService(wrapper, new DECallbackConverter<String, List<String>>(callback) {
+            @Override
+            protected List<String> convertFrom(final String json) {
+                Splittable obj = StringQuoter.split(json);
+                Splittable pathSplittable = obj.get("paths");
+
+                List<String> urlList = Lists.newArrayList();
+                for (String path : diskResourcePaths.getPaths()) {
+                    Splittable url = pathSplittable.get(path);
+                    if (url != null) {
+                        urlList.add(url.asString());
+                    }
+                }
+
+                urlList.sort(String::compareToIgnoreCase);
+
+                return urlList;
+            }
+
+        });
     }
 
     @Override
--- a/de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/DiskResourcePresenterImpl.java
+++ b/de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/DiskResourcePresenterImpl.java
@@ -508,15 +508,19 @@ public class DiskResourcePresenterImpl implements
 
     @Override
     public void onSendToEnsemblSelected(SendToEnsemblSelected event) {
-        // There is typically only one resource used.
-        for (DiskResource resource : event.getResourcesToSend()) {
-            InfoType infoType = InfoType.fromTypeString(resource.getInfoType());
-            if (infoType == null || !diskResourceUtil.isEnsemblInfoType(infoType)) {
+        final List<DiskResource> resourcesToSend = event.getResourcesToSend();
 
-                announcer.schedule(new ErrorAnnouncementConfig(appearance.unsupportedEnsemblInfoType()));
-                return;
-            }
-            eventBus.fireEvent(new RequestSendToEnsemblEvent((File)resource, infoType));
+        boolean allGenomeAndIndexTypes = resourcesToSend.stream().allMatch(resource -> {
+            final String path = resource.getPath();
+            InfoType infoType = InfoType.fromTypeString(resource.getInfoType());
+
+            return diskResourceUtil.isGenomeIndexFile(path) || diskResourceUtil.isEnsemblInfoType(infoType);
+        });
+
+        if (allGenomeAndIndexTypes) {
+            eventBus.fireEvent(new RequestSendToEnsemblEvent(resourcesToSend));
+        } else {
+            announcer.schedule(new ErrorAnnouncementConfig(appearance.unsupportedEnsemblInfoType()));
         }
     }
 
--- a/de-lib/src/main/java/org/iplantc/de/diskResource/client/views/toolbar/DiskResourceViewToolbarImpl.java
+++ b/de-lib/src/main/java/org/iplantc/de/diskResource/client/views/toolbar/DiskResourceViewToolbarImpl.java
@@ -3,7 +3,6 @@ package org.iplantc.de.diskResource.client.views.toolbar;
 import org.iplantc.de.client.events.EventBus;
 import org.iplantc.de.client.models.UserInfo;
 import org.iplantc.de.client.models.diskResources.DiskResource;
-import org.iplantc.de.client.models.diskResources.File;
 import org.iplantc.de.client.models.diskResources.Folder;
 import org.iplantc.de.client.models.viewer.InfoType;
 import org.iplantc.de.client.models.viewer.MimeType;
@@ -283,6 +282,8 @@ public class DiskResourceViewToolbarImpl extends Composite implements ToolbarVie
         final boolean isSelectionInTrash = isSelectionInTrash(selectedDiskResources);
         final boolean isFolderSelect = !isSelectionEmpty
                                        && firstItem instanceof Folder;
+        final boolean containsFile = diskResourceUtil.containsFile(selectedDiskResources);
+        final boolean containsFolder = diskResourceUtil.containsFolder(selectedDiskResources);
 
         moveToTrashMiEnabled = !isSelectionEmpty && isOwner && !isSelectionInTrash && !containsFilteredItems;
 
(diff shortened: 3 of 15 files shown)


de-lib/src/main/java/org/iplantc/de/diskResource/client/views/toolbar/DiskResourceViewToolbarImpl.java
```diff
@@ -283,46 +282,47 @@ public void onDiskResourceSelectionChanged(DiskResourceSelectionChangedEvent eve
...
                 && !containsFilteredItems;
         sendToEnsemblMiEnabled = !isSelectionEmpty
-                && isSingleSelection
-                && containsFile(selectedDiskResources)
+                && !containsFolder
                 && !isSelectionInTrash
-                && diskResourceUtil.isEnsemblInfoType(getInfoTypeFromSingletonCollection(selectedDiskResources))
-                && !containsFilteredItems;
+                && !containsFilteredItems
+                && containsOnlyEnsemblFiles(selectedDiskResources);
```
psarando: Here is the logic that enables the "View in Genome Browser" menu item when multiple genome files (or their index files) are selected, with the help of a new `containsOnlyEnsemblFiles` method.

Most of the other changes in this file were replacing the custom `containsFile` and `containsOnlyFolders` methods with booleans, so now we only have to call `diskResourceUtil.containsFile` and `diskResourceUtil.containsFolder` once, and use those results everywhere else in this method.

de-lib/src/main/java/org/iplantc/de/fileViewers/client/callbacks/EnsemblUtil.java
```diff
@@ -33,60 +29,49 @@
...
-            diskResourcePaths.setPaths(Arrays.asList(path, indexFilePath));
-        } else {
-            diskResourcePaths.setPaths(Arrays.asList(path));
+            if (!Strings.isNullOrEmpty(indexFilePath)) {
+                pathMap.put(indexFilePath, TYPE.FILE);
+            }
         }
 
-        diskResourceServiceFacade.getStat(diskResourceUtil.asStringPathTypeMap(list, TYPE.FILE),
+        diskResourceServiceFacade.getStat(pathMap,
```
psarando: Rather than iterating over the resources to send and building a list of `HasPath` items, which are then converted to a "path:type" map by `diskResourceUtil.asStringPathTypeMap`, I refactored this loop to build the `pathMap` directly. Then the keys (paths) of the `pathMap` are used in the `shareWithAnonymous` call below.

This also automatically de-duplicates paths of genome file index files (*.bai or *.tbi) that need to be shared with anon if the user already selected those files from the Data window (see the `indexFilePath` above).

de-lib/src/main/java/org/iplantc/de/fileViewers/client/callbacks/ShareAnonymousCallback.java
```diff
@@ -60,32 +60,33 @@ public void onFailure(Integer statusCode, Throwable caught) {
...
         if (container != null) {
             container.unmask();
         }
-        JSONObject obj = jsonUtil.getObject(result);
-        JSONObject paths = jsonUtil.getObject(obj, "paths");
-        showShareLink(jsonUtil.getString(paths, file.getPath()));
+
+        showShareLink(urls.stream()
+                          .filter(path -> !diskResourceUtil.isGenomeIndexFile(path))
+                          .collect(Collectors.toList()));
```
psarando: The `shareWithAnonymous` call now uses a callback converter to pass this callback a list of sorted URL strings for all the paths shared with the anon user.

So now this callback needs to filter out the genome index files from the dialog displayed on success, since the user can't use the URLs to those index files in the genome browser sites (but the index files are still required to be readable by those sites).

de-lib/src/main/java/org/iplantc/de/theme/base/client/fileViewers/FileViewerContextualHelpStrings.properties
```diff
@@ -1 +1,8 @@
-sendToEnsemblUrlHelp = To access this file in genome browsers, this file needs to be public. If you do not want this file to be public, you can change its permissions as follows: <b>1.</b> Navigate to the file in the data window <b>2.</b> Select the file <b>3.</b> Select <i>Share</i>-><i>Share with collaborators</i> from the menu. <b>4.</b> In the <i>Manage Sharing</i> window in the <i>Who has access</i> section and remove permissions to the <i>anonymous</i> user.
+sendToEnsemblUrlHelp = To access a file in genome browsers, the file needs to be public. \
+  If you do not want the file to be public, you can change its permissions as follows:\
+  <ol>\
+  <li><b>1.</b> Navigate to the file in the data window.</li>\
+  <li><b>2.</b> Select the file.</li>\
+  <li><b>3.</b> Select <i>Share</i>-><i>Share with collaborators</i> from the menu.</li>\
+  <li><b>4.</b> In the <i>Manage Sharing</i> window in the <i>Who has access</i> section and remove permissions to the <i>anonymous</i> user.</li>\
```
psarando: The help text won't display the HTML for these `li` items, unfortunately, so the numbers are still required as part of the text; but at least it will display each item with line breaks in between them.

sri2k1us: 😖  I had the same problem when trying to format introduction texts for the DE.

psarando: 😩

sri2k1us:
LGTM :shipit:

de-lib/src/main/java/org/iplantc/de/client/services/impl/DiskResourceServiceFacadeImpl.java
```diff
@@ -969,11 +969,31 @@ protected MetadataTemplate convertFrom(String object) {
...
+    public void shareWithAnonymous(final HasPaths diskResourcePaths, DECallback<List<String>> callback) {
         String address = deProperties.getDataMgmtBaseUrl() + "anon-files"; //$NON-NLS-1$
         final String body = encode(diskResourcePaths);
         ServiceCallWrapper wrapper = new ServiceCallWrapper(POST, address, body);
-        callService(wrapper, callback);
+
+        callService(wrapper, new DECallbackConverter<String, List<String>>(callback) {
+            @Override
+            protected List<String> convertFrom(final String json) {
+                Splittable obj = StringQuoter.split(json);
```
ash3rz: Could we move away from using Splittable here or is this a special case?

de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/DiskResourcePresenterImpl.java
```diff
@@ -508,15 +508,19 @@ public void onSendToCogeSelected(SendToCogeSelected event) {
...
+        boolean allGenomeAndIndexTypes = resourcesToSend.stream().allMatch(resource -> {
+            final String path = resource.getPath();
             InfoType infoType = InfoType.fromTypeString(resource.getInfoType());
-            if (infoType == null || !diskResourceUtil.isEnsemblInfoType(infoType)) {
 
-                announcer.schedule(new ErrorAnnouncementConfig(appearance.unsupportedEnsemblInfoType()));
-                return;
-            }
-            eventBus.fireEvent(new RequestSendToEnsemblEvent((File)resource, infoType));
+            return diskResourceUtil.isGenomeIndexFile(path) || diskResourceUtil.isEnsemblInfoType(infoType);
```
ash3rz: Yay Java 8!

de-lib/src/main/java/org/iplantc/de/fileViewers/client/callbacks/ShareAnonymousCallback.java
```diff
@@ -60,32 +60,33 @@ public void onFailure(Integer statusCode, Throwable caught) {
...
+    private void showShareLink(List<String> linkIds) {
         // Open dialog window with text selected.
         IPlantDialog dlg = new IPlantDialog(true);
         dlg.setHeading(appearance.sendToEnsemblMenuItem());
         attachHelp(dlg);
 
         dlg.setHideOnButtonClick(true);
         dlg.setResizable(false);
-        dlg.setSize("535", "175");
+        dlg.setSize("640", "320");
```
ash3rz: We should probably use an appearance class for these size values.  But if you're trying to get this in for Monday's drop, it's not a huge deal.

de-lib/src/main/java/org/iplantc/de/theme/base/client/fileViewers/FileViewerStrings.properties
```diff
@@ -35,6 +35,11 @@ label = Label
...
-
+sendToEnsemblePopupNote = Note: Anyone with access to a file''s URL can access that file.
+ensemblUrl = Please visit <a href='http://ensemblgenomes.org/info/genomes' target="_blank">Ensembl</a> \
+  / <a href='https://genome.ucsc.edu/goldenPath/customTracks/custTracks.html' target="_blank"> UCSC </a> \
+  / <a href='http://www.broadinstitute.org/igv/' target="_blank"> IGV </a> \
+  / <a href='http://gmod.org/wiki/GBrowse' target="_blank">GBrowse</a> \
+  / <a href='http://jbrowse.org/' target="_blank">jbrowse</a> \
+  / <a href='http://epigenomegateway.wustl.edu/browser/' target='_blank'> WashU EPIGenome Browser</a> \
+  and use one of the following URLs to import your bam / vcf / gff / bed / bigBed / bigWig file
```
ash3rz: Much more readable 👍

de-lib/src/main/java/org/iplantc/de/client/services/impl/DiskResourceServiceFacadeImpl.java
psarando: I'm pretty sure this is a special case, since the endpoint returns a response like the following:
```json
{
  "/path/1": "http://url/1",
  "/path/2": "http://url/2"
}
```
I'm not sure if AutoBeans can handle variable keys like this, and since we only need the URL values, it seemed easier to parse them into a list, while also keeping this parsing logic isolated in the service facade.

de-lib/src/main/java/org/iplantc/de/fileViewers/client/callbacks/ShareAnonymousCallback.java
psarando: Sure, that's easy enough since this class is already using an appearance 👍

psarando:
Thanks for the reviews! 👍

PR #158, opened 2017-06-19
Patch up NPE and index out of bounds error during DND

ash3rz:
I couldn't find what specific action caused the index out of bounds error, but I did find some NPEs and figured we could patch up the index error.  Let me know if we'd feel more comfortable trying to pin down what causes the error instead.


ash3rz:
Thanks for the review!

PR #160, opened 2017-06-20
Core 8914 Allow users to set entrypoint for their private tools

sri2k1us:
- Also refactor code to use Editor framework.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/dialogs/ToolAdminDetailsDialog.java
+++ b/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/dialogs/ToolAdminDetailsDialog.java
@@ -8,16 +8,8 @@ import org.iplantc.de.admin.desktop.shared.Belphegor;
 import org.iplantc.de.client.models.IsHideable;
 import org.iplantc.de.client.models.tool.Tool;
 import org.iplantc.de.client.models.tool.ToolAutoBeanFactory;
-import org.iplantc.de.client.models.tool.ToolContainer;
-import org.iplantc.de.client.models.tool.ToolDevice;
-import org.iplantc.de.client.models.tool.ToolImage;
-import org.iplantc.de.client.models.tool.ToolImplementation;
-import org.iplantc.de.client.models.tool.ToolTestData;
-import org.iplantc.de.client.models.tool.ToolVolume;
-import org.iplantc.de.client.models.tool.ToolVolumesFrom;
 import org.iplantc.de.commons.client.views.dialogs.IPlantDialog;
 
-import com.google.common.collect.Lists;
 import com.google.gwt.event.shared.HandlerRegistration;
 import com.google.gwt.event.shared.HasHandlers;
 import com.google.gwt.user.client.ui.HTML;
@@ -94,24 +86,9 @@ public class ToolAdminDetailsDialog extends IPlantDialog implements IsHideable,
 
     @Override
     public void show() {
-        final ToolContainer toolContainer = factory.getContainer().as();
-        toolContainer.setDeviceList(Lists.<ToolDevice>newArrayList());
-        toolContainer.setContainerVolumes(Lists.<ToolVolume>newArrayList());
-        toolContainer.setContainerVolumesFrom(Lists.<ToolVolumesFrom>newArrayList());
-        ToolImage image = factory.getImage().as();
-        image.setName(appearance.defaultImgPrefix());
-        toolContainer.setImage(image);
-
-        final ToolImplementation toolImplementation = factory.getImplementation().as();
-        final ToolTestData toolTestData = factory.getTest().as();
-        toolTestData.setInputFiles(Lists.<String>newArrayList());
-        toolTestData.setOutputFiles(Lists.<String>newArrayList());
-        toolImplementation.setTest(toolTestData);
-
-        final Tool tool = factory.getTool().as();
+        final Tool tool = factory.getDefaultTool().as();
         tool.setType(appearance.toolImportTypeDefaultValue());
-        tool.setContainer(toolContainer);
-        tool.setImplementation(toolImplementation);
+        tool.getContainer().getImage().setName(appearance.defaultImgPrefix());
 
         show(tool, Mode.EDIT);
     }
--- a/de-lib/src/main/java/org/iplantc/de/client/models/tool/ToolAutoBeanFactory.java
+++ b/de-lib/src/main/java/org/iplantc/de/client/models/tool/ToolAutoBeanFactory.java
@@ -39,7 +39,7 @@ public interface ToolAutoBeanFactory extends AutoBeanFactory {
      *
      * @return  a tool
      */
-    default AutoBean<Tool> geTool() {
+    default AutoBean<Tool> getDefaultTool() {
         Tool tool = getTool().as();
         ToolImage image = getImage().as();
         ToolContainer container = getContainer().as();
--- a/de-lib/src/main/java/org/iplantc/de/client/models/tool/ToolAutoBeanFactory.java
+++ b/de-lib/src/main/java/org/iplantc/de/client/models/tool/ToolAutoBeanFactory.java
@@ -7,6 +7,7 @@ import org.iplantc.de.client.models.errorHandling.SimpleServiceError;
 
 import com.google.web.bindery.autobean.shared.AutoBean;
 import com.google.web.bindery.autobean.shared.AutoBeanFactory;
+import com.google.web.bindery.autobean.shared.AutoBeanUtils;
 
 /**
  * @author sriram
@@ -32,5 +33,21 @@ public interface ToolAutoBeanFactory extends AutoBeanFactory {
     AutoBean<ToolTestData> getTest();
 
     AutoBean<SimpleServiceError> simpleServiceError();
+
+    /**
+     * Build a tool with default values
+     *
+     * @return  a tool
+     */
+    default AutoBean<Tool> geTool() {
+        Tool tool = getTool().as();
+        ToolImage image = getImage().as();
+        ToolContainer container = getContainer().as();
+
+        container.setImage(image);
+        tool.setContainer(container);
+
+        return AutoBeanUtils.getAutoBean(tool);
+    }
 
 }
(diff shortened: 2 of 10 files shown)


tedgin:
Please add javadoc comments describing the purpose of the new public functions and interfaces.

de-lib/src/main/java/org/iplantc/de/tools/client/presenter/ManageToolsViewPresenter.java
```diff
@@ -183,6 +189,12 @@ public void onFailure(Throwable throwable) {
             @Override
             public void onSuccess(EditToolDialog etd) {
                 etd.setSize(appearance.editDialogWidth(), appearance.editDialogHeight());
+                Tool tool = factory.getTool().as();
+                ToolImage image = factory.getImage().as();
+                ToolContainer container = factory.getContainer().as();
+                container.setImage(image);
+                tool.setContainer(container);
+                etd.editTool(tool);
```
psarando: Could this be refactored into a default factory method?

de-lib/src/main/java/org/iplantc/de/tools/client/views/manage/EditToolViewImpl.java
```diff
@@ -112,6 +129,16 @@ private SafeHtml buildRequiredFieldLabel(final String label) {
     }
 
     @Override
+    public Tool getTool() {
+        return editorDriver.flush();
+    }
+
+    @Override
+    public void editTool(Tool t) {
+        editorDriver.edit(t);
```
psarando: Nice refactor using the `EditorDriver` 👍

de-lib/src/main/java/org/iplantc/de/theme/base/client/tools/ToolMessages.java
```diff
@@ -86,6 +86,8 @@
     String manageSharing();
 
     String done();
+
+    String entryPoint();
```
psarando: I agree with @tedgin that we should add java docs to the interface methods (at least just for the methods added in this PR). It could also just be a copy of the [endpoint docs](https://github.com/cyverse-de/apps/blob/2.12.0/src/apps/routes/schemas/containers.clj#L56).

de-lib/src/main/java/org/iplantc/de/tools/client/presenter/ManageToolsViewPresenter.java
sri2k1us: 👍

ash3rz:
LGTM 👍

de-lib/src/main/java/org/iplantc/de/client/models/tool/ToolAutoBeanFactory.java
```diff
@@ -33,4 +34,20 @@
 
     AutoBean<SimpleServiceError> simpleServiceError();
 
+    /**
+     * Build a tool with default values
+     *
+     * @return  a tool
+     */
+    default AutoBean<Tool> geTool() {
+        Tool tool = getTool().as();
```
psarando: Will these `getTool` method names conflict?
It seems like this `getTool` method will be the only one accessible from now on?

Which may not be a bad thing, but if we keep this as the only `getTool` method, then we should refactor other places the old method was called from. For example, the call in `ToolAdminDetailsDialog` no longer needs to create and set a default container.

psarando: After checking, it looks like `ToolAdminDetailsDialog#show` is the only other place this factory's `getTool` method it called from.

ash3rz: Good catch.  Maybe the method should be `getDefaultTool` for clarity?

sri2k1us: Actually the method name is `geTool()`, note the missing `t` and that why java didn't complain. I went ahead and renamed the method name.

sri2k1us:
Can I merge this ? Still waiting on approval

de-lib/src/main/java/org/iplantc/de/tools/client/views/manage/EditToolViewImpl.java
```diff
@@ -45,46 +41,66 @@
...
+    @Ignore
     @UiField
     TextField time;
 
-    Hidden toolId;
+    @Path("container.entryPoint")
+    @UiField
+    TextField entryPoint;
+
+    Hidden id;
```
ash3rz: I didn't notice this last time I reviewed.  At a glance I didn't see it used anywhere, is it needed?

de-lib/src/main/java/org/iplantc/de/client/models/tool/ToolAutoBeanFactory.java
```diff
@@ -33,4 +34,20 @@
 
     AutoBean<SimpleServiceError> simpleServiceError();
 
+    /**
+     * Build a tool with default values
+     *
+     * @return  a tool
+     */
+    default AutoBean<Tool> getDefaultTool() {
```
psarando: I still think it might be a good idea for the `ToolAdminDetailsDialog` to take advantage of this new default method.

de-lib/src/main/java/org/iplantc/de/tools/client/views/manage/EditToolViewImpl.java
sri2k1us: This is the field that collects `entrypoint` for the image.

de-lib/src/main/java/org/iplantc/de/client/models/tool/ToolAutoBeanFactory.java
sri2k1us: 👍

de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/dialogs/ToolAdminDetailsDialog.java
```diff
@@ -94,24 +86,9 @@ public void show(final Tool tool, Mode mode) {
...
-        toolContainer.setContainerVolumesFrom(Lists.<ToolVolumesFrom>newArrayList());
-        ToolImage image = factory.getImage().as();
-        image.setName(appearance.defaultImgPrefix());
-        toolContainer.setImage(image);
-
-        final ToolImplementation toolImplementation = factory.getImplementation().as();
-        final ToolTestData toolTestData = factory.getTest().as();
-        toolTestData.setInputFiles(Lists.<String>newArrayList());
-        toolTestData.setOutputFiles(Lists.<String>newArrayList());
-        toolImplementation.setTest(toolTestData);
```
psarando: I think we still need to set these default `ToolImage` name, `ToolContainer` settings, and `ToolImplementation` settings here (`getDefaultTool` only sets an empty `ToolImage` and `ToolContainer`).

de-lib/src/main/java/org/iplantc/de/tools/client/views/manage/EditToolViewImpl.java
ash3rz: The `Hidden id` does?  That's what I was asking about.

de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/view/dialogs/ToolAdminDetailsDialog.java
sri2k1us: Hmm. Those are just empty lists / data. UI seems to work without them.

psarando: Oh OK, good to know. I thought they might be required for the editor driver 👍

ash3rz: Oh I didn't realize these were removed.  You might wanna test.  I vaguely remember I had to add all those empty lists to make sure the JSON structure was in tact.  It was something weird like if you didn't create the nested structure, then when you went to save something within ToolContainer or ToolTestData for example, those lists wouldn't actually save because it needed a list to already be there?  Something weird like that, I honestly don't remember the details after all this time though.  It could be fixed after the updates/upgrades we've done though.

sri2k1us: Yes. Doing more  test now.

sri2k1us: Yup saving fails which is annoying because we have to set all the empty as defaults. In that case do you mind If we the keep the original method instead using `default` method? I see no major improvement with using the `default` method.

sri2k1us: Or Change the default method to include setting these empty lists

psarando: I wouldn't add all those defaults to `getDefaultTool`, since private tools can't set volumes or devices.

psarando: Reverting this change is fine by me.

ash3rz: Yeah, sounds ok to me.  I think this is the only spot where we set up the whole JSON structure so... makes sense to leave it there.

de-lib/src/main/java/org/iplantc/de/tools/client/views/manage/EditToolViewImpl.java
sri2k1us: Ohh. OOPS! That can be removed since refactoring of `EditorDriver` took care the need for tracking the `id`

PR #161, opened 2017-06-20
core-8885 add PIDS limit belphegor tools admin UI.

sri2k1us:


de-lib/src/main/java/org/iplantc/de/theme/base/client/admin/toolAdmin/ToolAdminDisplayStrings.properties
```diff
@@ -16,6 +16,7 @@ dialogMakePublicText = Make public
 timeLimit = Time limit (seconds)
 restricted = Restricted
 defaultImgPrefix = docker.cyverse.org/
+pidsLimit = PIDS Limit
```
psarando: The `s` should be lower-case: `PIDs Limit`

ash3rz:
Agreed with Paul's comment, otherwise LGTM 🚢

PR #162, opened 2017-06-21
core-8919 set missing debug ids.

sri2k1us:


de-lib/src/main/java/org/iplantc/de/tools/shared/ToolsModule.java
```diff
@@ -25,6 +25,7 @@
 
         String SHARING_VIEW = ".view";
         String SHARING_PERMS = ".permPanel";
+        String Tool_SEARCH =".search";
```
ash3rz: Looks like a typo here, should be all caps for the variable name

sri2k1us: 👍

PR #163, opened 2017-06-26
CORE-8923 Implement session management for UI

sri2k1us:
- Configure Spring to use Redis for session management
- This code requires Ansible changes

### DIFF
--- a/de-webapp/src/main/java/org/iplantc/de/Application.java
+++ b/de-webapp/src/main/java/org/iplantc/de/Application.java
@@ -93,7 +93,7 @@ public class Application extends AbstractAnnotationConfigDispatcherServletInitia
         dispatcher.setLoadOnStartup(0);
         dispatcher.addMapping("/de/websocket");
         dispatcher.setInitParameter("org.atmosphere.cpr.packages", "org.iplantc.de.server");
-        LOG.error("************** Started AtmosphereServlet ***********");
+        LOG.info("************** Started AtmosphereServlet ***********");
     }
 
 
--- a/de-webapp/src/main/java/org/iplantc/de/server/HttpSessionConfig.java
+++ b/de-webapp/src/main/java/org/iplantc/de/server/HttpSessionConfig.java
@@ -1,0 +1,40 @@
+package org.iplantc.de.server;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.PropertySource;
+import org.springframework.context.support.PropertySourcesPlaceholderConfigurer;
+import org.springframework.session.data.redis.config.annotation.web.http.EnableRedisHttpSession;
+import org.springframework.data.redis.connection.jedis.JedisConnectionFactory;
+/**
+ * Created by sriram on 6/16/17.
+ */
+
+@EnableRedisHttpSession
+@Configuration
+@PropertySource("file:/etc/iplant/de/de.properties")
+public class HttpSessionConfig {
+
+
+    @Value("${redis.host}")
+    private String redisHostName;
+
+    @Value("${redis.port}")
+    private int redisPort;
+
+    @Bean
+    public static PropertySourcesPlaceholderConfigurer propertySourcesPlaceholderConfigurer() {
+        return new PropertySourcesPlaceholderConfigurer();
+    }
+
+    @Bean
+    JedisConnectionFactory jedisConnectionFactory() {
+        JedisConnectionFactory factory = new JedisConnectionFactory();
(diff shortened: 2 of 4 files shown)


de-webapp/src/main/java/org/iplantc/de/Application.java
```diff
@@ -93,7 +93,7 @@ public void onStartup(ServletContext container) {
         dispatcher.setLoadOnStartup(0);
         dispatcher.addMapping("/de/websocket");
         dispatcher.setInitParameter("org.atmosphere.cpr.packages", "org.iplantc.de.server");
-        LOG.error("************** Started AtmosphereServlet ***********");
+        LOG.info("************** Started AtmosphereServlet ***********");
```
sri2k1us: Not necessary change for this issue but it kept bothering me so changed the log level.

de-webapp/src/main/java/org/iplantc/de/server/HttpSessionConfig.java
```diff
@@ -0,0 +1,61 @@
...
+        factory.setHostName(getHostName());
+        factory.setPort(getPort());
+        factory.setDatabase(getDbNumber());
+        factory.setUsePool(true);
+        factory.setPassword(getPassword());
+        return factory;
+    }
+
+    private String getHostName() {
+        return environment.getProperty("org.iplantc.discoveryenvironment.redis.host");
```
ianmcorvidae: This part's almost certainly my own unfamiliarity with the UI code: how do these values get filled in? In particular, if these need to go into the config file, then they probably need to be added to the `de.properties.tmpl` file (and the ansible changes should include populating consul with the relevant values etc).

de-webapp/src/main/java/org/iplantc/de/server/HttpSessionConfig.java
```diff
@@ -0,0 +1,61 @@
...
+    public static PropertySourcesPlaceholderConfigurer propertySourcesPlaceholderConfigurer() {
+        return new PropertySourcesPlaceholderConfigurer();
+    }
+
+    @Bean
+    JedisConnectionFactory jedisConnectionFactory() {
+        JedisConnectionFactory factory = new JedisConnectionFactory();
+        factory.setHostName(getHostName());
+        factory.setPort(getPort());
+        factory.setDatabase(getDbNumber());
```
sri2k1us: 👍

de-webapp/src/main/java/org/iplantc/de/server/HttpSessionConfig.java
```diff
@@ -0,0 +1,61 @@
...
+        factory.setUsePool(true);
```
sri2k1us: Ahh good point. Although am not sure what will be a good pool size to configure. I will try to find out.

de-webapp/src/main/java/org/iplantc/de/server/HttpSessionConfig.java
sri2k1us: Spring gives us the ability read from de.properties. I have submitted a PR in `ansible` and `de-ansible`

ianmcorvidae: Right, that's what I was figuring. For consul configs, you'll need to update `de.properties.tmpl` in this repo -- that's the file which configures where values from consul are put into the de.properties used for consul configs. The `j2` file in the ansible repo is what makes it work correctly when generated by ansible, e.g. for config images, but the one in this repo is what controls it for consul configs.

Someday we'll figure out how to not need two files that do the same thing :(

ianmcorvidae:
The template file looks correct to me. All the property names are the same, the keys in consul match the ansible PR, and the various declarations are syntactically correct unless I've missed something.

sri2k1us:
Thanks for the reviews!

de-webapp/src/main/java/org/iplantc/de/server/HttpSessionConfig.java
ash3rz: Spring is still magic to me, but instead of doing these methods, will the `@Value("${org.iplantc.discoveryenvironment.redis.host}") private String redisHost` syntax not work here?  Similar to `DefaultJwtBuilder.java` for example.

sri2k1us: I think it should work. Let me give a shot! We have used both the approaches in our project. Found from some googling, It looks like `@Value` approach is best way to read property files.

ash3rz: Cool, I hadn't seen the other approach before so was mostly just curious. But if the `@Value` works and seems better then that's great  👍  Changes look good to me

PR #166, opened 2017-06-29
CORE-8910 Merge the Collaborators and Collaborator Lists into one view

ash3rz:
Majority of the changes are 
* Moving items from GroupView to ManageCollaboratorsView
* Deleting the old GroupView code
* Merging the GroupView appearance into the ManageCollaboratorsAppearance
* Setting up a parent callback so that when a user hits the Delete button, it will do all the logic to delete any users selected (one call no matter how many users selected) and delete any groups selected (the number of calls is equivalent to the number of selected groups).


de-lib/src/main/java/org/iplantc/de/collaborators/client/presenter/ManageCollaboratorsPresenter.java
```diff
@@ -372,25 +484,29 @@ public void onGroupSaved(GroupSaved event) {
...
             }
         });
     }
 
+    ParentDeleteSubjectsCallback getParentDeleteSubjectsCallback() {
+        return new ParentDeleteSubjectsCallback();
+    }
+
     AddMemberToGroupCallback getAddMemberToGroupCallback() {
         return new AddMemberToGroupCallback();
```
ash3rz: Sure, I can change the method names

de-lib/src/main/java/org/iplantc/de/collaborators/client/presenter/callbacks/ParentDeleteSubjectsAbstractCallback.java
```diff
@@ -0,0 +1,54 @@
...
+    int doneCount;
+    List<UpdateMemberResult> successUsers = Lists.newArrayList();
+    List<Group> successGroups = Lists.newArrayList();
+    List<Throwable> failures = Lists.newArrayList();
+
+    public void setCallbackCounter(int doneCount) {
+        this.doneCount = doneCount;
+    }
+
+    public synchronized void updateCounter() {
```
ash3rz: Oh whoops, guess that's not necessary then 😄

de-lib/src/main/java/org/iplantc/de/commons/client/ErrorHandler.java
```diff
@@ -107,22 +120,23 @@ public static void post(String errorSummary, Throwable caught) {
...
-        if (errorSummary == null) {
-            errorSummary = SafeHtmlUtils.fromString(appearance.error());
-        }
+        if (caughtList != null && !caughtList.isEmpty()) {
+            caughtList.forEach(caught -> {
+                GWT.log(errorSummary.asString(), caught);
 
-        if (caught != null) {
-            GWT.log(errorSummary.asString(), caught);
+                errorDetails.append(parseExceptionJson(caught) + NEWLINE);
```
psarando: 👍

ash3rz:
Thanks for the reviews!  I was thinking before I merge this, I might want to change the icon for Collaborator Lists from group to something more "list" related.  That way that group icon (or maybe even something else) could be used for Teams.

PR #167, opened 2017-06-29
CORE-8911 Add DND ability to add collaborators to lists

ash3rz:
Now you can DND collaborators (users and lists) to another list to add them to that list.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/collaborators/client/ManageCollaboratorsView.java
+++ b/de-lib/src/main/java/org/iplantc/de/collaborators/client/ManageCollaboratorsView.java
@@ -8,6 +8,7 @@ import org.iplantc.de.collaborators.client.events.GroupNameSelected;
 import org.iplantc.de.collaborators.client.events.RemoveCollaboratorSelected;
 import org.iplantc.de.collaborators.client.events.UserSearchResultSelected;
 
+import com.google.gwt.dom.client.Element;
 import com.google.gwt.resources.client.ImageResource;
 import com.google.gwt.safehtml.shared.SafeHtml;
 import com.google.gwt.user.client.ui.HasOneWidget;
@@ -107,6 +108,10 @@ public interface ManageCollaboratorsView extends IsWidget,
         String nameHeader();
 
         String institutionOrDescriptionHeader();
+
+        String onlyDNDToListSupported();
+
+        String membersAddedToGroupSuccess(Subject group, List<UpdateMemberResult> userSuccesses);
     }
 
     /**
@@ -227,4 +232,11 @@ public interface ManageCollaboratorsView extends IsWidget,
      * @return
      */
     List<Subject> getCollaborators();
+
+    /**
+     * Return the corresponding Subject in the grid for the given element
+     * @param as
+     * @return
+     */
+    Subject getSubjectFromElement(Element as);
 }
--- a/de-lib/src/main/java/org/iplantc/de/collaborators/client/gin/ManageCollaboratorsViewFactory.java
+++ b/de-lib/src/main/java/org/iplantc/de/collaborators/client/gin/ManageCollaboratorsViewFactory.java
@@ -1,11 +1,12 @@
 package org.iplantc.de.collaborators.client.gin;
 
 import org.iplantc.de.collaborators.client.ManageCollaboratorsView;
+import org.iplantc.de.collaborators.client.views.CollaboratorDNDHandler;
 
 /**
  * @author aramsey
  */
 public interface ManageCollaboratorsViewFactory {
 
-    ManageCollaboratorsView create(ManageCollaboratorsView.MODE mode);
+    ManageCollaboratorsView create(ManageCollaboratorsView.MODE mode, CollaboratorDNDHandler dndHandler);
 }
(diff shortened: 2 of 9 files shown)


de-lib/src/main/java/org/iplantc/de/theme/base/client/collaborators/GroupDisplayStrings.properties
```diff
@@ -1,3 +1,5 @@
+addGroup=Add List
+deleteGroup=Delete List
 deleteGroupConfirm=By deleting the collaborator list {0}, any data that has been shared with this list will immediately be un-shared.  Do you wish to continue deleting the collaborator list?
```
sri2k1us: Oh! I was thinking deleting a list will not impact permissions. Right now, If I remove a person from my collaborators, it will not un-share items with that collaborators. Is there a specific reason we are changing the existing functionality ?

ash3rz: Maybe @dennisroberts71 can better comment, but in my mind deleting a list is similar to deleting a user (actually deleting a user, not just removing them from your Collaborators).  It can't still be shared with them if they don't exist.
But you are correct - if you remove a user from your Collaborators, the sharing persists.  If you remove/delete a list, the sharing does not persist.

ash3rz: I can see how this could be confusing though for the users.  I believe the same thing happens if you remove a user from a list - anything shared with that list will no longer be shared with that user because they're no longer part of the list.  Do we need to rethink this?

sri2k1us: I thought `list` is expanded and shared with individual users. It makes more sense for `Teams` to behave this way (unshare items when the `team` is deleted or unshare with a user when that user is removed from the `team`. We should probably get some input from science team.

dennisroberts71: My initial thinking was that it would be nice for users to be able to choose to share something with a collaborator list directly rather than with all users in the collaborator list individually. Sharing something with individual users in a collaborator list would still be possible, of course, but if a user is removed from a collaborator list then anything shared with the collaborator list directly would no longer be accessible to the user (unless it was also shared with the user directly or with another group that the user is in. I can see where that could be confusing, though.

dennisroberts71: My thinking on the distinction between collaborator lists and teams was merely one of visibility. Users can see when they're in a team and teams can have multiple administrators. Other than that, the functionality was quite similar to that of collaborator lists.

dennisroberts71: I agree; we're going to have to get some input from the science team on this. The distinction between sharing with a group and sharing with an individual user is clear to me, but I really don't know if that distinction will be clear to DE users.

ramonawalls: I think my view of teams and collaborator lists (coll-lists) is the same as Dennis - the difference is a matter of visibility.

The primary use case for having coll-lists is to be able to share things with a group of users at one time, without having to go through that same list of users every time you want to share things with them. Just like teams, I think it will be very useful to have nested collaborator lists up to say three levels deep.

Unsharing gets a bit more complicated, because I can see a few different use cases.

1) I want to stop sharing with one member of a coll-list:
If I remove them from the list, then I should stop sharing everything that has been shared with the list just for that one user. Likewise, if I add a user to the list, everything that is already shared with that list will be shared with the new user.

2) I want to stop sharing with all members of a list and don't plan to use the list again: 
In this case, the easiest solution would be to delete the list, which would automatically remove sharing permissions for anyone on the list, unless something has also been shared with them individually as well, in which case maybe that permission could remain. Maybe that last part is too difficult to manage.

3) I want to stop sharing some or all of the items shared with a coll-list with all members of the list, but I plan to continue using the list:
In this case, I think I would have to remove sharing privileges for "list" from any item I no longer want the members of the list to have access to.  That would be very tedious if I have shared a lot of things with "list", but I don't see another way to do it. I guess another option would be if there was a  dialog for "show me everything shared with this list (or individual user for that matter) and then I could select items one by one or maybe select all.

Hope that is helpful! If not, bug me.

sri2k1us: @ramonawalls  Thanks for the inputs. With respect to `2` and I would like to suggest that we have the same behavior when an individual collaborator is deleted, we un-share everything with that user.

ramonawalls: Sriram, do you mean that if an individual collaborator is removed from a list, then everything that is shared with that list is unshared with that collaborator? If so, then yes, I agree.

dennisroberts71: @sri2k1us I'm not entirely sure I understand the suggestion. When a collaborator is deleted, do you want to unshare everything that the user has ever shared with that collaborator individually? We don't really have a good way to track that at the moment. In other words, we don't have a way to definitively know whether permission for the collaborator to access a resource was granted directly by the current user or not.

We could create a way to track that. The easiest way that I can think of is to create a special group for each collaborator (containing only that collaborator) and have the permissions that would be assigned to that collaborator assigned to that group instead. Then when a user deletes a collaborator, the special group could be deleted, which would effectively unshare everything that the user has ever shared with that collaborator.

ash3rz: To close the loop, we're thinking of implementing the following changes:
1) When a collaborator is removed from the default set of Collaborators, a warning will be displayed that tells the user that removing the collaborator will have no effect on existing permissions with that collaborator.
2) When a collaborator list is modified/deleted, the user will be presented with the option to either preserve the existing permissions or remove them.

ash3rz:
I don't believe there were any questions/comments on the code itself and we have a plan moving forward to make consequences of modifying/deleting collaborator lists clear, so I'm going to merge this PR.

PR #168, opened 2017-06-30
CORE-8928 Convert Collaborators dialog into a window

ash3rz:
NOTE: This is a branch on top of a branch.  Only the last commit is required for review.

Whether we decide to merge Teams into the Collaboration window or not, with the ability to create and update lists, it's probably better to convert the current Collaborators dialog into a window so it can be more interactive.


ash3rz:
Thanks for the review!

PR #169, opened 2017-06-30
Core 8929 More static id for Manage Tools Window and Tool Listing Dialog.

sri2k1us:


sri2k1us:
Thanks for the review.

PR #170, opened 2017-07-03
core-8926 Fix restricted char display.

sri2k1us:
`\n` and `\t` were not being escaped and hence they were not being displayed in the error message for restricted chars.


sri2k1us:
Looks like i have few more edge cases to before I can put this PR out.

PR #173, opened 2017-07-07
de-1148 fix handling of `&` in query param by Logger Util

sri2k1us:


sri2k1us:
Merging without reviews since this is a very simple change.

PR #175, opened 2017-07-13
core-8648 Do not create new public link for every public link request…

sri2k1us:
…. Instead get existing link or create a new one if none exists.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/grid/GridViewPresenterImpl.java
+++ b/de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/grid/GridViewPresenterImpl.java
@@ -5,17 +5,20 @@ import org.iplantc.de.client.events.diskResources.OpenFolderEvent;
 import org.iplantc.de.client.models.HasId;
 import org.iplantc.de.client.models.HasPath;
 import org.iplantc.de.client.models.dataLink.DataLink;
+import org.iplantc.de.client.models.dataLink.DataLinkFactory;
+import org.iplantc.de.client.models.dataLink.DataLinkList;
 import org.iplantc.de.client.models.diskResources.DiskResource;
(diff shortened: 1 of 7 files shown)


de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/grid/GridViewPresenterImpl.java
```diff
@@ -483,8 +499,55 @@ public void onSuccess(ShareResourceLinkDialog result) {
...
+                                                          diskResourceService.createDataLinks(Arrays.asList(
+                                                                  toBeShared.getPath()),
+                                                                                              new CreateDataLinksCallback());
+                                                      } else {
+                                                          showPublicLink(dlList);
+                                                      }
+
+                                                  }
+
+                                                  private List<DataLink> processResult(String result) {
```
ash3rz: Probably not for this PR, but it looks like at least some of the logic in `processResult` is also in `ListDataLinksCallback`.  It might be nice to refactor this so the logic isn't duplicated, maybe in a callback converter or something.

psarando: The logic in this method looks almost like a copy of the logic in the `ListDataLinksCallback#onSuccess` method.
Could the common logic in these methods be refactored into a `CallbackConverter` in `DiskResourceServiceFacadeImpl#listDataLinks`? Then the converter could parse a `DataLink` list that these other 2 methods could use directly?

sri2k1us: I actually wanted to re-use the entire `ListDataLinksCallback` but it had `Tree` component in it.  Let me see if I refactor some parts of callback into a converter.

PR #176, opened 2017-07-13
CORE-8940 Add static IDs to PreviewApp and PreviewJson views

ash3rz:


ash3rz:
Since these are just static IDs, I'll go ahead and merge.

PR #178, opened 2017-07-18
CORE-8900, 8901, and 8898 - Initial Teams view

ash3rz:
NOTE: Requires ansible changes before merging.  Also note that the commit to actually see Teams in the Collaboration window won't get added until Teams is done.

This isn't what the final product will look like, but here's how Teams look currently (I haven't moved the user search bar into the Collaborators tab or created the search bar for teams yet, so the final view will have the tabs moved up to the very top of the window without anything in between):
image

I also haven't fully flushed out differentiating a group as a list or team for every scenario, but I'm hoping with some endpoint changes that are coming later that can be finalized.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/client/models/collaborators/Subject.java
+++ b/de-lib/src/main/java/org/iplantc/de/client/models/collaborators/Subject.java
@@ -3,7 +3,6 @@ package org.iplantc.de.client.models.collaborators;
 import org.iplantc.de.client.models.HasSettableId;
 
 import com.google.common.base.Strings;
-import com.google.gwt.regexp.shared.MatchResult;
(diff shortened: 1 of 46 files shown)


ash3rz:
Me?  Big PRs?  Never!

I think some of the later PRs start to get messy, so I don't blame you. 😄   Thanks for the review!

psarando:
Ha ha! No problem! And don't get me wrong, I actually think this PR is just right for this feature. I think it's just the nature of boilerplate Java/GWT/GXT. It would actually be more annoying to review a separate PR for AutoBean changes, and a separate one for Appearance changes, then service facade changes, etc. 😉

ash3rz:
Thanks for the review!  Now that the release is over, I'll merge this in.

PR #179, opened 2017-07-25
CORE-8941 Add prompt for retaining permissions when deleting lists or list members

ash3rz:


de-lib/src/main/java/org/iplantc/de/theme/base/client/collaborators/CollaboratorsDisplayStrings.properties
```diff
@@ -18,4 +18,10 @@ memberDeleteFail=Unable to remove the following collaborators: {0,list}
 membersAddedToGroupSuccess=Added the following members to {0}: {1,list}
 newGroupDetailsHeading=Create a Collaborator List
 onlyDNDToListSupported=Drag and drop a collaborator to a list to add it to that list
+removePermissionsBtn=Remove Permissions
+retainPermissionsBtn=Retain Permissions
+retainPermissionsHeader=Retain Permissions?
+retainPermissionsMessage=When items are shared with a collaborator list, those items are shared directly with the list and not the individuals within the list. \
```
ash3rz: I'm mostly concerned about whether this wording is clear.  Feedback welcome.

ash3rz:
I want to make sure this gets in before the next drop, so I'm going to merge this.  Reviews still welcome, especially on whether the retain permissions wording is clear.

de-lib/src/main/java/org/iplantc/de/theme/base/client/collaborators/CollaboratorsDisplayStrings.properties
psarando: Sounds good to me 👍 
I might have written the 2nd line like the following, but I still think it's clear as is:
> So deleting a member from a list normally results in those items no longer being shared with that member, and deleting the list entirely results in those items no longer being shared with anyone in that list.

ash3rz: Your wording is probably clearer.  I'll update it - thanks!

PR #180, opened 2017-07-26
De 1158 Do not allow users to attempt publish apps / workflow that are not actually publishable

sri2k1us:


de-lib/src/test/java/org/iplantc/de/apps/client/presenter/toolBar/AppsToolbarPresenterImplTest.java
```diff
@@ -193,4 +210,46 @@
...
+        verify(submitAppDialogAsyncProviderMock).get(submitDialogCaptor.capture());
+        submitDialogCaptor.getValue().onSuccess(dlgMock);
+        verify(dlgMock).show(eq(app));
+    }
+
+    @Test
+    @Ignore
+    //TODO: Fix this test
+    /**
+     * This test will fail currently because AlertMessageBox initalization fails. Have to dig up mocking library.
```
ash3rz: You could probably skip that issue by putting the creation of that `AlertMessageBox` in a method that you override during testing.  Or really, I don't think we absolutely need to test the `false` condition since it just displays a warning.  And the `true` condition is tested above, which I think is sufficient.

de-lib/src/main/java/org/iplantc/de/client/services/impl/AppUserServiceFacadeImpl.java
```diff
@@ -439,4 +440,18 @@ public void unshareApp(AppUnSharingRequestList request, DECallback<String> callb
...
    }
+
+    @Override
+    public void isPublishable(String system_id, String id, DECallback<Boolean> callback) {
+        String address = APPS + "/" + system_id + "/" + id + "/is-publishable";
+        ServiceCallWrapper wrapper = new ServiceCallWrapper(address);
+        deServiceFacade.getServiceData(wrapper, new DECallbackConverter<String, Boolean>(callback) {
+            @Override
+            public Boolean convertFrom(String result) {
+                JSONObject obj = JsonUtil.getInstance().getObject(result);
```
psarando: I think it would be better to use an `AutoBean` here instead of `JsonUtil`, especially since the `reason` field was just added by cyverse-de/apps#74

PR #181, opened 2017-07-26
CORE-8902 Create interface for creating teams

ash3rz:
NOTE: This is a branch on top of another branch.  Only the `CORE-8902` commits need to be reviewed.

Here's what the form looks like right now:
image


de-lib/src/main/java/org/iplantc/de/teams/client/views/dialogs/SaveTeamProgressDialog.java
```diff
@@ -0,0 +1,65 @@
...
+
+import com.google.inject.Inject;
+
+import com.sencha.gxt.widget.core.client.box.ProgressMessageBox;
+
+/**
+ * A dialog that indicates to the user which step of the team creation/update process is
+ * in progress
+ */
+public class SaveTeamProgressDialog extends ProgressMessageBox {
```
psarando: 👍

de-lib/src/main/java/org/iplantc/de/teams/client/presenter/EditTeamPresenterImpl.java
```diff
@@ -0,0 +1,302 @@
...
+            view.addNonMembers(Lists.newArrayList(privilege));
+        }
+
+    }
+
+    @Override
+    public void onRemoveMemberPrivilegeSelected(RemoveMemberPrivilegeSelected event) {
+        Privilege privilege = event.getPrivilege();
+        if (EditTeamView.MODE.CREATE == mode) {
+            view.removeMemberPrivilege(privilege);
```
psarando: Maybe I overlooked something, but does this mean you can't remove a member when editing a team? Only when creating a team? So would that mean that once a team is created, members can only "opt-out" and admins can't manually remove anyone after?

de-lib/src/main/java/org/iplantc/de/client/services/GroupServiceFacade.java
```diff
@@ -38,7 +42,14 @@
...
      */
-    void addGroup(Group group, AsyncCallback<Group> callback);
+    void addList(Group group, AsyncCallback<Group> callback);
+
+    /**
+     * Create a Team
+     * @param group
+     * @param callback
+     */
+    void addTeam(Group group, List<PrivilegeType> publicPrivileges, AsyncCallback<Group> callback);
```
ash3rz: Oops, good catch

de-lib/src/main/java/org/iplantc/de/teams/client/EditTeamView.java
```diff
@@ -0,0 +1,115 @@
...
+                                      RemoveNonMemberPrivilegeSelected.HasRemoveNonMemberPrivilegeSelectedHandlers,
+                                      AddPublicUserSelected.HasAddPublicUserSelectedHandlers {
+
+    enum MODE {
+        CREATE,
+        EDIT
+    }
+
+    String ALL_PUBLIC_USERS_NAME = "All Public Users";
+    String ALL_PUBLIC_USERS_ID = "GrouperAll";
```
ash3rz: Probably a good idea, I can refactor for that.

de-lib/src/main/java/org/iplantc/de/teams/client/presenter/EditTeamPresenterImpl.java
ash3rz: My intention was that there are 2 cases where a user is in this dialog - creating a team for the first time or editing a team.  
If you're creating a team, add a member, and then decide to delete the member, that's purely within the UI.  No members or privileges have actually been added to the backend, so we just remove the line from the liststore.
If you're editing a team, then 2 service calls are made sequentially - first to remove the selected user as a member and then remove their privileges.  But those changes are in the next PR for editing teams 😄   This PR is all just for creating a team, which I know is a bit confusing.

de-lib/src/main/java/org/iplantc/de/teams/client/presenter/EditTeamPresenterImpl.java
```diff
@@ -0,0 +1,302 @@
...
+
+    @Override
+    public void onAddPublicUserSelected(AddPublicUserSelected event) {
+        addPublicUser();
+    }
+
+    List<UpdatePrivilegeRequest> convertPrivilegesToUpdateRequest(List<Privilege> privileges) {
+        List<UpdatePrivilegeRequest> updateList = Lists.newArrayList();
+
+        privileges.forEach(new Consumer<Privilege>() {
```
ash3rz: Actually no, I couldn't figure out how to do this with a stream 😄   I'll look into it again though because I do prefer that syntax over `foreach`.

psarando: Oh OK 😄 
In that case, I should point out that in my example `privilege -> { ... return update; }` is meant to contain the same logic as the current `accept` method, except the last `updateList.add(update)` line is replaced by `return update` (and this also means the external `updateList` is no longer required).

de-lib/src/main/java/org/iplantc/de/teams/client/presenter/EditTeamPresenterImpl.java
psarando: Ahh, OK, that makes sense now 👍 
Maybe future-us could use a short 1-liner comment here (and below) mentioning something like `remove directly from store. No service calls required in CREATE mode`?

ash3rz: Sure, never hurts to add comments

ash3rz:
Thanks for the review!

PR #182, opened 2017-07-27
CORE-8958 Add help button to Collaboration window

ash3rz:
Also updated the help button text to mention collaborator lists


ash3rz:
I think this PR is pretty straightforward and I want it in before the next drop, so I'm going to merge it.  Reviews still welcome though.

PR #183, opened 2017-07-28
Core 8946 Enable users to auto generate HT Analysis path list file

sri2k1us:
ht-automation


sri2k1us:
Request a quick review. I would like this to go out on the next drop.

de-lib/src/main/java/org/iplantc/de/client/services/impl/DiskResourceServiceFacadeImpl.java
```diff
@@ -1041,6 +1042,26 @@ public String downloadTemplate(String templateid) {
...
+        address.append("dest=" + URL.encodeQueryString(request.getDest()));
+        address.append("&name-pattern=" + URL.encodeQueryString(request.getPattern()));
+        address.append("&folders-only=" + request.isFoldersOnly());
+        address.append("&recursive=" + true);
+        if (request.getInfoTypes() != null && request.getInfoTypes().size() > 0) {
+            for (String infotype : request.getInfoTypes()) {
+                address.append("&info-type=" + infotype);
+            }
+        }
+        Splittable s = AutoBeanCodex.encode(AutoBeanUtils.getAutoBean(request));
```
ash3rz: I don't see the request splittable `s` being used anywhere.  Can this be removed?

de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/toolbar/ToolbarViewPresenterImpl.java
```diff
@@ -394,4 +410,63 @@ public void onSuccess(String result) {
...
+            public void onFailure(Integer statusCode, Throwable exception) {
+                ErrorHandler.post(htAppearance.requestFailed(), exception);
+            }
+
+            @Override
+            public void onSuccess(String result) {
+                dialog.hide();
+                IplantAnnouncer.getInstance()
+                               .schedule(new SuccessAnnouncementConfig(htAppearance.requestSuccess()));
+                JSONObject jsonObject = JsonUtil.getInstance()
```
ash3rz: I think we should be using a callback converter for the `requestHTPathlistFile` service call instead of using the JsonUtil class here.

de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/toolbar/ToolbarViewPresenterImpl.java
```diff
@@ -394,4 +410,63 @@ public void onSuccess(String result) {
...
+        drFacade.requestHTPathlistFile(request, new DataCallback<String>() {
+            @Override
+            public void onFailure(Integer statusCode, Throwable exception) {
+                ErrorHandler.post(htAppearance.requestFailed(), exception);
+            }
+
+            @Override
+            public void onSuccess(String result) {
+                dialog.hide();
+                IplantAnnouncer.getInstance()
```
ash3rz: We should probably inject the `IplantAnnouncer` instead of calling `getInstance` here.

de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/toolbar/ToolbarViewPresenterImpl.java
```diff
@@ -394,4 +410,63 @@ public void onSuccess(String result) {
         });
     }
 
+    @Override
+    public void onAutomateHTPathlist() {
+        drFacade.getInfoTypes(new DataCallback<List<InfoType>>() {
```
ash3rz: I don't think a test was added for this method?

de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/toolbar/ToolbarViewPresenterImpl.java
```diff
@@ -394,4 +410,63 @@ public void onSuccess(String result) {
...
+                });
+                dialog.show();
+            }
+        });
+
+    }
+
+    private void requestHTPathListCreation(HTPathListAutomationDialog dialog,
+                                           HTPathListRequest request) {
+        dialog.mask(htAppearance.processing());
```
ash3rz: I don't think a test was added for this method either?

de-lib/src/main/java/org/iplantc/de/client/services/impl/DiskResourceServiceFacadeImpl.java
```diff
@@ -1041,6 +1042,26 @@ public String downloadTemplate(String templateid) {
         return address;
     }
 
+    @Override
+    public void requestHTPathlistFile(HTPathListRequest request, DECallback<String> callback) {
+        StringBuilder address =
+                new StringBuilder(deProperties.getDataMgmtBaseUrl() + "path-list-creator?");
+        address.append("dest=" + URL.encodeQueryString(request.getDest()));
+        address.append("&name-pattern=" + URL.encodeQueryString(request.getPattern()));
```
psarando: Does this still work if the user does not enter any pattern?
I'm not sure if the service will throw an error if the URL query is `... &name-pattern=&folders-only=false ...`

de-lib/src/main/java/org/iplantc/de/client/services/impl/DiskResourceServiceFacadeImpl.java
```diff
@@ -1041,6 +1042,26 @@ public String downloadTemplate(String templateid) {
...
+        address.append("&recursive=" + true);
+        if (request.getInfoTypes() != null && request.getInfoTypes().size() > 0) {
+            for (String infotype : request.getInfoTypes()) {
+                address.append("&info-type=" + infotype);
+            }
+        }
+        Splittable s = AutoBeanCodex.encode(AutoBeanUtils.getAutoBean(request));
+        Splittable spath = StringQuoter.createSplittable();
+        ServiceCallWrapper wrapper =
+                new ServiceCallWrapper(POST, address.toString(), spath.getPayload());
```
psarando: I'm not seeing how the paths from `Splittable s` are getting into `spath` here (maybe I'm just in a post-lunch haze)?

Maybe it would be better to use a `HasPaths` AutoBean here to generate the payload?
Then you could call `HasPaths#setPaths` with `request.getPaths`.
Or maybe even encode `request` directly into a `HasPaths` AutoBean?

de-lib/src/main/java/org/iplantc/de/client/services/impl/DiskResourceServiceFacadeImpl.java
sri2k1us: 👍

de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/toolbar/ToolbarViewPresenterImpl.java
sri2k1us: I haven't gotten to writing test yet. I wanted to get PR out in time for Monday drop

de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/toolbar/ToolbarViewPresenterImpl.java
sri2k1us: 👍

de-lib/src/main/java/org/iplantc/de/theme/base/client/diskResource/DiskResourceMessages.properties
```diff
@@ -87,3 +86,18 @@ overWiteMetadata = Some metadata already exists. Do you want to overwrite them ?
 doiRequestFail =  DOI Request failed!
 doiRequestSuccess = Your request for DOI has been submitted! You will receive notification when there are updates to this request.
 fileUploadsSuccess = The file(s) {0,list} were successfully uploaded.
+inputLbl = Select folder(s) / file(s) to process for the HT path list file contents:
```
psarando: Maybe we should be more specific with this line to avoid confusion on how this will work?
Maybe this instruction could say something like
> Select folder(s) whose contents will processed into the HT path list file contents.
> Individual file(s) may also be selected.

The service endpoint's documentation also says the following, which might be helpful, but it's also longer:
>  The paths for folders given in the request will not be included in the results, but if any file paths are included in the request, those paths will also be processed

de-lib/src/main/java/org/iplantc/de/theme/base/client/diskResource/DiskResourceMessages.properties
```diff
@@ -87,3 +86,18 @@ overWiteMetadata = Some metadata already exists. Do you want to overwrite them ?
...
 doiRequestSuccess = Your request for DOI has been submitted! You will receive notification when there are updates to this request.
 fileUploadsSuccess = The file(s) {0,list} were successfully uploaded.
+inputLbl = Select folder(s) / file(s) to process for the HT path list file contents: 
+folderPathOnlyLbl = Include folders only
+folderPathOnlyPrompt = Include only folder path(s) in my HT analysis path list file     
+selectorEmptyText = Drag and drop file(s) / folder(s). 
+patternMatchLbl = Include only file / folder path(s) when file / folder name matches this <a href='https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html' target='_blank'>pattern</a> 
+infoTypeLbl = Include only file path(s) whose infoType(s) match
+destLbl = Select a destination where the HT analysis path list file will be saved:
+patternMatchEmptyText= e.g:\\.csv$
```
psarando: I think a space after `e.g.:` might make the regex here a little more readable.

de-lib/src/main/java/org/iplantc/de/theme/base/client/diskResource/toolbar/ToolbarDisplayMessages.properties
```diff
@@ -48,3 +48,5 @@ needDOI = I need DOI
 editViewMetadataMenuItem = Edit / View Metadata
 downloadTemplateMenuItem = Download Template
 doiUserAgreement = &nbsp;Check here to indicate that you have read and agree to the terms of the <a href='http://www.cyverse.org/data-commons-user-agreement' target='_blank'>Data Commons</a>.
+newPathListMenuText = New HT analysis path list file...
```
psarando: 👍

de-lib/src/main/java/org/iplantc/de/client/services/impl/DiskResourceServiceFacadeImpl.java
sri2k1us: Just found this bug. Fixed it.

de-lib/src/main/java/org/iplantc/de/client/services/impl/DiskResourceServiceFacadeImpl.java
sri2k1us: Hmm Definitely missing some code. Did i delete it by accident? But remember writing logic to build path array.

sri2k1us: Yup. Had deleted with when I was cleaning up some debug code. IntelliJ File history to the rescue.

de-lib/src/main/java/org/iplantc/de/client/services/impl/DiskResourceServiceFacadeImpl.java
sri2k1us: Lol. This was related to the accidental code deletion. It was being used!

de-lib/src/main/java/org/iplantc/de/client/services/impl/DiskResourceServiceFacadeImpl.java
```diff
@@ -1041,6 +1042,48 @@ public String downloadTemplate(String templateid) {
...
+        address.append("&folders-only=" + request.isFoldersOnly());
+        address.append("&recursive=" + true);
+        if (request.getInfoTypes() != null && request.getInfoTypes().size() > 0) {
+            for (String infotype : request.getInfoTypes()) {
+                address.append("&info-type=" + infotype);
+            }
+        }
+        Splittable s = AutoBeanCodex.encode(AutoBeanUtils.getAutoBean(request));
+        Splittable spath = StringQuoter.createSplittable();
+        s.get("paths").assign(spath,"paths");
```
psarando: I'd still prefer to use our `HasPaths` AutoBean here, but this works too.

sri2k1us: So I am just making another change to ` HTPathListRequest`  to extend `HasPaths`. Even with that I don't see how I could HasPaths AutoBean to send paths in the request body. I need to convert it `Splittable` so that I can place the path array under `paths` key.

psarando: I guess I was thinking of something like this, where maybe the request can be encoded then decoded into a new `HasPaths` AutoBean: https://github.com/cyverse-de/ui/blob/2.13.0/de-lib/src/main/java/org/iplantc/de/client/services/impl/DiskResourceServiceFacadeImpl.java#L506

Or even just getting a new `HasPaths` from the factory and setting its paths from the request, then the new `HasPaths` can be encoded as the request body: https://github.com/cyverse-de/ui/blob/2.13.0/de-lib/src/main/java/org/iplantc/de/client/services/impl/DiskResourceServiceFacadeImpl.java#L602

sri2k1us: Right! Since I am using `HTPathListRequest` to collect both query parameter and post body, I thought I could just covert the whole request into `Splittable` and extract just the `paths` which could directly attached as body

sri2k1us:
Thanks for the reviews. I will have another PR for tests.

de-lib/src/main/java/org/iplantc/de/client/services/impl/DiskResourceServiceFacadeImpl.java
ash3rz: Hm... would it make more sense to have them separate then? The `ToolbarViewPresenter` would get the request from the dialog, then call a different method to get the paths.  The presenter would then put the paths in a `HasPaths` object.  Then, the service call would accept the `HtPathListRequest`, `HasPaths`, and the callback.  Then you don't have to do any manipulation here in the service facade.

PR #185, opened 2017-07-31
Create flag for whether group name looks clickable in SubjectNameCell

ash3rz:


de-lib/src/main/java/org/iplantc/de/collaborators/client/views/cells/SubjectNameCell.java
```diff
@@ -50,14 +51,20 @@ public void onBrowserEvent(Context context,
 
     public SubjectNameCell() {
         super(CLICK);
+        this.clickableGroupName = true;
```
psarando: Could these 2 lines be replaced with `this(true)`?

de-lib/src/main/java/org/iplantc/de/theme/base/client/collaborators/cells/SubjectNameCellDefaultAppearance.java
```diff
@@ -42,14 +45,21 @@ public SubjectNameCellDefaultAppearance(IplantResources iplantResources, Templat
...
 
     @Override
-    public void render(SafeHtmlBuilder safeHtmlBuilder, Subject subject, String debugID) {
+    public void render(SafeHtmlBuilder safeHtmlBuilder, Subject subject, String debugID, boolean clickableGroupName) {
         String subjectName = subject.getSubjectDisplayName();
         if (subject.getSourceId().equals(Subject.GROUP_IDENTIFIER)) {
-            safeHtmlBuilder.append(templates.group(iplantResources.list().getSafeUri(),
-                                                   SubjectNameCell.SubjectNameCellAppearance.CLICKABLE_ELEMENT_NAME,
-                                                   subjectName,
-                                                   defaultStyle,
```
psarando: Rather than creating a new `XTemplate` interface method, you could just update this line to `clickableGroupName ? defaultStyle : ""`?

Setting the HTML attribute `class=''` will have the same effect as not adding the `class` attribute at all.

ash3rz: Ah, much cleaner.  I'll refactor

de-lib/src/main/java/org/iplantc/de/collaborators/client/views/cells/SubjectNameCell.java
ash3rz: Also a good point.  I'll refactor.

ash3rz:
I pushed the changes.  I'm going to merge now so it's ready for QA.  Thanks for the review!

PR #187, opened 2017-08-02
CORE-8903 Editing teams

ash3rz:
A lot of the extra changes in `EditTeamPresenterImpl` are due to wanting to organize all the methods in there a little better, so sorry for the clutter.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/client/models/collaborators/Subject.java
+++ b/de-lib/src/main/java/org/iplantc/de/client/models/collaborators/Subject.java
@@ -1,6 +1,7 @@
 package org.iplantc.de.client.models.collaborators;
 
 import org.iplantc.de.client.models.HasSettableId;
+import org.iplantc.de.client.models.groups.Group;
 
 import com.google.common.base.Strings;
 import com.google.gwt.regexp.shared.RegExp;
@@ -65,7 +66,7 @@ public interface Subject extends HasSettableId, HasName {
     void setDisplayName(String name);
 
     default boolean isGroup() {
-        return GROUP_IDENTIFIER.equals(getSourceId());
+        return GROUP_IDENTIFIER.equals(getSourceId()) || this instanceof Group;
     }
 
     default String getSubjectDisplayName() {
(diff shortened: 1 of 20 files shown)


de.properties.tmpl
```diff
@@ -214,7 +214,8 @@ org.iplantc.discoveryenvironment.redis.password = {{ key (printf "%s/de/redis-pa
 ###############################################################################
 # Groups Settings
 ###############################################################################
-org.iplantc.discoveryenvironment.groups.grouper-all-id = {{ key (printf "%s/de/grouper-id"  $base) }}
+org.iplantc.discoveryenvironment.groups.grouper-all-id = {{ key (printf "%s/de/grouper-all-id"  $base) }}
 org.iplantc.discoveryenvironment.groups.grouper-all-display-name = {{ key (printf "%s/de/grouper-all-display-name"  $base) }}
+org.iplantc.discoveryenvironment.groups.grouper-id = {{ key (printf "%s/de/grouper-id"  $base) }}
```
psarando: Don't forget: we didn't add this `de/grouper-id` key in cyverse-de/ansible#46, so now this config should be reading directly from `apps/grouper-user`.

de-lib/src/main/java/org/iplantc/de/client/models/collaborators/Subject.java
```diff
@@ -73,7 +73,7 @@ default String getSubjectDisplayName() {
         if (Strings.isNullOrEmpty(subjectName)) {
             subjectName = getFirstName() + " " + getLastName();
         }
-        if (!isGroup() || !hasGroupLongName(subjectName)) {
+        if (!isGroup() && !hasGroupLongName(subjectName)) {
             return subjectName;
```
ash3rz: Yeah, I need to clean this up - it is absolutely confusing.  This was my lazy way of fixing an issue I had and it makes no sense but it works and then I moved on to the next issue 😄 . 
I'm slightly tempted to ditch the `default` method idea and create a more robust util class that can handle the differences between a `Group` and a `Subject`, but I'm not sure yet.  I'll definitely refactor though.

de-lib/src/main/java/org/iplantc/de/client/services/GroupServiceFacade.java
```diff
@@ -103,7 +112,15 @@
      * @param group
      * @param callback
      */
-    void updateGroup(String originalGroup, Group group, AsyncCallback<Group> callback);
+    void updateList(String originalGroup, Group group, AsyncCallback<Group> callback);
+
+    /**
+     * Update the details of a Collaborator List
```
ash3rz: Oops, copy-paste error 😄

de.properties.tmpl
ash3rz: Ah yes, good catch!

ash3rz:
Thanks for the review!

PR #188, opened 2017-08-02
CORE-8907 Add team search field

ash3rz:
NOTE: This is a branch on top of other branches.  The commits for this PR start at `CORE-8907 Add team search service calls`

This basically adds a search field similar to what we have in the Data/Apps/Analyses windows, but for teams.


ash3rz:
Thanks for the review!  I'll wait till after the other PRs have been reviewed and after the release before merging.

PR #190, opened 2017-08-11
Core 8971 Integrate Intercom API

sri2k1us:
- Inject Intercom javascript only if it is enabled through our properties
- Create a separate module to interact with Intercom API
- Track user activities using Intercom API (if enabled)
- This changes requires Ansible changes to configure and enable Intercom

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/intercom/client/IntercomFacade.java
+++ b/de-lib/src/main/java/org/iplantc/de/intercom/client/IntercomFacade.java
@@ -1,30 +1,70 @@
 package org.iplantc.de.intercom.client;
 
+import org.iplantc.de.shared.DEProperties;
+
+import com.google.web.bindery.autobean.shared.Splittable;
+
 /**
+ * A facade class that manages interaction with Intercom
  * Created by sriram on 8/1/17.
  */
 public class IntercomFacade {
 
-    public static native void login(String userId,
-                                    String email,
-                                    String appId,
-                                    String companyId,
-                                    String companyName) /*-{
-        $wnd.Intercom('boot', {
-            app_id: appId,
-            email: email,
-            user_id: userId,
-            created_at: Date.now(),
-            company: {
-                id: companyId,
-                name: companyName
-            }
-        })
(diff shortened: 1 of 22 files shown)


de-lib/src/main/java/org/iplantc/de/intercom/client/IntercomFacade.java
```diff
@@ -0,0 +1,70 @@
...
+
+import org.iplantc.de.shared.DEProperties;
+
+import com.google.web.bindery.autobean.shared.Splittable;
+
+/**
+ * A facade class that manages interaction with Intercom
+ * Created by sriram on 8/1/17.
+ */
+public class IntercomFacade {
```
ash3rz: Can we make this class Injectable?  It seems most of the places we're using it are in the views, but if we end up needing this in a presenter, then the line where we add the tracking will need to be testable.  I think the current usage (IntercomFacade.trackEvent()) would cause Mockito to throw a NPE.  Also, then we could inject `DEProperties`.

sri2k1us: All the methods and fields are `static` (required for JSNI) in this class so I didn't make it Injectable. Also, I found Juice / GIN static injection is not recommended. See - https://github.com/google/guice/wiki/Injections#static-injections

ash3rz: Ah ok, thanks for the link!

de-lib/src/main/java/org/iplantc/de/notifications/client/utils/NotifyInfo.java
```diff
@@ -53,8 +53,10 @@ protected void positionAnnouncer() {
 
         IplantAnnouncement popup = announcements.peek();
 
+        //we support a minimum of 800 x 600
```
psarando: I thought our minimum window size support was 1024x768?

sri2k1us: Ohh ya good catch. I should test with 1024 x 768 just to make the notification is still visible.

PR #193, opened 2017-08-18
CORE-8978 Update Teams UI to match updated mock up

ash3rz:
This is basically removing the Teams menu, leaving only the `Create Team` button behind.  The rest of the buttons to edit, leave, join, etc. a team will be in the `EditTeamDialog`.

image
image
image


ash3rz:
Ah, just remembered I have one more change I should add to this PR - I'll close then reopen when it's ready.

PR #194, opened 2017-08-23
CORE-8978 Update teams to match updated mock up

ash3rz:
To summarize: 
* The Teams menu is gone and replaced with just the "Create Team" button.  The rest of the menu items (Delete Team, Join Team, etc.) are buttons in the EditTeamDialog that are enabled/disabled based on a user's membership and privileges.
* The details dialog is gone, I'm reusing the EditTeamDialog with a "View" mode for non-admins
* If a user has both `read` and `optin` privileges, they are shown a single fake privilege (`readOptin`) that represents both.  This enables us to stick with the idea that each user has 1 privilege in a hierarchy of privileges.

image

image

image


tedgin:
For the View-only version of the Edit window, could the window be shrunk to eliminate all the empty space? Also, when the user isn't a member of a team, i.e., she doesn't have read access, the Members list pain could be hidden and the window could be further shrunk.

For the

sri2k1us:
Looks good to me. But the screen shot-2 I see some text being cut-off on the right side of the screen. Not sure if thats problem with screen shot or there is a rendering problem

ash3rz:
Makes sense - I'll look into adjusting the height of the window depending on what the user can see.

And yes, looks to be a rendering problem.  I'll fix before merging.

Thanks for the reviews!

ash3rz:
I pushed the two fixes for the comments I received.  I'll go ahead and merge now - thanks for the reviews!

PR #195, opened 2017-08-23
CORE-8977 Deleting a team

ash3rz:
This enables an admin to delete a team they created.


ash3rz:
Thanks for the review!

PR #196, opened 2017-08-24
core-8926 fix truncated validation message when tool name begins with

sri2k1us:


sri2k1us:
Directly merging since they changes are so minor.

PR #197, opened 2017-08-28
Core 8989 Allow users to reset agave token from the DE.

sri2k1us:


de-lib/src/test/java/org/iplantc/de/preferences/client/presenter/PreferencesPresenterImplTest.java
```diff
@@ -46,8 +58,8 @@ public void setUp() {
         when(viewMock.getValue()).thenReturn(userSettingsMock);
 
         uut = new PreferencesPresenterImpl(viewMock,
-                                           serviceFacadeMock,
-                                           appearanceMock);
+                                           serviceFacadeMock, oauthServiceFacadeMock,
+                                           appearanceMock, constantsMock);
```
ash3rz: Also minor, but each parameter should probably be on its own line.

de-lib/src/main/java/org/iplantc/de/client/services/impl/OauthServiceFacadeImpl.java
```diff
@@ -22,12 +26,14 @@
 
     private final DiscEnvApiService deServiceFacade;
     private CommonModelAutoBeanFactory factory;
+    private final DEClientConstants constants;
 
     @Inject
     public OauthServiceFacadeImpl(final DiscEnvApiService deServiceFacade,
-                                  CommonModelAutoBeanFactory factory) {
+                                  CommonModelAutoBeanFactory factory, final DEClientConstants constants) {
```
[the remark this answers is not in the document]
sri2k1us: 👍 . Wonder why Idea wouldn't format parameters this way automatically. I will look in to it.

de-lib/src/test/java/org/iplantc/de/preferences/client/presenter/PreferencesPresenterImplTest.java
sri2k1us: 👍

PR #198, opened 2017-08-29
CORE-8899 Allow users to request to join a team, allow team admins to approve or deny those requests

ash3rz:
NOTE: This is a branch on top of a branch. The commits for this PR start at `CORE-8899 Add service calls for joining a team`


psarando:
LGTM 👍

de-lib/src/main/java/org/iplantc/de/notifications/client/utils/NotificationUtil.java
```diff
@@ -0,0 +1,272 @@
...
+                GWT.log("NEW  category");
+                break;
+
+            case SYSTEM:
+                GWT.log("SYSTEM  category");
+                break;
+
+            case PERMANENTIDREQUEST:
+                GWT.log("Permanent Id request category");
+                msg.setContext(payload.getPayload());
```
psarando: I guess we should keep on eye on notifications to make sure they're still working as expected, since this looks slightly different than the code it's replacing in the `NotificationCallbackWrapper`: for example, `PERMANENTIDREQUEST` payloads weren't getting set in that callback, but that might not make a difference or break anything.

ash3rz: Good catch.  I did notice (I think it was this PR) that some of the notification filters weren't working so it's possible I fixed one thing but broke something else so I'll double check 👍

ash3rz: This was actually a really good catch - because this logic was sort of duplicated in 2 different areas (once for the Notification window, and again for the Notification dropdown on the desktop) there were a couple things that worked in one spot and not the other.  I'll put up a PR for fixes soon.

PR #200, opened 2017-08-29
CORE-8990 Refactor Subject name comparator to fix NPE

ash3rz:
Fixes NPE for subject entities that don't have a first or last name.

Now comparing group names, then last name, then first name, then name, then id


sri2k1us:
LGTM but given that there are too many logical paths (if / else), it would be nice to write some test around this code.

ash3rz:
Ah, good point.  I'll add tests.

ash3rz:
I added tests, so I think I'll merge now. Thanks for the review!

PR #202, opened 2017-08-30
core-8989 add static id to reset button

sri2k1us:


sri2k1us:
Merging directly since this code just add static id.

PR #205, opened 2017-09-07
CORE-8994 Switch apps listings to use "dot" menu

ash3rz:
Tile view:
image

Grid view:
image


ash3rz:
Thanks for the review!

PR #206, opened 2017-09-11
CORE-4434 Move user search bar to the far left in the sharing view

ash3rz:


ash3rz:
Thanks for the review!

PR #207, opened 2017-09-12
CORE-8917 Enable subject search highlighting on searched fields

ash3rz:
image

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/theme/base/client/apps/AppSearchHighlightDefaultAppearance.java
+++ b/de-lib/src/main/java/org/iplantc/de/theme/base/client/apps/AppSearchHighlightDefaultAppearance.java
@@ -2,6 +2,8 @@ package org.iplantc.de.theme.base.client.apps;
 
 import com.google.common.base.Strings;
 import com.google.gwt.regexp.shared.RegExp;
+import com.google.gwt.safehtml.shared.SafeHtml;
+import com.google.gwt.safehtml.shared.SafeHtmlUtils;
 
 /**
  * @author jstroot
@@ -13,10 +15,10 @@ public class AppSearchHighlightDefaultAppearance implements AppSearchHighlightAp
     private RegExp regExp;
 
     @Override
-    public String highlightText(String name, String pattern) {
+    public SafeHtml highlightText(String name, String pattern) {
         if(Strings.isNullOrEmpty(pattern)){
             regExp = null;
-            return name;
+            return SafeHtmlUtils.fromTrustedString(name);
         }
 
         // XXX JDS Keep an eye on performance.
@@ -26,6 +28,6 @@ public class AppSearchHighlightDefaultAppearance implements AppSearchHighlightAp
             || !regExp.getSource().equals(pattern)){
             regExp = RegExp.compile(pattern, "ig");
         }
-        return regExp.replace(name, REPLACEMENT_START + "$1" + REPLACEMENT_END);
+        return SafeHtmlUtils.fromTrustedString(regExp.replace(name, REPLACEMENT_START + "$1" + REPLACEMENT_END));
     }
 }
--- a/de-lib/src/main/java/org/iplantc/de/theme/base/client/apps/AppSearchHighlightDefaultAppearance.java
+++ b/de-lib/src/main/java/org/iplantc/de/theme/base/client/apps/AppSearchHighlightDefaultAppearance.java
@@ -25,8 +25,12 @@ public class AppSearchHighlightDefaultAppearance implements AppSearchHighlightAp
 
         // Initialize or recompile regExp if necessary
         if(regExp == null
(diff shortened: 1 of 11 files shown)


de-lib/src/main/java/org/iplantc/de/theme/base/client/apps/AppSearchHighlightDefaultAppearance.java
```diff
@@ -26,6 +28,6 @@ public String highlightText(String name, String pattern) {
             || !regExp.getSource().equals(pattern)){
             regExp = RegExp.compile(pattern, "ig");
         }
-        return regExp.replace(name, REPLACEMENT_START + "$1" + REPLACEMENT_END);
+        return SafeHtmlUtils.fromTrustedString(regExp.replace(name, REPLACEMENT_START + "$1" + REPLACEMENT_END));
```
ianmcorvidae: This is sort of a dumping-ground for some thoughts, not quite sure where to put them.

* What happens if the search term is itself something which includes regular expression metacharacters? i.e., what would this code do if I searched for say `fo*` -- would the highlighting then highlight `foo`? Similarly, what if the search term is an invalid regular expression pattern, e.g. `[la` (I made a user who comes up with this search)
* What happens with multi-word searches? For some reason I can't get `mcewen ian` to work, but I do come up if you search `mcew ia`. I'm guessing I'd not get highlighted there (not that it'd be an issue since the name's always visible, but you see what I'm getting at I'm sure)
* is there any way to HTML inject with this, since it's pulling from a trusted string? I guess only if there's a search result which matches, but I'm not sure. Anyway, my test user is named `[la <script type="text/javascript">alert("Hello")</script>` so hopefully that'll get tested. I couldn't get it to with the current stuff but maybe the highlighting will change that, I dunno.

ash3rz: These are really good questions and highlighting some breakages between this and prod - mainly searching with regular expressions (which it seems you can really only search with a `*` as a wildcard character from what I can see in prod i.e. `ra*sey` will yield results with "ramsey" and also "Sa**ra**h Demp**sey**", but `ra?sey` yields nothing, same with `rams[ea]y`).  I'll dig into this and push some fixes.

As for your real name, it would seem that's a backend issue. "Ian McEwen" finds you but "McEwen Ian" returns back an empty response.

ianmcorvidae: Ah, yeah, I figured it not showing up with the full backwards name was the backend somehow. I mentioned searching `mcew ia` for that reason; I do show up in that, but I think it wouldn't get highlighted, much like your `ra*sey` search probably wouldn't highlight much. Maybe a better example for me would be `hm0 ariz`, which also brings me up but definitely wouldn't get highlighted (and without the processing/highlighting it wouldn't be clear why it matched). `hm0*ariz` also matches but interestingly *only* me, so apparently multiple words isn't quite treated the same as a wildcard.

Basically there's weird stuff that happens with spaces and asterisks, anyway. It seems like the potential way to fix this would be to split the search on whitespace and asterisks (maybe more? maybe just all nonalphabetic characters?) and highlight each of those sections individually, but of course then you also have to make sure you don't end up with overlaps and annoyances like that. Might be a later-pass sort of thing but I wanted to draw attention to it even if it's just an issue we'll deal with until a later date.

de-lib/src/main/java/org/iplantc/de/theme/base/client/collaborators/util/UserSearchFieldDefaultAppearance.java
```diff
@@ -20,41 +24,44 @@
...
-            sb.append(userTemplate.render(name, subject.getInstitution(), null));
-        }
+    public void render(Cell.Context context, Subject subject, String searchTerm, SafeHtmlBuilder sb) {
+        SafeHtml name = getHighlightedName(subject, searchTerm);
+        String institution = getInstitution(subject);
+        SafeHtml email = getHighlightedEmail(subject, searchTerm);
+        SafeHtml id = getHighlightedId(subject, searchTerm);
+        SafeUri uri = getImageUri(subject);
+
+        sb.append(userTemplate.render(name, institution, email, id, uri));
```
psarando: Tidy! ✨

ash3rz:
I pushed an additional commit for this item.  It's not pretty by any means, and I'm sure there exists an easier way to accomplish this, but... it works 😄   I haven't figured out how to tackle proper highlighting for the case that Ian pointed out with regex characters in the results, but maybe that can wait for a later time.

ianmcorvidae:
I think you're right and the remaining cases which could cause issues are obscure and stupid enough we don't need to worry about it!

ash3rz:
Thanks for the reviews!

PR #208, opened 2017-09-12
DEOPS-87: implement an authentication success handler with a configur…

slr71:
…able base URL


slr71:
Thanks for the review!

PR #210, opened 2017-09-12
CORE-4434 Fix issue where the explain button would appear if there we…

ash3rz:
…re no shares


ash3rz:
I decided to stop fighting GXT's `forceLayout()` method which wasn't working for me and focus on the logic... hopefully this should fix it.  Will go ahead and merge since the change is so small.

PR #211, opened 2017-09-13
CORE-8996 Move Data window icons behind a dot menu

ash3rz:
image


de-lib/src/main/java/org/iplantc/de/theme/base/client/diskResource/grid/cells/DiskResourceDotMenuDisplayStrings.properties
```diff
@@ -0,0 +1,2 @@
+dataLinkText=Create / View Public Link
```
ianmcorvidae: Tiny, but it's probably most accurate to say "Public Links", since the dialog it brings up will let you create several even for one file.

de-lib/src/main/java/org/iplantc/de/theme/base/client/diskResource/DiskResourceMessages.properties
```diff
@@ -100,4 +100,5 @@ requestSuccess = Your request completed successfully!
 requestFailed = Unable to generate HT analysis path list file. Please try again!
 validationMessage = Please complete all the required fields!
 select = select...
+featureNotSupported=Feature not supported - invalid resource name
```
ianmcorvidae: When does this come up? I only see this line and the one in the corresponding java file, nothing that uses it. But I'm assuming something must.

de-lib/src/main/java/org/iplantc/de/theme/base/client/diskResource/grid/cells/DiskResourceDotMenuDefaultAppearance.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    @Override
+    public ImageResource dataLinkIcon() {
+        return iplantResources.linkIcon();
+    }
+
+    @Override
+    public String shareText() {
+        return diskResourceMessages.share();
```
ianmcorvidae: I think this string is a change, and this previously said "Share with Collaborators..." (as opposed to "share via public link" or similar). I don't know that I see this as a problematic change but I didn't see it mentioned so I wanted to ask if it was intentional.

(I also notice the public link one changed, but it seems the previous text was "Share Path", which is less than useful, so...)

de-lib/src/main/java/org/iplantc/de/theme/base/client/diskResource/grid/cells/DiskResourceDotMenuDefaultAppearance.java
```diff
@@ -0,0 +1,112 @@
...
+        appDotMenuDefaultAppearance.render(context, sb, debugId);
+    }
+
+    @Override
+    public ImageResource dotMenuIcon() {
+        return appDotMenuDefaultAppearance.dotMenuIcon();
+    }
+
+    @Override
+    public String favoriteText(DiskResource diskResource) {
```
ianmcorvidae: (semi-random spot, I mostly am reviewing from the screenshot)

I think this string has a period at the end, where everything else in the menu doesn't. It looks a bit weird that way, not sure the ideal way to fix (I personally like the ones without periods better)

ash3rz: I like them better without the periods too, I'll update 👍

de-lib/src/main/java/org/iplantc/de/theme/base/client/diskResource/DiskResourceMessages.properties
ash3rz: Ah, this is leftover.  I was going to have all the icons available but with this error message for any disk resources with bad names, but then I realized we disable everything and remove the icons in that case now so I went ahead and made it so that the dot menu wouldn't even render. I'll remove this.

de-lib/src/main/java/org/iplantc/de/theme/base/client/diskResource/grid/cells/DiskResourceDotMenuDisplayStrings.properties
ash3rz: At least in prod, this brings up the smaller dialog with just a single public link that you can copy and paste somewhere else, not the view where you can choose to create or destroy public links.

ianmcorvidae: Oh, you're right. I forgot about that dialog, I think, heh.

de-lib/src/main/java/org/iplantc/de/theme/base/client/diskResource/grid/cells/DiskResourceDotMenuDefaultAppearance.java
ash3rz: It should probably be "Share with Collaborators..." to be consistent with the menu, so I'll change that.  The public link one is slightly different than the menu item since the menu item brings up the full editing links view whereas this either creates a public link or fetches a public link and shows it so... I'm not sure.  Maybe it'd be better to have the text match and update this so they show the same view?

ianmcorvidae: I think it makes sense to keep the public links as two different things (as I mentioned I think I forgot the more shortcut-style simple dialog existed, heh). I guess for this exact line it makes sense to have the menu and this match, if they bring up the same dialog, though?

ash3rz: Yeah, if they bring up the same thing, then the text should match, I agree.  OK, I'll keep the shortcut link dialog and keep the text slightly different then for this case.

ash3rz:
Thanks for the review (from earlier)!

PR #192, opened 2017-08-17
CORE-8980 Add Ontology Lookup Service autocomplete widget to DE Metadata Template forms

psarando:
This PR will add a `MetadataTermSearchField` widget that can create a ComboBox for use as a Metadata Template attribute type.

The ComboBox can lookup terms using the Ontology Lookup Service ["select" endpoint](http://www.ebi.ac.uk/ols/docs/api#_select_terms) to populate its drop-down list. This ComboBox will allow the user to either select one of the suggested terms, or enter any other text they wish as the metadata attribute's value.

The `MetadataTermSearchField` can also be used as a `Converter` for the ComboBox widget in a GridRowEditing column, such as in the AVU editing grid.

OLS ComboBox screen shot

The OLS searches are restricted to only `class` entity types by default. The search results are presented in the ComboBox drop-down with the ontology class `label` listed first in bold, and underneath the `label` the `ontology_prefix` and `IRI` are displayed in italics.

PR #213, opened 2017-09-19
CORE-8982 Add Ontology Lookup Service autocomplete widget to Belphegor

psarando:
Belphegor can now select `OLS_ONTOLOGY_TERM` as a Metadata Template Attribute type, and admins may configure the OLS query param settings for each OLS attribute in a template.

**Note**: This PR requires changes in cyverse-de/metadata#8 be merged first.

The `Value(s)` column in the attributes grid of the `Edit Template Attributes` dialog has been updated to display either `Enum value(s)` for enum attribute types, `Query settings` for `OLS_ONTOLOGY_TERM` types, or no string for all other types. The button in this column will also now ignore clicks for other attribute types:
screenshot-2017-9-18 tool admin ols attr type

Separate OLS query parameters can be saved by admins for each OLS attribute in a Metadata Template. When DE users edit values for these attributes types using the Metadata Template view (see cyverse-de/ui#192), these settings will be used in the OLS query to populate the term suggestions:
screenshot-2017-9-18 tool admin edit ols query params


de-lib/src/main/java/org/iplantc/de/admin/desktop/client/metadata/view/OntologyLookupServiceParamListEditor.ui.xml
```diff
@@ -0,0 +1,56 @@
...
+    <form:FieldSet ui:field="header" layoutData="{layoutData}" >
+        <container:VerticalLayoutContainer>
+            <container:child layoutData="{layoutData}">
+                <form:FieldLabel ui:field="helpLabel"
+                                 labelAlign="TOP"/>
+            </container:child>
+            <container:child layoutData="{layoutData}">
+                <toolbar:ToolBar>
+                    <c:TextButton ui:field="addBtn"
+                                  debugId='idAddBtn'
```
ash3rz: I think we're leaning towards not using this type of debug ID because it's easy to create duplicate IDs this way.  I'm not sure how much testing goes into Belphy, but if we do want debug IDs, it should be through the `ensureDebugId()` methods that append names to the parent element names.

de-lib/src/main/java/org/iplantc/de/admin/desktop/client/metadata/view/EditMetadataTemplateViewImpl.java
```diff
@@ -219,7 +236,12 @@ public void onSelect(SelectEvent event) {
                 Context c = event.getContext();
                 int row = c.getIndex();
                 MetadataTemplateAttribute mta = store.get(row);
-                editEnumValues(mta);
+
+                if (MetadataTemplateAttributeType.ENUM.toString().equals(mta.getType())) {
+                    editEnumValues(mta);
+                } else if (MetadataTemplateAttributeType.OLS_ONTOLOGY_TERM.toString().equals(mta.getType())) {
+                    editOntologyLookupServiceParams(mta);
+                }
```
ash3rz: Not necessarily a change for this PR, but I think this whole if-else block should probably be in the presenter.  I think we're moving towards the pattern of having the presenter be in control of any dialogs being shown.

de-lib/src/main/java/org/iplantc/de/admin/desktop/client/metadata/view/OntologyLookupServiceParamListEditor.ui.xml
psarando: Good catch. I think this was a case of clipboard inheritance from [SubmitAppForPublicUseView.ui.xml](https://github.com/cyverse-de/ui/blob/57d07ab9fdf546a59dfbe910c8477960e74c5216/de-lib/src/main/java/org/iplantc/de/apps/client/views/submit/SubmitAppForPublicUseView.ui.xml#L145-L153).

I look into updating this before merging 👍

de-lib/src/main/java/org/iplantc/de/admin/desktop/client/metadata/view/EditMetadataTemplateViewImpl.java
psarando: I was considering that too, but it doesn't look like there are any presenters for [EditMetadataTemplateView](https://github.com/cyverse-de/ui/blob/57d07ab9fdf546a59dfbe910c8477960e74c5216/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/metadata/view/EditMetadataTemplateView.java) right now.

This class could also use some appearance refactoring, so that and the presenter might be saved for another PR...

ash3rz: Heh, it was early so I didn't even notice there's no presenter 😄   Yeah, makes sense for another PR.

sri2k1us: LGTM :shipit:

psarando:
Thanks for the reviews! 👍

PR #215, opened 2017-09-25
MISC Notification fixes

ash3rz:
Mostly some notification fixes, but also adding an error message in the Data window if the listing fails (like we've been seeing in the dev environments, but currently without any indication to the user that the listing failed)


ash3rz:
I forgot I do actually want this in for the release since it has some fixes, so I'm going to merge since the changes are relatively small.  Reviews still welcome.

PR #216, opened 2017-09-27
DE-1164 Fix logic for comparing Sharing equality

ash3rz:
The `equals` logic was slightly incorrect because it was only checking if the user and the permission value were the same between two `Sharing` objects.  However, in the case we just discovered where a user may want to change the permissions for multiple files at a time for one user, it is necessary to check the user, the permission value, and also that the resource is the same between the two objects being compared.


ianmcorvidae:
Looks like a straightforwardly reasonable change to me! :shipit:

ash3rz:
Thanks for the review!

PR #217, opened 2017-10-03
CORE-9000 + CORE-9001: Add 'Unified Astronomy Thesaurus' ontology term autocomplete support to DE and Belphegor Metadata Template forms

psarando:
This PR will add [Unified Astronomy Thesaurus](https://vocabs.ands.org.au/the-unified-astronomy-thesaurus) ontology term autocomplete support to DE and Belphegor Metadata Template forms.

This Metadata Template attribute field looks and works similar to the "Ontology Lookup Service" attribute field added by cyverse-de/ui#192, except the UAT search service offers much more limited search parameters, so there are no query param settings available for UAT term attributes in Belphegor.

This PR may be easier to review by each individual commit.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/server/auth/BaseUrlConnector.java
+++ b/de-lib/src/main/java/org/iplantc/de/server/auth/BaseUrlConnector.java
@@ -14,6 +14,7 @@ import org.apache.http.client.methods.HttpPut;
 import org.apache.http.client.methods.HttpRequestBase;
 import org.apache.http.client.params.HttpClientParams;
 import org.apache.http.client.utils.URIBuilder;
+import org.springframework.beans.factory.annotation.Value;
 
 import java.io.IOException;
 import java.net.URISyntaxException;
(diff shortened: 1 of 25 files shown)


de-lib/src/main/java/org/iplantc/de/server/auth/BaseUrlConnector.java
```diff
@@ -120,7 +123,11 @@ protected String addQueryParam(String uriString, String name, String value) thro
      * @throws IOException if a URI representation is invalid or an encoding error occurs.
      */
     protected String addIpAddress(String uriString, HttpServletRequest request) throws IOException {
-        return addQueryParam(uriString, "ip-address", request.getRemoteAddr());
+        if (uriString.startsWith(terrainBaseUrl)) {
+            return addQueryParam(uriString, "ip-address", request.getRemoteAddr());
```
psarando: This change was required because the [UAT search endpoint](https://vocabs.ands.org.au/repository/api/lda/aas/the-unified-astronomy-thesaurus/2-0-0/concept.json?_pageSize=50&labelcontains=star) would return a `Bad Request` error if the `ip-address` parameter was included. This should also prevent our users' IP addresses from being sent to any other 3rd-party service endpoints.

slr71: That makes sense. The only endpoint we really need the IP address for is the bootstrap endpoint anyway. 👍

psarando:
Great, thanks for all the reviews! 👍

PR #219, opened 2017-10-04
CORE-9007, CORE-9009 Teams help text update and bug fix 

ash3rz:
The bug fix was for the fake privilege "readOptIn" which needs to be split into "read" and "optin" which I missed in one spot.


tedgin:
LGTM :shipit:

ash3rz:
Thanks for the review!

PR #220, opened 2017-10-05
Misc fixes

ash3rz:
* Added some missing loading masks for teams
* Fixed the text for data links in the Data window dot menu
* Also disabled the data link button in the Data window dot menu if the user has a file selected that they do not have own permissions on

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/diskResource/client/views/widgets/DiskResourceDotMenu.java
+++ b/de-lib/src/main/java/org/iplantc/de/diskResource/client/views/widgets/DiskResourceDotMenu.java
@@ -51,7 +51,7 @@ public class DiskResourceDotMenu extends Menu {
 
     void addMenuItems(DiskResource diskResource) {
         favoriteBtn = new MenuItem(appearance.favoriteText(diskResource), appearance.favoriteIcon(diskResource));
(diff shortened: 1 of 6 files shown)


de-lib/src/main/java/org/iplantc/de/diskResource/client/views/widgets/DiskResourceDotMenu.java
```diff
@@ -65,6 +66,9 @@ void addMenuItems(DiskResource diskResource) {
 
     void checkButtonStatus(DiskResource diskResource) {
         shareBtn.setEnabled(isOwner(diskResource));
+        if (diskResource instanceof File) {
+            dataLinkBtn.setEnabled(isOwner(diskResource));
```
ianmcorvidae: The service level only checks for the user having write permissions; presumably this should match (unless the previous code only allowed this for ownership, in which case I guess we should retain that?)

ash3rz: Yeah, the previous code (and I tested it in prod just to be sure) only enables that option with `own`, which is strange if the service is different.  At least for now, the old and new match which I think is ok.

ash3rz:
Thanks for the review!

PR #222, opened 2017-10-06
Switch to new gradle plugin that supports JsInterop

ash3rz:
As I said in chat, I tested this with Paul's React POC PR and everything seemed to work just fine.  SDM is working.  Prod builds seem to be working.

I did see an issue filed in the gradle plugin [repo](https://github.com/jiakuan/gwt-gradle-plugin) about the *nocache.js file not being generated, but I see it getting loaded into the browser when I do a production build so I don't think it's an issue for us.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,13 +2,10 @@ buildscript {
     repositories {
         mavenCentral()
         jcenter()
-        maven {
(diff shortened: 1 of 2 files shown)


ash3rz:
Thanks for the review!

PR #226, opened 2017-10-17
core-7445 fix filename logic.

sri2k1us:


sri2k1us:
Merging directly since this is minor change.

de-lib/src/main/java/org/iplantc/de/theme/base/client/tools/NewToolRequestFormViewDefaultAppearance.java
```diff
@@ -114,9 +114,9 @@ public String fileNameValidationMsg() {
 
     @Override
     public String getFileName(String filename) {
-        if (Strings.isNullOrEmpty(filename) || !GXT.isChrome()) {
+        if (Strings.isNullOrEmpty(filename) || !filename.contains(BROWSER_FAKEPATH)) {
```
psarando: Since we're always returning `filename.substring(12)` if this check fails, maybe should check for `filename.startsWith(BROWSER_FAKEPATH)` instead of `contains`?

sri2k1us: Good point 👍 . I will make that change.

PR #227, opened 2017-10-17
CORE-9038 Prevent user from seeing default collab list in search results

ash3rz:
There's no real harm in the user seeing the `default` collaborator list, it's just that they also can't really do anything with it.  If they click on it, they get an error message about being unable to add `default` to their collaborators (which makes sense since they're trying to add `default` to `default` without knowing it).


ash3rz:
Thanks for the review!

PR #230, opened 2017-10-18
CORE-9041 Clear out the teams grid if no search results are returned for clarity

ash3rz:
I also added in a change so that if you clear out the search text, the "All Teams" filter will be applied.
Then I also realized the team creator names were missing when searching, so I added that.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/teams/client/presenter/TeamSearchRpcProxy.java
+++ b/de-lib/src/main/java/org/iplantc/de/teams/client/presenter/TeamSearchRpcProxy.java
@@ -114,7 +114,12 @@ public class TeamSearchRpcProxy extends RpcProxy<FilterPagingLoadConfig, PagingL
     }
 
     void addCreatorToTeams(List<Group> teams, FastMap<Subject> creatorFastMap) {
-        teams.forEach(group -> group.setCreator(creatorFastMap.get(getCreatorId(group)).getSubjectDisplayName()));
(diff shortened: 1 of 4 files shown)


de-lib/src/main/java/org/iplantc/de/teams/client/presenter/TeamSearchRpcProxy.java
```diff
@@ -65,17 +72,66 @@ public void onFailure(Throwable caught) {
...
     PagingLoadResultBean<Group> getLoadResult(List<Group> result, FilterPagingLoadConfig loadConfig) {
         return new PagingLoadResultBean<>(result, result.size(), loadConfig.getOffset());
     }
 
+    List<String> getCreatorIds(List<Group> teams) {
+        return teams.stream().map(this::getCreatorId).distinct().collect(Collectors.toList());
+    }
+
+    void addCreatorToTeams(List<Group> teams, FastMap<Subject> creatorFastMap) {
+        teams.forEach(group -> group.setCreator(creatorFastMap.get(getCreatorId(group)).getSubjectDisplayName()));
```
psarando: Since `getCreatorId` has the possibility to return `null`, it seems like this line could throw an NPE (either by the `FastMap` or from the `getSubjectDisplayName` call).

Maybe we should check the return value of `getCreatorId` for null?

ash3rz:
Thanks for the review!

PR #231, opened 2017-10-19
Core 9039 Display tool restrictions in tool info dialog.

sri2k1us:


de-lib/src/main/java/org/iplantc/de/tools/client/views/manage/ToolInfoViewImpl.java
```diff
@@ -70,6 +71,32 @@ private void buildToolInfoView(Tool tool) {
...
+                                                                         tool.getTimeLimit())),
+                          new AbstractHtmlLayoutContainer.HtmlData(".cell13"));
+        htmlContainer.add(new Label(
+                                  appearance.networkingMode() + ((Strings.isNullOrEmpty(tool.getContainer()
+                                                                                            .getNetworkMode())
+                                                                  || tool.getContainer()
+                                                                         .getNetworkMode()
+                                                                         .equalsIgnoreCase(appearance.bridge())) ?
+                                                                 appearance.enabled() :
+                                                                 appearance.disabled())),
```
ash3rz: This might just be a preference but for these 5 new Labels I think it makes more sense to have the appearance return the entire string result. 
So something like `htmlContainer.add(new Label(appearance.networkingMode(tool.getContainer.getNetworkMode()))`. 
This way, the view isn't building what the final string will be, that responsibility goes to the appearance and everything can be found in the appearance.

sri2k1us: Right! But I was thinking those strings are different. Appearance tells me what to display and view can decide how to display.

ash3rz: Sure, I guess in my mind I see it as the view chooses where the Label is in the layout (the `how to display`), and then the appearance decides `what to display` by deciding in full what the text will be. So if the text needed to change, I would expect there to be a change in the appearance not the view. Again, it might just be my own personal preference. PR is good to merge 👍

sri2k1us:
Thanks for the reviews!

de-lib/src/main/java/org/iplantc/de/tools/client/views/manage/ToolInfoViewImpl.java
psarando: Also I think it's useful if the appearance can also decide where the value goes in the label string.
Right now, the code is assuming `labelPrefix + value`, but if we pass the value to the appearance, then it can decide where to position it in the future: `appearance.networkingMode(networkingModeValue)`.

I also think this code would look cleaner if the label-value-logic was refactored out of the `htmlContainer.add(new Label(...))` calls:
```java
String networkMode = tool.getContainer().getNetworkMode();
String networkingModeValue = (Strings.isNullOrEmpty(networkMode) || networkMode.equalsIgnoreCase(appearance.bridge())) ?
    appearance.enabled() :
    appearance.disabled());
```

sri2k1us: I guess I was going for concise 1 liners. But coming back to appearance,  I think there is a basic assumption that `label` and `appearance` is going to be in one line. But that can change as well. For example
>Networking:
>   Disabled

In this case, I just have to change the underlying html template and not have rewrite the appearance. May be I could move the logic of what should be displayed for value to appearance. That way I could keep still keep the `label` and `value` separate in appearance but code is little bit cleaner in view.

psarando: > In this case, I just have to change the underlying html template

👍 

Speaking of concise 1-liners:
```java
String networkingModeValue = (appearance.bridge().equalsIgnoreCase(tool.getContainer().getNetworkMode()) ? appearance.enabled() : appearance.disabled());
```
I have no strong opinion on where this logic goes.

PR #232, opened 2017-10-20
de-1165 Do not allow space in bioproject name. Refactor to use injection

sri2k1us:


psarando:
Are we sure this is really a bug that needs to be fixed?

I just tested this in prod, and I was able to create a metadata JSON file from a BioProject that has a space in the name (`My BioProjectTest`), and that JSON file passed the metadata validation of the SRA submission app. Also I don't think the name of the folder is used in the submission anyway. Only the metadata on the folders are submitted, and the XSD allows spaces in the project title: https://github.com/iPlantCollaborativeOpenSource/docker-builds/blob/7855c29a03c3cc8b018a1ef0f4f8513b5631b394/ncbi-submit/schemas/SP.common.xsd#L274-L280.

de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/toolbar/ToolbarViewPresenterImpl.java
```diff
@@ -217,31 +220,29 @@ protected CreateFolderDialog getCreateFolderDialog(Folder selectedFolder) {
...
+                        parentPresenter.onCreateNcbiSraFolderStructure(selectedFolder,
+                                                                       dialog.getProjectTxt(),
+                                                                       dialog.getBiosampNum(),
+                                                                       dialog.getLibNum());
+                        dialog.hide();
+                    }
+                });
+
+                dialog.addCancelButtonSelectHandler(event -> dialog.hide());
+                dialog.show(selectedFolder);
```
ash3rz: Good refactor!  I think I did the same thing in my current open PR or upcoming PR 😄

sri2k1us:
@psarando I will check with Jason Williams but if it is working even with spaces in the folder name, I can just remove the validation code and still use this as a refactor change :)

sri2k1us:
Closing this PR as this issue is not a bug really. All the injection refactors are taken care of by @ash3rz refactor PR.

PR #233, opened 2017-10-20
CORE-9048 Make app tile logo look unclickable if app is disabled

ash3rz:


ash3rz:
Thanks for the review!

PR #235, opened 2017-10-23
de-1129 fix bug with preferences save. Adjust dialog height to hiding…

sri2k1us:
… `Reset HPC Token`


sri2k1us:
Thanks for the review.

PR #236, opened 2017-10-24
CORE-9014 Enable cancel action for uncompleted batch jobs.

psarando:
This PR will enable the Analyses window's cancel menu items for batch jobs that have any submitted or running child jobs, even if the parent job is already marked as completed.


psarando:
Thanks for the reviews! 👍 
I'll wait to merge this until after cyverse-de/apps#79 is merged, since this won't have any effect until those changes are live.

PR #237, opened 2017-10-25
core-8350 Save current view (tile / grid) to session and restore previously selected view on next user session

sri2k1us:


de-lib/src/main/java/org/iplantc/de/apps/client/AppsView.java
```diff
@@ -37,7 +37,7 @@
 
         AppCategory getSelectedAppCategory();
 
-        void go(HasOneWidget container, HasId selectedAppCategory, HasId selectedApp);
+        void go(HasOneWidget container, HasId selectedAppCategory, HasId selectedApp, String actView);
```
ash3rz: Maybe we should rename this to `activeView` so it's clear?

de-lib/src/main/java/org/iplantc/de/apps/client/presenter/AppsViewPresenterImpl.java
```diff
@@ -94,13 +95,15 @@ public AppCategory getSelectedAppCategory() {
 
     @Override
     public void go(final HasOneWidget container,
-                   final HasId selectedAppCategory,
-                   final HasId selectedApp) {
+                   final HasId selectedAppCategory, final HasId selectedApp, final String actview) {
```
ash3rz: These parameters should probably all be on separate lines for consistency. Also, we should probably rename the variable to `activeView` with camelCase.

I think there are also some tests for this class so you might want to check for breakages.

de-lib/src/main/java/org/iplantc/de/apps/client/presenter/list/AppsListPresenterImpl.java
```diff
@@ -188,6 +189,36 @@ public void loadApps(List<App> apps) {
...
+    }
+
+    /**
+     * Set new active view
+     *
+     * @param newView view to set as active
+     */
+    @Override
+    public void setActiveView(String newView) {
+        if (AppsTileViewImpl.GRID_VIEW.equals(newView)) {
```
ash3rz: I think this should be `AppsListView.GRID_VIEW.equals(newView)` instead?

sri2k1us:
Updated variable names. Tests ran without issues. Thanks for the review!

PR #242, opened 2017-10-31
De 1129 Populate webhook types from database.

sri2k1us:


de-lib/src/main/java/org/iplantc/de/preferences/client/presenter/PreferencesPresenterImpl.java
```diff
@@ -62,16 +64,29 @@ public PreferencesPresenterImpl(PreferencesView view,
...
+        view.initAndShow(userSettings, typeList);
     }
 
-    void setUpView() {
-        view.initAndShow(userSettings);
+    void getWebhookTypes() {
+        serviceFacade.getWebhookTypes(new AppsCallback<WebhookTypeList>() {
+            @Override
+            public void onFailure(Integer statusCode, Throwable exception) {
+                announcer.schedule(new ErrorAnnouncementConfig(appearance.webhooktypeFailure()));
```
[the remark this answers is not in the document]
sri2k1us: good catch! 👍

de-lib/src/main/java/org/iplantc/de/client/models/webhooks/WebhookType.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+public interface WebhookType extends HasId {
+
+    String getType();
+
+    void setType(String type);
+
+    String getTemplate();
+
+    void setTemplate(String template);
```
psarando: Is the user is able to view or update a webhook type's template?
If not, we could remove the template from this model and the endpoint, and it will save some bandwidth so these strings don't have to shuffle back and forth from the clients.

sri2k1us: Ya we can remove the template field from here and make this field optional on the services. 👍

PR #244, opened 2017-11-02
Core 9062 increase height of Data window row.

sri2k1us:


sri2k1us:
Merging directly since it one line change.

PR #245, opened 2017-11-03
core-9061 Allow users to copy path to file or folder.

sri2k1us:


de-lib/src/main/java/org/iplantc/de/commons/client/views/dialogs/ClipboardCopyEnabledDialog.java
```diff
@@ -0,0 +1,73 @@
...
+
+        String errorCopying();
+    }
+
+    protected TextBox textBox;
+    protected TextButton copyBtn;
+
+    Appearance appearance = GWT.create(Appearance.class);
+
+    public ClipboardCopyEnabledDialog() {
```
ash3rz: Can this class be injected?

sri2k1us: Actually this class is meant to be extended like `IPlantDialog`. So the class thats extending this one can do the injection.

de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/grid/GridViewPresenterImpl.java
```diff
@@ -358,6 +360,24 @@ public void onSuccess(CommentsDialog result) {
     }
 
     @Override
+    public void onCopyPathSelected(CopyPathSelected event) {
+        final DiskResource dr = event.getDiskResource();
+        shareLinkDialogProvider.get(new AsyncCallback<ShareResourceLinkDialog>() {
+            @Override
+            public void onFailure(Throwable caught) {
+                announcer.schedule(new ErrorAnnouncementConfig(appearance.shareByLinkFailure()));
```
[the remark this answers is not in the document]
sri2k1us: +1

de-lib/src/main/java/org/iplantc/de/diskResource/client/views/sharing/dialogs/ShareResourceLinkDialog.java
```diff
@@ -1,42 +1,35 @@
...
+
+        HorizontalPanel panel = new HorizontalPanel();
     }
 
     public void show(final String link) {
         textBox.setValue(link);
         textBox.selectAll();
+        ensureDebugId(DiskResourceModule.Ids.SHARE_LINK_DLG);
+        textBox.getElement()
+               .setId(DiskResourceModule.Ids.SHARE_LINK_DLG + DiskResourceModule.Ids.LINK_TEXT);
```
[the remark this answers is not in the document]
sri2k1us: `ensureDebugId` won't set `id` on the actual `input` element. It sets it on the enclosing `div`. That's why i had the explicitly set here.

de-lib/src/main/java/org/iplantc/de/diskResource/share/DiskResourceModule.java
```diff
@@ -108,6 +110,8 @@
         String GENOME_IMPORT_BTN = ".importBtn";
         String RENAME_RESOURCE_DLG = "renameResourceDlg";
         String TEXT_FIELD = ".txtField";
+        String LINK_TEXT = ".linkText";
+        String SHARE_LINK_DLG = ".shareLinkDlg";
```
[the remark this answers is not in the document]
sri2k1us: 👍

de-lib/src/main/java/org/iplantc/de/diskResource/client/views/sharing/dialogs/ShareResourceLinkDialog.java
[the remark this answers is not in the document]
ash3rz: Ah, I didn't catch the difference. This could probably still be moved into the `onEnsureDebugId` method though so they're all together.

de-lib/src/main/java/org/iplantc/de/commons/client/views/dialogs/ClipboardCopyEnabledDialog.java
```diff
@@ -0,0 +1,73 @@
...
+
+        textBox = new TextBox();
+        textBox.setReadOnly(true);
+        textBox.setWidth(appearance.textboxWidth());
+        textBox.setHeight(appearance.textboxHeight());
+
+        panel.add(textBox);
+
+        if (CopyToClipboardUtil.isSupported()) {
+            copyBtn = new TextButton(appearance.copyToClipboard());
```
psarando: 👍

de-lib/src/main/java/org/iplantc/de/client/util/CopyToClipboardUtil.java
```diff
@@ -0,0 +1,58 @@
...
+    }-*/;
+
+    /**
+     * Check if copy is supported by browser
+     *
+     * @return boolean
+     */
+    public static native boolean isSupported() /*-{
+        try {
+            return $doc.queryCommandSupported('copy');
```
psarando: We might also want to check if the browser has the `queryCommandSupported` function in the first place (it could prevent extraneous exceptions and console logging).

For example, from my React POC: https://github.com/psarando/cyverse-de-ui/blob/a2635553befc7a29ef0273c12e3d29137c2757ac/react-components/src/clipboardFunctions.js#L9-L10

de-lib/src/main/java/org/iplantc/de/client/util/CopyToClipboardUtil.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * Copy text value from the textfield to clipboard
+     *
+     * @param id id of the textfield
+     * @return boolean whether copy succeeded or failed.
+     */
+    public static native boolean copyToClipboard(String id) /*-{
+        var copyText = $doc.getElementById(id);
+        copyText.select();
```
psarando: We may as well also move this inside the try/catch block, in case a bad ID is passed in?

de-lib/src/main/java/org/iplantc/de/client/util/CopyToClipboardUtil.java
```diff
@@ -0,0 +1,58 @@
...
+        try {
+            if ($doc.queryCommandEnabled('copy')) {
+                return $doc.execCommand('copy');
+            } else {
+                return false;
```
psarando: If the very last line of this function is `return false;` then we don't need this `else` or the `return` from inside the catch block.

PR #250, opened 2017-11-14
Potentially fix Intercom widget position

ash3rz:
I was looking at this documentation:
https://docs.intercom.com/configure-intercom-for-your-product-or-site/customize-the-intercom-messenger/customize-the-intercom-messenger-technical

Seems to be working
image
compared to de-2 right now
image


ash3rz:
Thanks for the review!

PR #252, opened 2017-11-16
CORE-9071 Fix folder selection saving even when choosing "Cancel"

ash3rz:


ash3rz:
Thanks for the review!

PR #253, opened 2017-12-08
CORE-9087 Fix bug where closing previously minimized windows does not…

ash3rz:
… clear their state

Sencha does not actually have a minimize method that does anything and also does not have any checks that a window is already registered with the singleton WindowManager class before registering it again in the show() method.  In the DE, when a minimize button was clicked, the hide() method was actually called which tries to unregister the window, but then IplantWindowBase was configured to re-register the window so WindowManager could still keep track of it.  When the window was restored, the show() method was called which then re-registered the window, creating a duplicate in WindowManager.  Any time the user then tried to close the window, the duplicate would still exist.  Trying to open a new window of the same type would then result in the DE's DesktopWindowManager class to fetch the duplicate window which contained the old state.

NOTE: This is an existing bug which doesn't need to go into the 12/12 release.


ash3rz:
Thanks for the reviews!

PR #254, opened 2017-12-12
CORE-9088 Fix bug where you can't reopen the Data window if you have …

ash3rz:
…a file open

In DesktopWindowManager we're checking to see if any of the current open windows has a state ID that starts with the data window's type which is DATA.  Since DATA_VIEWER starts with DATA, DesktopWindowManager would just bring the file viewer window to the front.


ash3rz:
Thanks for the review!

PR #255, opened 2017-12-12
CORE-9090 Always show the search boxes in Data/Analysis/App windows

ash3rz:
Very small change and I haven't seen any side effects when I was testing.


ash3rz:
Thanks for the review!

PR #256, opened 2018-01-04
core-8350 Proof-of-Concept: Use Web local storage to store app and data window preferences

sri2k1us:
Previously I attempted to fix this issue by storing these preferences in de user session. But It wont work because we use session only to restore previous workspace configuration (re-open windows that were open during last login) and not when opening a windows subsequently.  So I wanted to give Web local storage a try.

There is a downside to using Web local storage. These preferences won't persist across computers / workstations / laptops. But I guess most of our users use their own laptop / workstation ?!? If this approach is acceptable, then we can expand this solution to store other window level preferences like preferred window height and width for each window.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/desktop/client/views/windows/AboutApplicationWindow.java
+++ b/de-lib/src/main/java/org/iplantc/de/desktop/client/views/windows/AboutApplicationWindow.java
@@ -43,6 +43,10 @@ public class AboutApplicationWindow extends IplantWindowBase {
         ImageResource iplantAbout();
 
         String iplantcAboutPadText();
+
+        String windowWidth();
+
+        String windowHeight();
     }
     private final AboutApplicationServiceAsync aboutApplicationService;
     private final AboutApplicationAppearance appearance;
@@ -57,7 +61,8 @@ public class AboutApplicationWindow extends IplantWindowBase {
 
         String width = getSavedWidth(WindowType.ABOUT.toString());
         String height = getSavedHeight(WindowType.ABOUT.toString());
-        setSize((width == null) ? "320" : width, (height == null) ? "260" : height);
+        setSize((width == null) ? appearance.windowWidth() : width,
+                (height == null) ? appearance.windowHeight() : height);
 
         setHeading(appearance.headingText());
         ensureDebugId(DeModule.WindowIds.ABOUT_WINDOW);
--- a/de-lib/src/main/java/org/iplantc/de/desktop/client/views/windows/AboutApplicationWindow.java
+++ b/de-lib/src/main/java/org/iplantc/de/desktop/client/views/windows/AboutApplicationWindow.java
@@ -2,7 +2,9 @@ package org.iplantc.de.desktop.client.views.windows;
 
 import org.iplantc.de.client.models.AboutApplicationData;
 import org.iplantc.de.client.models.CommonModelAutoBeanFactory;
+import org.iplantc.de.client.models.UserInfo;
 import org.iplantc.de.client.models.WindowState;
+import org.iplantc.de.client.models.WindowType;
 import org.iplantc.de.commons.client.ErrorHandler;
 import org.iplantc.de.commons.client.views.window.configs.ConfigFactory;
 import org.iplantc.de.desktop.shared.DeModule;
@@ -48,10 +50,15 @@ public class AboutApplicationWindow extends IplantWindowBase {
 
     @Inject
     AboutApplicationWindow(final AboutApplicationServiceAsync aboutApplicationService,
-                           final AboutApplicationAppearance appearance) {
+                           final AboutApplicationAppearance appearance, final UserInfo userInfo) {
         this.aboutApplicationService = aboutApplicationService;
         this.appearance = appearance;
-        setSize("320", "260");
+        this.userInfo = userInfo;
+
+        String width = getSavedWidth(WindowType.ABOUT.toString());
+        String height = getSavedHeight(WindowType.ABOUT.toString());
+        setSize((width == null) ? "320" : width, (height == null) ? "260" : height);
+
         setHeading(appearance.headingText());
         ensureDebugId(DeModule.WindowIds.ABOUT_WINDOW);
         executeServiceCall();
@@ -103,5 +110,12 @@ public class AboutApplicationWindow extends IplantWindowBase {
                 compose();
             }
         });
+    }
+
+    @Override
+    public void hide() {
+        saveHeight(WindowType.ABOUT.toString());
+        saveWidth(WindowType.ABOUT.toString());
+        super.hide();
     }
 }
--- a/de-lib/src/main/java/org/iplantc/de/desktop/client/views/windows/DEAppsWindow.java
+++ b/de-lib/src/main/java/org/iplantc/de/desktop/client/views/windows/DEAppsWindow.java
@@ -4,6 +4,7 @@ import org.iplantc.de.apps.client.AppsView;
 import org.iplantc.de.apps.shared.AppsModule;
 import org.iplantc.de.client.models.UserInfo;
 import org.iplantc.de.client.models.WindowState;
+import org.iplantc.de.client.models.WindowType;
 import org.iplantc.de.client.util.WebStorageUtil;
 import org.iplantc.de.commons.client.util.WindowUtil;
 import org.iplantc.de.commons.client.views.window.configs.AppsWindowConfig;
@@ -25,19 +26,21 @@ public class DEAppsWindow extends IplantWindowBase {
 
     public static final String APPS = "#apps";
     public static final String DE_APPS_ACTIVEVIEW = "de.apps.activeview#";
+
     private final AppsView.Presenter presenter;
 
     @Inject
-    UserInfo userInfo;
-
-
-    @Inject
-    DEAppsWindow(final AppsView.Presenter presenter, final IplantDisplayStrings displayStrings) {
+    DEAppsWindow(final AppsView.Presenter presenter,
+                 final IplantDisplayStrings displayStrings,
+                 final UserInfo userInfo) {
         this.presenter = presenter;
+        this.userInfo = userInfo;
 
         // This must be set before we render view
         ensureDebugId(DeModule.WindowIds.APPS_WINDOW);
-        setSize("820", "400");
+        String width = getSavedWidth(WindowType.APPS.toString());
+        String height = getSavedHeight(WindowType.APPS.toString());
+        setSize((width == null) ? "820" : width, (height == null) ? "400" : height);
         setMinWidth(540);
         setHeading(displayStrings.applications());
     }
@@ -71,10 +74,12 @@ public class DEAppsWindow extends IplantWindowBase {
     }
 
     @Override
-    public void doHide() {
+    public void hide() {
         WebStorageUtil.writeToStorage(DE_APPS_ACTIVEVIEW + userInfo.getUsername(),
                                       presenter.getActiveView());
-        super.doHide();
+        saveHeight(WindowType.APPS.toString());
+        saveWidth(WindowType.APPS.toString());
+        super.hide();
     }
 
     @Override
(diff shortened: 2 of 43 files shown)


de-lib/src/main/java/org/iplantc/de/desktop/client/views/windows/DEAppsWindow.java
```diff
@@ -43,7 +46,8 @@
         presenter.go(this,
                      appsWindowConfig.getSelectedAppCategory(),
                      appsWindowConfig.getSelectedApp(),
-                     appsWindowConfig.getView());
+                     WebStorageUtil.readFromStorage(
+                             DE_APPS_ACTIVEVIEW + UserInfo.getInstance().getUsername()));
```
ash3rz: We should probably inject UserInfo

de-lib/src/main/java/org/iplantc/de/desktop/client/views/windows/DeDiskResourceWindow.java
```diff
@@ -69,7 +72,10 @@
                                                                 resourcesToSelect);
         final String uniqueWindowTag = (diskResourceWindowConfig.getTag() == null) ? "" : "." + diskResourceWindowConfig.getTag();
         ensureDebugId(DeModule.WindowIds.DISK_RESOURCE_WINDOW + uniqueWindowTag);
-        presenter.go(this);
+        String minimizeDetails  = WebStorageUtil.readFromStorage(
+                DE_DATA_DETAILSPANEL_COLLAPSE + UserInfo
```
ash3rz: Same here - probably should inject UserInfo

psarando:
Why not store these preferences with the `user-preferences` service?
Then these preferences could persist across computers/devices.

ash3rz:
That's a good question.  I remember we dismissed the idea of adding this to user preferences before (like maybe 2 years ago now) because I think we wanted window preferences to be separate, though I can't remember the particular reason why we wanted them separate given at the time. I've been doing a bit of a bug scrub and I see we have other window preferences requested: https://pods.iplantcollaborative.org/jira/browse/CORE-3713

With that in mind, maybe it would also make sense to have these be user preferences that can be turned on or off?

ash3rz:
Actually I take that back - I think what we talked about was having window preferences be visually separate, not separate services.  So for each window you'd have a menu where you could choose to enable saving the configuration or whatever else for just that window type.

sri2k1us:
One reason i think we should not store this on the server side is that users working on different laptop / workstations will end-up with different screen sizes. So the window sizes and column which works well in one device may not work well in a different device.  We actually encounter this problem when restore session.

sri2k1us:
@ash3rz - I think we can just store users preference for window sizes and column widths without having them to set it up in a settings UI. And the next time they open a window, we could restore the window size and column widths they previously used for that window.

ash3rz:
As an aside, do we still have that issue when restoring sessions?  I thought that was fixed a while ago.

I could see users wanting different settings for different devices though, I agree.

sri2k1us:
@ash3rz Yes it is fixed. Sorry I was not clear about that!

sri2k1us:
Updated PR with code that stores and restores window sizes from local storage.

de-lib/src/main/java/org/iplantc/de/desktop/client/views/windows/AboutApplicationWindow.java
```diff
@@ -48,10 +50,15 @@
...
-                           final AboutApplicationAppearance appearance) {
+                           final AboutApplicationAppearance appearance, final UserInfo userInfo) {
         this.aboutApplicationService = aboutApplicationService;
         this.appearance = appearance;
-        setSize("320", "260");
+        this.userInfo = userInfo;
+
+        String width = getSavedWidth(WindowType.ABOUT.toString());
+        String height = getSavedHeight(WindowType.ABOUT.toString());
+        setSize((width == null) ? "320" : width, (height == null) ? "260" : height);
```
psarando: It might be safer to check the saved width/height with `Strings.isNullOrEmpty` (in case an empty string gets saved for some reason).
This also might be a good opportunity to move the default values into an appearance.

de-lib/src/main/java/org/iplantc/de/desktop/client/views/windows/SystemMessagesWindow.java
```diff
@@ -20,9 +22,16 @@
...
+
     @Inject
-    SystemMessagesWindow(final IplantDisplayStrings displayStrings) {
+    SystemMessagesWindow(final IplantDisplayStrings displayStrings,
+                         final UserInfo userInfo) {
+        this.userInfo = userInfo;
         setHeading(displayStrings.systemMessagesLabel());
+        String width = getSavedWidth(WindowType.SYSTEM_MESSAGES.toString());
+        String height = getSavedHeight(WindowType.SYSTEM_MESSAGES.toString());
+        setSize((width == null) ? computeDefaultWidth() + "" : width, (height == null) ? computeDefaultHeight() + "" : height);
```
psarando: It looks like the `setWidth` and `setHeight` calls below will override this `setSize` call?

de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/DiskResourcePresenterImpl.java
```diff
@@ -644,14 +645,19 @@ public void go(HasOneWidget container,
                    final List<? extends HasId> diskResourcesToSelect) {
 
         if ((folderToSelect == null) || Strings.isNullOrEmpty(folderToSelect.getPath())) {
-            go(container);
+            go(container, true);
```
psarando: Does this mean new data windows will open with the details panel collapsed by default?

It may not be obvious to new users that they can view more details of a selected item by clicking the `<<` button at the top-right of the grid (under the `Trash` menu and next to the folder name grid-header).

de-lib/src/main/java/org/iplantc/de/desktop/client/views/windows/SystemMessagesWindow.java
sri2k1us: Yup those needs to removed! Sorry Oversight on my part!

de-lib/src/main/java/org/iplantc/de/desktop/client/views/windows/AboutApplicationWindow.java
sri2k1us: 👍

de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/DiskResourcePresenterImpl.java
sri2k1us: This `go()` method is only used from folder selector. So I set them to true by default. But in general Data window opens with details panel expanded. Once you collapse the details panel, we store that as preference in the local storage.

de-lib/src/main/java/org/iplantc/de/desktop/client/views/windows/DeDiskResourceWindow.java
```diff
@@ -35,18 +38,23 @@
...
         this.presenterFactory = presenterFactory;
         this.displayStrings = displayStrings;
+        this.userInfo = userInfo;
         setHeading(displayStrings.data());
-        setSize("900", "480");
+        String width = getSavedWidth(WindowType.DATA.toString());
+        String height = getSavedHeight(WindowType.DATA.toString());
+        setSize((width == null) ? "820" : width, (height == null) ? "400" : height);
         setMinWidth(900);
         setMinHeight(480);
```
ash3rz: Not required for this PR, but this is probably a good time to move all of our window sizes to appearance classes.  

Also, for this change in particular I'm not sure what will happen if we set the size to one thing, but then set the minimum sizes to something greater than what was initially set?

psarando: I think I've forced this situation in testing before, by opening a window in a really small browser. The window opens smaller than the min settings, but as soon as you try to resize it, it snaps to the minimum sizes and forces the user to only increase the size from there.

de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/DiskResourcePresenterImpl.java
psarando: Ah, I didn't catch that this `go` method is only called from the file/folder select dialogs before.
And now I see where the collapsed settings is saved in the window (and not in the presenter) 👍

de-lib/src/main/java/org/iplantc/de/theme/base/client/tools/ManageToolsViewDefaultAppearance.java
```diff
@@ -183,6 +183,16 @@ public String toolInfoError() {
...
     }
 
+    @Override
+    public String windowWidth() {
+        return "600";
+    }
+
+    @Override
+    public String windowHeight() {
+        return "500";
```
psarando: The original dimensions of this window were 800x600.
Are we intentionally reducing it to 600x500 now?

ianmcorvidae:
From an architectural sort of level, I still sort of wonder what goes here, what goes in the user-sessions storage, and what goes in user-preferences. Prior to this, I think the summary is:

* preferences are stuff users can directly change in the preferences window
* sessions are saved periodically or on request and store the open windows and their states

I guess both the sticky view and the window sizes are sort of a hybrid, where we want it to act like a preference (persists even if you close windows), but not directly something you can change explicitly in the preferences window, and also something that's currently saved in the session but we want to be saved even when they're closed (unless window sizes aren't saved, I forget. But, it'd make sense if they were so I think it's the same idea. similarly column widths). So I guess my questions are:

* What else looks like this? Some of our windows have tabs which could be stickier, and possibly other "view change" things like filters could be made sticky. But I'm not sure if we'd want to, of course.
* What takes priority, this or the saved session? That is, when opening from a session, are we still saving these things (or are they being removed from the saved session handling), and if we are does the local storage or the saved session take priority? Is it the same for all of this class of things, or does it differ? (i.e., might it be the case that we always use the local storage for one thing, but prefer the saved session for something else)
* Do we ever need to worry about the above interaction causing weird situations -- for example, someone might make their window bigger in order to see more tiles (vs. the list view which is more compact) or their columns wider to see more of the list view (vs. the more horizontally compact tiles), so is it possible for us to get into a state where upon reopening it'd create a too-small tile-view window or a too-large list view window or a too-narrow list view window or a too-wide tile view window? If it is, do we care?
* Have I missed any of the intended use cases for this new type of thing?

(hopefully this comment isn't too long, I know I have a habit of asking too many questions too soon, heh)

de-lib/src/main/java/org/iplantc/de/theme/base/client/tools/ManageToolsViewDefaultAppearance.java
sri2k1us: Yes I did change it intentionally. Felt it was too big but I can always change it back if others feels differently.

sri2k1us:
- We save just what windows were open and their current selection in user-sessions. We don't save their size.

- At the moment, we are planning to store window sizes and column widths for all the windows. For apps window, we store if users preference of tile / grid view. For data window, we store if the details panel was collapsed or expanded. The use-case for this is that more often than not users have to adjust the window sizes and column widths before they could begin working when switching from a laptop to workstation or vice-versa. 

- I think things like current tab, filters should be saved in session so that they persist across devices. 

- Things saved in user session is not saved in local storage. So I guess we don't need to decide priority .

- We store the window sizes, preferred view and column widths in local storage when the window is closed. In other words we store what user was seeing in terms of window dimensions when the window was closed. So I think we will not run in weird edge case where window is too small or to big. Even if we run into some weird condition, we have a set minimum width / height set for most window so that should fix these edge cases.  

Hopefully this answers @ianmcorvidae questions.

sri2k1us:
Oops I take it back. We do save window size and position in session. I guess we could start saving the window size in local storage only.

de-lib/src/main/java/org/iplantc/de/theme/base/client/tools/ManageToolsViewDefaultAppearance.java
psarando: I don't have an issue with the new size. I was just making sure it wasn't an untested accident ✅

sri2k1us:
I am going to close this pull request and open a new one since I have taken a different approach for this issue.

PR #257, opened 2018-01-05
core-7530 display image name/tag instead of path in App details.

sri2k1us:


de-lib/src/main/java/org/iplantc/de/apps/client/AppDetailsView.java
```diff
@@ -40,6 +40,8 @@
                                         DetailsCategoryClicked.HasDetailsCategoryClickedHandlers {
 
     interface AppDetailsAppearance {
+        String imageLabel();
```
ash3rz: Really really minor, but this is the only variable above the `AppDetailsStyle` interface so it looks a bit out of place.

sri2k1us:
Thanks for the review.

PR #258, opened 2018-01-05
CORE-5143 Close FileViewer after prompt to save

ash3rz:
This is just when the user attempts to close the file with unsaved changes.  If the save is successful, the viewer will close.


sri2k1us:
LGTM :shipit:

ash3rz:
Thanks for the review!

PR #259, opened 2018-01-06
CORE-4809 Add horizontal scroll to AppLaunch and switch to GWT Checkbox

ash3rz:
Sencha's Checkbox appearance contains "white-space: nowrap;" with no clear way to change that.  This causes the Retain Inputs text to stretch beyond the window's width.

I didn't immediately see any other elements that didn't adjust with the width of the window, but just in case, I added horizontal scroll support.


ash3rz:
Thanks for the review!

PR #261, opened 2018-01-20
CORE-8572: ignore empty strings when adding granted authorities to a …

slr71:
…user


slr71:
Thanks for the review!

PR #266, opened 2018-01-26
CORE-9107 Fix Preferences sizing issue

ash3rz:
QA Preferences tests are failing because the Default Output Folder items aren't able to fit in the field set they're in.  I updated that section from having a fixed height to using a CssFloatLayoutContainer so the field set will expand to fit the size of the children.  Also added a scroll bar in case all of the preferences extend beyond the size of the Preferences dialog.


ash3rz:
The changes are actually only a few lines, but it looks like more because the spacing was off in the xml file.  I'll go ahead and merge but reviews still welcome.

PR #272, opened 2018-02-01
CORE-9152 Add missing key to SelectionItem before adding to hierarchi…

ash3rz:
…cal lists TreeStore

When adding an item to the TreeStore, the store uses SelectionItemModelKeyProvider to get the item's key.  This key provider assumes, at some point prior, a SelectionItem.TMP_ID_TAG key has already been set. This fix was copied from SelectionItemPropertyEditor's onAddButtonClicked method.


ash3rz:
This is a pretty small change, so I'll go ahead and merge.  Reviews still welcome though.

PR #273, opened 2018-02-01
Add AppEditor/AppLaunch static IDs

ash3rz:


ash3rz:
Since this is just adding static IDs, I think I'll go ahead and merge.  Reviews still welcome though.

PR #275, opened 2018-02-08
core-3713 Save column preferences for Data window in local storage.

sri2k1us:


de-lib/src/main/java/org/iplantc/de/diskResource/client/DiskResourceView.java
```diff
@@ -46,6 +48,10 @@
 
     void setWestPanelWidth(String width);
 
+    ColumnModel<DiskResource> getColumns();
+
+    void setColumnPreferences(Map<String, String> preferences);
+
```
tedgin: Please create Javadoc comments describing the purpose of each new public method.

de-lib/src/main/java/org/iplantc/de/diskResource/client/GridView.java
```diff
@@ -267,4 +268,7 @@
     LiveGridCheckBoxSelectionModel getSelectionModel();
 
     void setSingleSelect();
+
+    void setColumnPreferences(Map<String, String> preferences);
```
tedgin: Please add javadoc comments describing purpose of new public method.

de-lib/src/main/java/org/iplantc/de/diskResource/client/views/grid/DiskResourceColumnModel.java
```diff
@@ -49,6 +49,8 @@
                                                                                   CopyPathSelected.HasCopyPathSelectedEventHandlers {
 
 
+    private static List<ColumnConfig<DiskResource, ?>> list;
```
tedgin: Why did you a local variable of a static method get promoted to a private static field?  It's not being used anywhere else. Also, the original static method overwrites it each time, so it's not even acting like a cached copy.

sri2k1us: Good catch. I started with this approach and realized later i don't need get `ColumnConfig` from here. This change is not required.

de-lib/src/main/java/org/iplantc/de/diskResource/client/GridView.java
sri2k1us: 👍

de-lib/src/main/java/org/iplantc/de/diskResource/client/DiskResourceView.java
sri2k1us: 👍

tedgin:
It now looks good to me.

PR #276, opened 2018-02-14
CORE-9157 Enable UI to try to reconnect to AMQP after a failure to connect

ash3rz:
** Merge only after the 2.19 Release **

If the UI can't connect to AMQP, the `Connection` returned from the `ConnectionFactory` in `AMQPConnectionManager.java` is null. The changes are mostly just making sure we keep doing null checks and trying to get new connections from `ConnectionFactory`.  If despite that, we still end up with a null `Channel` in `MessageHandler.java`, the websocket gets closed.  
From there, [this logic](https://github.com/cyverse-de/ui/blob/master/de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/DesktopPresenterImpl.java#L221-L224) tries to reopen the websocket and then restarts the cycle in `AMQPConnectionManager`.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/DesktopPresenterImpl.java
+++ b/de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/DesktopPresenterImpl.java
@@ -211,12 +211,12 @@ public class DesktopPresenterImpl implements DesktopView.Presenter {
     }
 
     private void initNotificationWebSocket() {
-        notificationWebSocketManager = NotificationWebSocketManager.getInstace();
+        notificationWebSocketManager = NotificationWebSocketManager.getInstance();
         notificationWebSocketManager.openWebSocket(new WebsocketListener() {
 
             @Override
             public void onClose() {
-                GWT.log("Wesbsocket onClose()");
+                GWT.log("WebSocket onClose()");
                 //if websocket connection closed unexpectedly, retry connection!
                 if(!loggedOut) {
                     GWT.log("reconnecting...");
@@ -238,12 +238,12 @@ public class DesktopPresenterImpl implements DesktopView.Presenter {
     }
 
     private void initSystemMessageWebSocket() {
-        systemMessageWebSocketManager = SystemMessageWebSocketManager.getInstace();
+        systemMessageWebSocketManager = SystemMessageWebSocketManager.getInstance();
         systemMessageWebSocketManager.openWebSocket(new WebsocketListener() {
 
             @Override
             public void onClose() {
-                GWT.log("Wesbsocket onClose()");
+                GWT.log("WebSocket onClose()");
                 //if websocket connection closed unexpectedly, retry connection!
                 if(!loggedOut) {
                     GWT.log("reconnecting...");
--- a/de-webapp/src/main/java/org/iplantc/de/server/websocket/MessageHandler.java
+++ b/de-webapp/src/main/java/org/iplantc/de/server/websocket/MessageHandler.java
@@ -39,36 +39,37 @@ public abstract class MessageHandler extends WebSocketHandlerAdapter {
 
 
     @Override
-    public void onOpen(WebSocket webSocket) throws IOException {
-        notificationReceiver = new ReceiveNotificationsDirect();
-        logger.debug("Web socket connection opened!");
-        String username = getUserName(webSocket);
-        logger.debug("user name:" + username);
+    public void onOpen(WebSocket webSocket) {
+        try {
+            notificationReceiver = new ReceiveNotificationsDirect();
+            logger.debug("Web socket connection opened!");
+            String username = getUserName(webSocket);
+            logger.debug("user name:" + username);
 
-        final Channel msgChannel = notificationReceiver.createChannel();
-        if (msgChannel == null) {
+            final Channel msgChannel = notificationReceiver.createChannel();
+            String queue = bindQueue(username, msgChannel);
+            consumeMessage(msgChannel, queue, webSocket);
+
+            webSocket.resource().addEventListener(new WebSocketEventListenerAdapter() {
+                @Override
+                public void onDisconnect(AtmosphereResourceEvent event) {
+                    try {
+                        if (msgChannel != null) {
+                            msgChannel.close();
+                        }
+                    } catch (IOException e) {
+                        logger.error("Exception aborting channel", e);
+
+                    } catch (TimeoutException e) {
+                        e.printStackTrace();
+                    }
+                }
+            });
+        }
+        catch (IOException exception) {
             onError(webSocket, new WebSocketProcessor.WebSocketException("AMQP channel was null", null));
             webSocket.close();
-            return;
         }
-        String queue = bindQueue(username, msgChannel);
-        consumeMessage(msgChannel, queue, webSocket);
-
-        webSocket.resource().addEventListener(new WebSocketEventListenerAdapter() {
-            @Override
-            public void onDisconnect(AtmosphereResourceEvent event) {
-                try {
-                    if (msgChannel != null) {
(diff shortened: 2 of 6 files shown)


de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/DesktopPresenterImpl.java
```diff
@@ -211,12 +211,12 @@ public DesktopPresenterImpl(final DesktopView view,
     }
 
     private void initNotificationWebSocket() {
-        notificationWebSocketManager = NotificationWebSocketManager.getInstace();
+        notificationWebSocketManager = NotificationWebSocketManager.getInstance();
```
slr71: 👍

de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/DesktopPresenterImpl.java
```diff
@@ -211,12 +211,12 @@ public DesktopPresenterImpl(final DesktopView view,
...
         notificationWebSocketManager.openWebSocket(new WebsocketListener() {
 
             @Override
             public void onClose() {
-                GWT.log("Wesbsocket onClose()");
+                GWT.log("WebSocket onClose()");
```
slr71: 👍

de-webapp/src/main/java/org/iplantc/de/server/websocket/ReceiveNotificationsDirect.java
```diff
@@ -28,24 +28,31 @@
...
-       LOG.info("amqp Connection created created!");
+       connection = getConnection();
+       LOG.info("amqp Connection created!");
+    }
+
+    private Connection getConnection() {
+        if (connection == null) {
+            connection = AMQPConnectionManager.getInstance().getConnection();
+        }
+        return connection;
```
slr71: Does `getConnection` ever return a null pointer then return an actual connection in a later call? I suppose that a null pointer only means that the connection couldn't be established, and not necessarily that the the web socket itself is bad. I wonder if it might be useful to have `ReceiveNotificationsDirect` throw an `IOException` instead and have `MessageHandler` or some higher level object catch the exception and try to reestablish the connection later. Doing that would remove the need to make the `connection` property mutable, but I'm not entirely certain that it wouldn't create some other problems.

de-webapp/src/main/java/org/iplantc/de/server/websocket/MessageHandler.java
```diff
@@ -46,6 +46,11 @@ public void onOpen(WebSocket webSocket) throws IOException {
         logger.debug("user name:" + username);
 
         final Channel msgChannel = notificationReceiver.createChannel();
+        if (msgChannel == null) {
+            onError(webSocket, new WebSocketProcessor.WebSocketException("AMQP channel was null", null));
+            webSocket.close();
+            return;
+        }
```
slr71: This may not be necessary, but should `webSocket.close()` also be called if `notificationReceiver.createChannel()` does not return `null`?

de-webapp/src/main/java/org/iplantc/de/server/websocket/amqp/AMQPConnectionManager.java
```diff
@@ -56,6 +61,9 @@ public static AMQPConnectionManager getInstance() {
     }
 
     public Connection getConnection() {
+        if (connection == null) {
+            connection = createConnection();
+        }
```
slr71: I wonder if another call should be made to verify that the connection is good (in addition to verifying that it's not null) might be necessary. For example, the connection might have been lost because of network shenanigans, but the `connection` property still might not be null.

de-webapp/src/main/java/org/iplantc/de/server/websocket/amqp/AMQPConnectionManager.java
```diff
@@ -46,6 +50,7 @@ private AMQPConnectionManager() {
             LOG.error("Exception when creating AMQP connection!" + e.getMessage());
         }
 
+        return connection;
     }
```
slr71: The assignment to the `connection` property happens both inside `createConnection` and in all of the places where `createConnection` is called. It might be useful to remove the assignments either from the places where `createConnection` is called or from `createConnection` itself.

ash3rz: Ah, good catch!

de-webapp/src/main/java/org/iplantc/de/server/websocket/amqp/AMQPConnectionManager.java
ash3rz: That's a good question.  As far as I can recall, when the UI has stopped receiving notifications, the connection has been null so I wasn't too concerned about other cases, but I can look into it.

de-webapp/src/main/java/org/iplantc/de/server/websocket/MessageHandler.java
ash3rz: If it's possible for there to be another case where there's an issue with the channel but the channel is not null, then perhaps.  Outside of that though, I don't think the websocket needs to be closed.

de-webapp/src/main/java/org/iplantc/de/server/websocket/ReceiveNotificationsDirect.java
ash3rz: I was testing this by updating the firewall rules on the AMQP box.  So with that in mind, it was possible for getConnection to start returning an actual connection after having returned a null pointer once I updated the firewall to stop blocking me.
I'm definitely not the most familiar with this part of the code, so if there's a better way to handle this I'm definitely open 😄

slr71: Ah, that makes sense. I also learned a little bit about how the code works since writing this comment, so it makes more sense to me now. 😄  Does the client-side code attempt to reconnect if the web socket is closed? If it does then one approach might be to close the web socket if the AMQP connection can't be established.

ash3rz: Yes, there's a check [here](https://github.com/cyverse-de/ui/blob/master/de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/DesktopPresenterImpl.java#L221-L224) when the websocket gets closed.  If the user wasn't trying to log out, then we try to reopen the websocket and then all this connection logic gets re-run again.

It would probably be cleaner to close the websocket right away in `AMQPConnectionManager` like you mentioned, but then I'm not sure how to get access to the websocket from that stage in order to close it?  Or maybe there's a different way to close it I'm not aware of?

slr71: Ah, good news. One way to do it would be to throw an exception in `ReceiveNotificationsDirect`'s constructor if the connection can't be established. Then `MessageHandler.onOpen` can catch the exception and close the web socket. Another option would be to have `ReceiveNotificationsDirect.createChannel` simply return null if the `connection` property is null. If you did that then the code that you've added to `MessageHandler.onOpen` would detect the problem and close the web socket.

I think I like the idea of having the constructor throw an `IOException` if the connection can't be established. In that case, the error is being handled closer to its source. Other options would definitely work, though.

ash3rz: OK perfect, I'll give that a go tomorrow - thanks for the help!

ash3rz:
Thanks for the review!

PR #283, opened 2018-03-30
CORE-5929 Add static IDs to file viewers

ash3rz:


ash3rz:
Thanks for the review!

PR #284, opened 2018-04-02
CORE-9153 Remove system messages from DE and Belphegor

ash3rz:
I removed everything that I could find related to system messages.  I tested Belphy and the DE locally, didn't see any error messages, and regular notifications are still working.


ash3rz:
Since this is mostly just a ton of deletes, I'm going to go ahead and merge this.  Reviews still welcome though.

PR #285, opened 2018-04-04
CORE-9156 Make App Editor window error more specific regarding requir…

ash3rz:
…ed fields


ash3rz:
I'll go ahead and merge this since it's a small change.  Reviews still welcome.

yeea:
Do we want to distinguish if just one or both fields are empty? Not a huge deal in any case, so LGTM

ash3rz:
Good question. I was thinking since it's only two fields that might be blank and they're next to each other in the App Editor panel that suggesting checking both fields would suffice.  If having the error message be more specific would be clearer though, it's an easy enough change.

yeea:
Nah, it's clear enough, IMO. Just thought I'd throw the nitpick out there just in case :)

PR #289, opened 2018-04-19
React data details and App stats

sri2k1us:


de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/details/DetailsViewPresenterImpl.java
```diff
@@ -26,49 +46,152 @@
...
+    }
+
+    @Override
+    public void fetchTagsForResource(String diskResourceId, TagsFetchCallback callback) {
+        tags.clear();
+        metadataService.getTags(diskResourceId, new AsyncCallback<List<Tag>>() {
+            @Override
+            public void onFailure(Throwable caught) {
+                // FIXME Move to appearance
+                ErrorHandler.post("Unable to retrieve tags!", caught);
```
ash3rz: This would probably be a good time to move these strings to an appearance like the comment says 😄

de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/callbacks/TagsSearchCallback.java
```diff
@@ -0,0 +1,14 @@
...
+
+import jsinterop.annotations.JsFunction;
+
+/**
+ * Created by sriram on 3/9/18.
+ */
+@JsFunction
+@SuppressWarnings("unusable-by-js")
+public interface TagsSearchCallback {
+    void searchComplete(Splittable[] tags);
```
ash3rz: I'm probably missing something, but is there a reason for `Splittable[]` over `Splittable`?

de-lib/src/main/java/org/iplantc/de/diskResource/client/views/details/DetailsViewImpl.java
```diff
@@ -430,7 +227,7 @@ public void onFetchDetailsCompleted(FetchDetailsCompleted event) {
     protected void onEnsureDebugId(String baseID) {
```
ash3rz: Looks like this can just be removed now?

de-lib/src/main/java/org/iplantc/de/theme/base/client/diskResource/details/DetailsViewDefaultAppearance.java
```diff
@@ -201,6 +203,38 @@ public String loadingMask() {
...
+    @Override
+    @SuppressWarnings("unusable-by-js")
+    public Splittable tagSearchListStyle() {
+        return StringQuoter.split("{\"fontSize\": \"small\",\"width\": \"100%\"};");
+    }
+
+    @Override
+    @SuppressWarnings("unusable-by-js")
+    public Splittable infoTypeSelectStyle() {
+        return StringQuoter.split("{\"width\": \"100px\",\"fontSize\": \"small\"};");
```
ash3rz: This would probably be more readable in `DetailsViewStyle.gss` - looks like some other CSS styles were added there already

react-components/src/apps/admin/AppStats.js
```diff
@@ -0,0 +1,328 @@
...
+                    onClick={this.handleNextButtonClick}
+                    disabled={page >= Math.ceil(count / rowsPerPage) - 1}
+                    aria-label='Next Page'
+                >
+                    {theme.direction === 'rtl' ? <KeyboardArrowLeft /> : <KeyboardArrowRight />}
+                </IconButton>
+                <IconButton
+                    onClick={this.handleLastPageButtonClick}
+                    disabled={page >= Math.ceil(count / rowsPerPage) - 1}
+                    aria-label='Last Page'
```
ash3rz: Should these First/Previous/Next/Last Page labels be moved to the appearance?  There are also a couple different styles defined that maybe should move as well?

react-components/src/data/details/InfoTypeSelectionList.js
```diff
@@ -0,0 +1,48 @@
...
+                maxHeight={200}
+                hintText={this.props.appearance.selectInfoType()}
+                style={this.props.appearance.infoTypeSelectStyle()}>
+                {items}
+            </SelectField>
+        );
+    }
+
+}
+
```
ash3rz: Might be good to define some PropTypes here since it seems certain props are expected

react-components/src/data/details/Tag.js
```diff
@@ -0,0 +1,38 @@
...
+                     onClick={this.onTagClick}>{this.props.tag.value}</div>
+                <div title={this.props.appearance.removeTag()} id={this.props.tag.value}
+                     className={this.props.appearance.css().tagRemoveStyle()}
+                     onClick={() => this.props.removeTag(this.props.tag)}> X
+                </div>
+            </div>
+        );
+    }
+}
+
```
ash3rz: Some PropTypes here would be good

react-components/src/data/details/TagPanel.js
```diff
@@ -0,0 +1,108 @@
...
+                <div id="panel" className={this.props.appearance.css().tagPanelStyle()}>
+                    { tagItems }
+                </div>
+            </div>
+        );
+    }
+
+}
+
+
```
ash3rz: Some PropTypes here would be good

react-components/stories/index.stories.js
```diff
@@ -6,7 +6,19 @@ import { action } from '@storybook/addon-actions';
...
 storiesOf('apps/details', module).add('CategoryTree', () => <CategoryTreeTest logger={action('hierarchy')} />);
 storiesOf('apps/details', module).add('ToolDetails', () => <ToolDetailsTest/>);
 storiesOf('util', module).add('CopyTextArea', () => <CopyTextAreaTest/>);
+
+storiesOf('AppStats', module).add('with test stats', () => <AppStatsTest/>);
+
+storiesOf('BasicDetails', module).add('with test diskresource details', () => <BasicDetailsTest logger={action('size')}/>);
+storiesOf('InfoTypeSelectionList',module).add('with test diskresource details', () =>  <InfoTypeSelectionListTest />);
+storiesOf('Tag', module).add('with test diskresource details', () => <TagTest logger={action('tag')}/>);
+storiesOf('TagPanel', module).add('witht test diskresource details', () => <TagPanelTest logger={action('tag')}/>);
```
ash3rz: Might be good to update these to match the nested folder structure like Paul had set up

de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/callbacks/TagsSearchCallback.java
ash3rz: I'm probably missing something, but is there a reason to use `Splittable[]` over just `Splittable`?

de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/details/DetailsViewPresenterImpl.java
ash3rz: This is probably a good time to move these strings over to the appearance like the comment says 😄

ash3rz:
Well that's annoying - I navigated away from my review and when I came back it appeared to have removed some of my comments, so I added them back, but now they're duplicated.  Sorry for the confusion.

Also, maybe some screenshots of the new views in the DE would be helpful?

de-lib/src/main/java/org/iplantc/de/admin/apps/client/AdminAppStatsGridViewImpl.java
```diff
@@ -1,222 +1,40 @@
...
-        grid.asWidget().ensureDebugId(baseID + Belphegor.AppStatIds.GRID);
-        grid.addViewReadyHandler(new ViewReadyEvent.ViewReadyHandler() {
-            @Override
-            public void onViewReady(ViewReadyEvent event) {
-                StaticIdHelper.getInstance()
-                              .gridColumnHeaders(baseID + Belphegor.AppStatIds.GRID
-                                                 + Belphegor.AppStatIds.COL_HEADER, grid);
-            }
+    @Override
+    public void load(Presenter presenter) {
```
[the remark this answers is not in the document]
sri2k1us: Yes a design discussion is definitely needed! I had mini on desk discussions with @psarando about the callbacks. It stills needs some work to add callbacks for errors returned from service.

de-lib/src/main/java/org/iplantc/de/diskResource/client/events/selection/SetInfoTypeSelected.java
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @author jstroot
+ */
+public class SetInfoTypeSelected extends GwtEvent<SetInfoTypeSelected.SetInfoTypeSelectedHandler> {
+
+    public static interface SetInfoTypeSelectedHandler extends EventHandler {
+        void onSetInfoTypeSelected(SetInfoTypeSelected event);
+    }
+
+    public static interface HasResetInfoTypeSelectedHandlers {
```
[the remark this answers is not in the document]
sri2k1us: 👍 Good catch. Missed it during refactor...

de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/callbacks/TagsSearchCallback.java
sri2k1us: I used `Splittable` array since there will be more than one tag returned. Each `Splittable` represents a`tag`. The Datasource for the MUI component I used requires an array.

de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/details/DetailsViewPresenterImpl.java
sri2k1us: 👍

de-lib/src/main/java/org/iplantc/de/theme/base/client/diskResource/details/DetailsViewDefaultAppearance.java
sri2k1us: `gss` returns a class name. MUI components can't be styled with `className` like a regular `HTML` tag since they don't have that property.

react-components/src/apps/admin/AppStats.js
sri2k1us: I wanted to but I couldn't figure out how to pass that in `props` yet!

react-components/config/webpack.config.prod.js
```diff
@@ -260,23 +261,23 @@ module.exports = {
...
+    //  },
+    //  output: {
+    //    comments: false,
+    //    // Turned on because emoji and regex is not minified properly using default
+    //    // https://github.com/facebookincubator/create-react-app/issues/2488
+    //    ascii_only: true,
+    //  },
+    //  sourceMap: shouldUseSourceMap,
+    //}),
     // Note: this won't work without ExtractTextPlugin.extract(..) in `loaders`.
```
sri2k1us: Oops I missed this in rebase I guess.

react-components/stories/index.stories.js
sri2k1us: 👍

de-lib/src/main/java/org/iplantc/de/admin/apps/client/AdminAppStatsGridViewImpl.java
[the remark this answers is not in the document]
psarando: I agree we should have a design discussion about this 👍 
I'd like to address one thing here, but we can go over it in more detail in the design discussion:

I thought the relationship of presenter to view was actually the opposite. The fact that the presenter's interface is defined in the view means that the view knows about the presenter, but the presenter does not need to know anything about the view. This also makes it easy to swap out views (e.g. replace GXT with React) without having to change the presenter.

react-components/src/apps/details.js
```diff
@@ -1,7 +1,45 @@
...
+};
+
+export {
+    CategoryTree,
+    CopyTextArea,
+    ToolDetails,
+    renderCategoryTree,
+    renderCopyTextArea,
+    renderToolDetails
+>>>>>>> 471d1ebd7... Add ToolDetails and CategoryTree React integration.
```
psarando: Yep, it looks like you need to run another `git rebase --interactive master` but remove the first 2 commits of mine that were already merged by my PR:

* Add ToolDetails and CategoryTree React integration.
* Add npm install, build, and clean tasks to gradle scripts.

react-components/config/webpack.config.prod.js
psarando: This would be fixed by a new `rebase`: https://github.com/cyverse-de/ui/pull/289#discussion_r184244301

de-lib/src/main/java/org/iplantc/de/admin/apps/client/AdminAppStatsGridViewImpl.java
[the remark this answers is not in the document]
ash3rz: I guess it's a little bit of both - the presenter doesn't care about the view as long as the view is implementing that particular interface.  The view's interface does contain the presenter's interface, but the view's implementation is ignorant - all that matters is that each event is handled by something that can at least handle that particular event (which could be several different presenters not within its interface).

react-components/src/apps/details.js
sri2k1us: Sounds good. Will do that!

react-components/src/data/details/TagPanel.js
```diff
@@ -0,0 +1,110 @@
...
+class TagPanel extends Component {
+
+    constructor(props) {
+        super(props);
+        this.handleRemoveClick = this.handleRemoveClick.bind(this);
+        this.handleTagSearch = this.handleTagSearch.bind(this);
+        this.handleTagSelect = this.handleTagSelect.bind(this);
+        this.findTag = this.findTag.bind(this);
+        this.fetchTags = this.fetchTags.bind(this);
```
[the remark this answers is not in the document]
ash3rz: In that React tutorial that I was following, it was done this way with the assignment too, which is what I've been following. From my understanding, the assignment is supposed to make it such that that method always has that particular context for `this`.  What's confusing though, is you can also override the context again at any point by doing another `*.bind(newContext)` so...  ¯\_(ツ)_/¯
If it works, your way without the assignment seems shorter at least which is nice - I haven't tried it myself.

de-lib/src/main/java/org/iplantc/de/admin/apps/client/presenter/callbacks/AppStatsSearchCallback.java
```diff
@@ -0,0 +1,14 @@
...
+
+import com.google.web.bindery.autobean.shared.Splittable;
+
+import jsinterop.annotations.JsFunction;
+
+/**
+ * Created by sriram on 3/27/18.
+ */
+@JsFunction
+@SuppressWarnings("unusable-by-js")
```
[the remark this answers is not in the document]
psarando: So what's the warning here? It thinks `Splittable[]` is unusable?

Maybe if we pass the service results as a single `Splittable results` with the array in `results.apps`, then it will get rid of the warning?

react-components/src/data/details/TagPanel.js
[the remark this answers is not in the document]
psarando: OK, I see it in this tutorial as well: https://reactjs.org/docs/handling-events.html
It looks like this is the correct way to use `bind`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this#The_bind_method

So let's keep this pattern.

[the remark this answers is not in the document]
ash3rz: I think the alternative is cleaner and it seems to provide the context you need/intend most (if not all?) of the time.  The impression I got from Tyler McGinnis's tutorial is that he thought it's still worth it to continue to use the old way and train yourself to always keep the context in mind, so I'm still using `bind` all over the place.

sri2k1us:
Will Open a new PR.

PR #290, opened 2018-05-11
CORE-9222 Increase row height to fix analysis window scroll issue

ash3rz:
Same fix as [CORE-9062](https://pods.iplantcollaborative.org/jira/browse/CORE-9062)


ash3rz:
Thanks for the reviews!

PR #292, opened 2018-05-17
CORE-9224 Update analysis reference genome dropdown with latest genomes

ash3rz:
It seems that the editor will fill the combobox with whatever was supplied to it, regardless of what's actually in the ListStore. So for this, I'm forcing the value to be chosen from the store, and in the case where the store is initially empty, I pick a value after the store has been populated.


de-lib/src/main/java/org/iplantc/de/apps/widgets/client/view/editors/arguments/converters/ReferenceGenomeEditorConverter.java
```diff
@@ -0,0 +1,52 @@
...
+    private IsField<T> field;
+
+    public ReferenceGenomeEditorConverter(IsField<T> field, Converter<Splittable, T> converter) {
+        super(field, converter);
+
+        this.field = field;
+        ComboBox<ReferenceGenome> comboBox = (ComboBox<ReferenceGenome>)field;
+        this.store = comboBox.getListView().getStore();
+        store.addStoreAddHandler(event -> {
+            updateValueFromStore();
```
psarando: slick ✨

ash3rz:
Thanks for the review!

PR #297, opened 2018-06-04
Muiv1

sri2k1us:
- Add option to pass values to reaxt-intl.
- Refactor code to pass created tag to prompt.

### DIFF
--- a/react-components/src/util/I18NWrapper.js
+++ b/react-components/src/util/I18NWrapper.js
@@ -16,8 +16,14 @@ export default function withI18N(WrappedComponent, intlData) {
     }
 }
 
-export function getMessage(id) {
+export function getMessage(id, options) {
+    if (options && options.values) {
+        return (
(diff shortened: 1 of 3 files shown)


react-components/src/util/I18NWrapper.js
```diff
@@ -16,8 +16,14 @@ export default function withI18N(WrappedComponent, intlData) {
     }
 }
 
-export function getMessage(id) {
+export function getMessage(id, options) {
+    if (options && options.values) {
```
psarando: You might be able to use destructuring here, and if `FormattedMessage` doesn't care about null/undefined `values`, then you wouldn't need the `if` check:
```jsx
export function getMessage(id, { values }) {
     return (
         <FormattedMessage id={id} values={values} />
     );
 }
```

psarando: Oh wait, I don't think this will work if the 2nd argument is null/undefined. Nevermind 😊

PR #298, opened 2018-06-05
Migrate from aphrodite to react-jss. Update MUI, react and react-dom versions.

sri2k1us:
- Note: I see some warning in storybook after migrating to JSS but I am not sure how to fix it yet. 
`data/BasicDetails/with test diskresource details error: ["Warning: Material-UI: the key `loadingStyle` provided to the classes property is not implemented in _class.
You can only override one of the following: "]
0: "Warning: Material-UI: the key `loadingStyle` provided to the classes property is not implemented in _class.
You can only override one of the following: `

### DIFF
--- a/react-components/package.json
+++ b/react-components/package.json
@@ -13,9 +13,8 @@
     "react-scripts": "^1.0.14"
   },
   "dependencies": {
-    "@material-ui/core": "^1.0.0",
+    "@material-ui/core": "^1.2.0",
     "@material-ui/icons": "^1.0.0",
-    "aphrodite": "^2.2.0",
     "autoprefixer": "7.1.2",
     "babel-core": "^6.26.0",
     "babel-eslint": "7.2.3",
@@ -46,9 +45,9 @@
     "postcss-loader": "2.0.6",
     "promise": "8.0.1",
     "rc-tree": "^1.4.1",
-    "react": "^16.3.2",
+    "react": "^16.4.0",
     "react-dev-utils": "^4.1.0",
-    "react-dom": "^16.2.0",
+    "react-dom": "^16.4.0",
     "react-intl": "^2.4.0",
     "react-jss": "^8.4.0",
     "react-select": "^1.2.1",
--- a/react-components/src/data/details/BasicDetails.js
+++ b/react-components/src/data/details/BasicDetails.js
@@ -4,18 +4,17 @@
 import React, {Component} from "react";
 import moment from "moment";
 import Button from "@material-ui/core/Button";
-import Dialog, {
-    DialogActions,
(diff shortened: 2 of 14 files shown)


react-components/src/data/details/BasicDetails.js
```diff
@@ -52,7 +51,7 @@ function Md5(props) {
             <DEHyperlink text={getMessage("view")}/>
         </td>);
     } else {
-        return <td id={props.id} className={css(styles.value)}>{getMessage("folders")}</td>;
+        return <td id={props.id} className={props.classes.detailsValue}>{getMessage("emptyValue")}</td>;
```
psarando: Did `getMessage("folders")` intentionally change to `getMessage("emptyValue")` here?

react-components/src/data/details/BasicDetails.js
```diff
@@ -319,7 +324,9 @@ class BasicDetails extends Component {
                         tags={this.state.tags.values}
                         dataSource={this.state.dataSource}
                         onTagClick={this.handleTagClick}
-                        handleTagSelect={this.handleTagSelect}/>
+                        handleTagSelect={this.handleTagSelect}
+                    />
+                    <div>
```
psarando: Is this extra `div` wrapper required around this `Dialog`?
If not, we should remove it; but if it is required, maybe the `Dialog` should be re-indented inside it.

react-components/src/data/details/BasicDetails.js
sri2k1us: It was a bug. So if MD5 is null or empty I should display empty value but it was displaying `folders` 🤦‍♂️

react-components/src/data/details/BasicDetails.js
sri2k1us: Yup those `<div>` needs to removed. I was trying to figure why `Dialog` was not working. Forgot to remove it once I figured out I need to update `react` and `react-dom`

react-components/package.json
```diff
@@ -46,9 +45,9 @@
     "postcss-loader": "2.0.6",
     "promise": "8.0.1",
     "rc-tree": "^1.4.1",
-    "react": "^16.3.2",
+    "react": "^16.4.0",
     "react-dev-utils": "^4.1.0",
-    "react-dom": "^16.2.0",
+    "react-dom": "^16.4.0",
```
ash3rz: Should `aphrodite` be removed from package.json or are we still using it elsewhere?

sri2k1us: `aphrodite` can be removed from package.json

sri2k1us:
Thanks for the reviews.

PR #299, opened 2018-06-11
core-9225 Add interactive job url to notifications.

sri2k1us:


de-lib/src/main/java/org/iplantc/de/notifications/client/utils/NotificationUtil.java
```diff
@@ -66,7 +68,10 @@
     @Inject AsyncProviderWrapper<JoinTeamRequestDialog> joinRequestDlgProvider;
     @Inject AsyncProviderWrapper<DenyJoinRequestDetailsDialog> denyDetailsDlgProvider;
     @Inject DesktopView.Presenter.DesktopPresenterAppearance appearance;
-
+    //@Inject
+    //NotificationView.NotificationViewAppearance notificationViewAppearance;  This doesn't seem to work
+    NotificationView.NotificationViewAppearance notificationViewAppearance =
+            GWT.create(NotificationView.NotificationViewAppearance.class);
     @Inject
     public NotificationUtil() {
```
ash3rz: That's weird. What if instead of injecting it with the annotation, you inject it through the constructor?

sri2k1us: I tried that as well. But that require a refactor to `NotificationCallbackConverter` since that class is not injected right now.

ash3rz: Ah ok, not really something worth worrying about in this PR 👍

PR #302, opened 2018-06-11
CORE-9080 Convert data search form to use the new search service

ash3rz:
(sorry, big PR!)
image


de-lib/src/main/java/org/iplantc/de/diskResource/client/views/search/cells/DiskResourceQueryForm.java
```diff
@@ -1,344 +1,65 @@
...
+import com.sencha.gxt.core.client.Style;
 import com.sencha.gxt.core.client.dom.XElement;
 import com.sencha.gxt.core.client.util.BaseEventPreview;
-import com.sencha.gxt.core.client.util.DateWrapper;
-import com.sencha.gxt.data.shared.LabelProvider;
 import com.sencha.gxt.widget.core.client.Composite;
-import com.sencha.gxt.widget.core.client.button.TextButton;
-import com.sencha.gxt.widget.core.client.container.AbstractHtmlLayoutContainer.HtmlData;
-import com.sencha.gxt.widget.core.client.container.HtmlLayoutContainer;
+import com.sencha.gxt.widget.core.client.container.VerticalLayoutContainer;
```
ash3rz: I still have some Sencha imports in here mostly to have the `DiskResourceSearchCell` be able to interact with the form.  Once we replace `DiskResourceSearchField`, these can all go away.

react-components/src/util/TriggerField.js
```diff
@@ -0,0 +1,95 @@
+import SearchField from './SearchField';
```
ash3rz: I added this component as well as `SearchField.js` for whenever we eventually convert the entire Data window to React.  I tried to replace `DiskResourceSearchField` with this component, but it currently looks terrible mixed in with the GXT widgets.

de-lib/src/main/java/org/iplantc/de/diskResource/client/views/search/cells/DiskResourceQueryForm.java
```diff
@@ -1,344 +1,65 @@
...
-        String includeTrash();
-
-        String searchBtnText();
-
-        String emptyTimeText();
-
-        String fileNameClass();
-
-        String createdWithinClass();
+public class DiskResourceQueryForm extends Composite implements SearchView {
```
sri2k1us: Do you have to extend from Sencha's `Composite` ?. I used GWT's `IsWidget` instead and worked for me.

de-lib/src/main/java/org/iplantc/de/diskResource/client/views/search/cells/DiskResourceQueryForm.java
```diff
@@ -354,22 +75,34 @@ public void show(Element parent, AnchorAlignment anchorAlignment) {
...
+    }
+
+    @Override
+    public void hide() {
+        if (showing) {
+            onHide();
+            RootPanel.get().remove(this);
+            eventPreview.remove();
+            showing = false;
+            hidden = true;
```
sri2k1us: I might be missing something but I don't see `hidden` used anywhere.

sri2k1us: Ahh got it, it is hiding the component when set. But In that case can we check the if the component is visible just using `hidden` and get rid of `showing` ?

react-components/src/data/search/SaveSearchButton.js
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    handleClose() {
+        this.setState({open: false})
+    }
+
+    render() {
+        let { value, onChange, parentId } = this.props;
+        let disabled = this.props.disabled;
+
```
sri2k1us: these two statements can be combined

react-components/src/data/search/EditTagDialog.js
```diff
@@ -0,0 +1,97 @@
...
+
+import Button from '@material-ui/core/Button';
+import Dialog from '@material-ui/core/Dialog';
+import DialogActions from '@material-ui/core/DialogActions';
+import DialogContent from '@material-ui/core/DialogContent';
+import DialogTitle from '@material-ui/core/DialogTitle';
+import PropTypes from 'prop-types';
+import React, {Component} from 'react';
+import TextField from '@material-ui/core/TextField';
+
```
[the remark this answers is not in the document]
ash3rz: Oh ok, I remember Codacy was complaining about not using double quotes, but I didn't realize we had decided to use double quotes too.  I'll switch 👍

react-components/src/data/search/SearchForm.js
```diff
@@ -0,0 +1,374 @@
...
+                                   id={ids.metadataAttributeHas}
+                                   label={getMessage('metadataAttributeHas')}
+                                   component={renderTextField}/>
+                        </td>
+                        <td>
+                            <Field name='ownedBy'
+                                   id={ids.ownedBy}
+                                   label={getMessage('ownedBy')}
+                                   placeholder={intl.formatMessage({id: 'enterCyVerseUserName'})}
+                                   component={renderTextField}/>
```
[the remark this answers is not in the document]
ash3rz: Yes, unfortunately the TextField placeholder will only accept a String (https://material-ui.com/api/text-field/) and getMessage returns an object

react-components/src/data/search/SaveSearchButton.js
ash3rz: Ah, good catch 👍

de-lib/src/main/java/org/iplantc/de/diskResource/client/views/search/cells/DiskResourceQueryForm.java
ash3rz: I tried GWT's Composite, let me try IsWidget

ianmcorvidae:
Only thing that vaguely pops out as weird to me is probably not an issue, which is that the "shared with" clause semantics are probably different (the old code, AFAIK, required the requesting user to be an owner, but it looks like the new one doesn't). 

Also, yay for at least reducing the number of lines of code a *little* bit, lol

de-lib/src/main/java/org/iplantc/de/client/services/impl/SearchServiceFacadeImpl.java
```diff
@@ -324,67 +270,22 @@ public void saveQueryTemplates(List<DiskResourceQueryTemplate> queryTemplates, A
...
-        addressSb.append("&includeTrash="
-                         + queryTemplate.isIncludeTrashItems());
-        addressSb.append(limitParameter);
-        addressSb.append(offsetParameter);
-        addressSb.append(typeParameter);
-        addressSb.append(sortParameter);
-        ServiceCallWrapper wrapper = new ServiceCallWrapper(GET, addressSb.toString());
-        deServiceFacade.getServiceData(wrapper, new SubmitSearchCallbackConverter(callback, queryTemplate, userInfo, drFactory));
+        StringQuoter.create(limit).assign(query, "size");
+        StringQuoter.create(offset).assign(query, "from");
```
ianmcorvidae: This is *probably* just me not knowing how StringQuoter and splittables work, but just to verify, this won't produce a string in the output, hopefully? (i.e. it should produce `{"size": 10}` and not `{"size": "10"}`, since the latter won't currently work. probably could make it work if it's a pain, but let me know anyway)

cyverse-org:
Can I suggest changing icon to magnifier as that as more common instead of
funnel.
Regards,
Nirav

de-lib/src/main/java/org/iplantc/de/client/services/impl/SearchServiceFacadeImpl.java
ash3rz: Correct - right now it's sending `{"size": 500}`

Also, I will double check the `Shared With` differences - good catch!

de-lib/src/main/java/org/iplantc/de/client/util/SearchModelUtils.java
```diff
@@ -25,6 +26,7 @@
 
     private final List<String> fileSizeUnits = Lists.newArrayList("KB", "MB", "GB", "TB");
     private final SearchAutoBeanFactory factory = GWT.create(SearchAutoBeanFactory.class);
```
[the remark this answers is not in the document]
ash3rz: I don't think so - it's a singleton.  We can inject singletons, but if there's a way to add injection within the singleton I haven't found it yet 😄 If there's a way to do it, we can refactor all of our singletons though.

de-lib/src/main/java/org/iplantc/de/client/services/impl/DataSearchQueryBuilder.java
```diff
@@ -1,480 +1,418 @@
...
-        return child;
+    /**
+     * Takes a splittable, assigns the specified key and value
+     * { "key" : { "someKey" : "someValue" } }
+     * @param keySplittable
+     * @param key
+     * @param value
+     */
+    private void assignKeyValue(Splittable keySplittable, String key, Splittable value) {
+        value.assign(keySplittable, key);
```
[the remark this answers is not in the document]
ash3rz: I'm in the same boat (obviously).  The Splittable methods are never intuitive/clear to me 😄

de-lib/src/main/java/org/iplantc/de/client/services/impl/SearchServiceFacadeImpl.java
```diff
@@ -324,67 +270,22 @@ public void saveQueryTemplates(List<DiskResourceQueryTemplate> queryTemplates, A
...
-                                                                                 searchAbFactory));
-    }
-
-    @Override
-    public void submitSearchFromQueryTemplate(DiskResourceQueryTemplate queryTemplate, FilterPagingLoadConfigBean loadConfig, TYPE searchType, AsyncCallback<List<DiskResource>> callback) {
-        DataSearchQueryBuilder builder = new DataSearchQueryBuilder(queryTemplate, userInfo);
-        String queryParameter = "";
-        String tags = "";
-        
-        String buildFullQuery = builder.buildFullQuery();
```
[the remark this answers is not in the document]
ash3rz: Ah, good catch!

de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/search/DataSearchPresenterImpl.java
```diff
@@ -256,4 +295,124 @@ private boolean templateHasChanged(DiskResourceQueryTemplate template,
...
+    }
+
+    @Override
+    public void show(XElement parent, Style.AnchorAlignment anchorAlignment) {
+        view.show(parent, anchorAlignment, getCurrentProps());
+    }
+
+    @Override
+    public void clearSearch() {
+        ReactSearchForm.SearchFormProps props = getCurrentProps();
```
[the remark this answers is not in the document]
ash3rz: Yup, another good catch 👍

react-components/src/data/search/SearchForm.js
```diff
@@ -0,0 +1,374 @@
...
+
+                    <tr>
+                        <td>
+                            <table>
+                                <tbody>
+                                <tr>
+                                    <td className={classes.fullWidth}>
+                                        <Field name='fileSizeRange.min'
+                                               type='number'
+                                               parse={value => value ? Number(value) : null}
```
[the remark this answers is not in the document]
ash3rz: It works, even with 0.  I initially had `value => Number(value)`, but then if you touched the input field at all it would automatically assign 0 as the value with no way of removing it 😖

react-components/src/data/search/SearchForm.js
```diff
@@ -0,0 +1,374 @@
...
+import TextField from '@material-ui/core/TextField';
+
+class SearchForm extends Component {
+
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            openEditTagDlg: false,
+            selectedTag: null
```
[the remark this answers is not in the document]
ash3rz: Oops, good catch. Leftover from before I refactored the tag stuff out into its own to prevent this file from being even more massive

ash3rz:
I pushed some more changes based on the PR feedback.  It ended up being a force push so an email notification with the new commits may not have gone out.  I believe the new commits should be covered [here](https://github.com/cyverse-de/ui/pull/302/files/703cb6d304133e6f95c2103f989e01a7b2f8ddfe..ab6d440988ba5c75734a1e01b071b7274e0d2efc)

react-components/src/lib.js
```diff
@@ -1,15 +1,32 @@
...
 // gwt-react needs React and ReactDOM on the global object
 window.React = React;
 window.ReactDOM = ReactDOM;
 
 
-const getDefaultTheme = () => createMuiTheme({});
+const getDefaultTheme = () => createMuiTheme({
+    palette: {
+        primary: {
+            main: palette.blue,
```
psarando: 🎉

react-components/src/lib.js
```diff
@@ -1,15 +1,32 @@
...
+        },
+        secondary: {
+            main: palette.lightGreen
```
psarando: How does `lightBlue` look for `secondary`?

The `lightGreen` reminds me of the [cyverse-ui](https://cyverse.github.io/cyverse-ui/) `success` color, but I'm not sure how much of that color palette is going to carry over to the MUI v1.x upgrade.

ash3rz: Oh right, cyverse-ui looks to be using `lightBlue`.  I didn't see us using `secondary` anywhere so I just arbitrarily picked `lightGreen`.  I'll switch it though 👍

ash3rz:
Thanks so much for the reviews! 🙇   I'll probably merge this tomorrow after I remember what all I have to update when adding new configs 😅

de-lib/src/main/java/org/iplantc/de/client/util/SearchModelUtils.java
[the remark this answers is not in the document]
sri2k1us: Please see - DEGinModule for example of Singleton Injection.
`    @Provides @Singleton public MessagePoller createMessagePoller() {
        return MessagePoller.getInstance();
    }
`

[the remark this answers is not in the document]
ash3rz: Right, I know how to inject a singleton, I do not know how to add injection *inside* of the singleton is what I was saying. You were asking if we could inject `SearchAutoBeanFactory` inside of the `SearchModelUtils` singleton. I don't think any of our singletons have injection within.

PR #303, opened 2018-06-12
CORE-8430 Adjust App Stats table layout and heights.

psarando:
This PR will adjust the App Stats table layout so that only the table rows will scroll, and not the toolbar or pagination action footer.

The App Stats table height is now also dynamic, so that it attempts to fill in all available space on window resize.

Also fixed the `ActionsComponent` prop in the `TablePagination` component so the custom pagination icons are used once again (maybe got lost in the MUI v1 upgrade?).

screenshot-2018-6-11 app stats scrolling


psarando:
I recommend selecting `Hide whitespace changes` in the `Diff settings` when reviewing these changes.

sri2k1us:
LGTM. Nicely done!

psarando:
Thanks for the reviews! 👍

PR #304, opened 2018-06-19
CORE-9246 Add analysis ID to user support request email

ash3rz:


de-lib/src/main/java/org/iplantc/de/theme/base/client/analyses/support/submit-support.html
```diff
@@ -1,4 +1,5 @@
 <p style="margin:10px;"><b>Analysis:</b> {data.name}</p>
+<p style="margin:10px;"><b>Analysis ID:</b> {data.id}</p>
```
psarando: Should this be `{data.analysisId}`?

ash3rz: No, what gets displayed on the screen to the user before they submit their request is `Analysis.id`.  For the actual email, we populate `AnalysisSupportRequestFields.analysisId` with the value in `Analysis.id`

psarando: Ah, I see. I thought this was the email template at first, since it's named `submit-support.html`.

ash3rz:
Thanks for the review!

PR #305, opened 2018-06-22
CORE-9247 Fix DE, All, and OSG Analysis Type filters

ash3rz:
The DE filter was showing all results
OSG filter didn't show any results
All filter didn't show any results if you had switched to a different filter and then switched back to All


ash3rz:
Thanks for the review!

PR #306, opened 2018-06-27
CORE-9081: switched to the CAS 3.0 protocol, which required a newer v…

slr71:
…ersion of Spring Security

### DIFF
--- a/de-webapp/src/main/resources/templates/hello.html
+++ b/de-webapp/src/main/resources/templates/hello.html
@@ -1,0 +1,10 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>Hello Spring Boot!</title>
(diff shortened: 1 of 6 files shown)


de-webapp/src/main/resources/templates/hello.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>Hello Spring Boot!</title>
+    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
+</head>
+<body>
+<p>Hello Spring Boot!</p>
+</body>
+</html>
```
slr71: The newer version of Spring Boot complained about the missing directory, so I just put a fake file in here.

slr71:
Me too. 😅  Thanks for the review!

PR #307, opened 2018-06-28
CORE-8626: gzip/bgzip'd VCF support

ianmcorvidae:
Straightforward, I think, but I'd appreciate another set of eyes. Haven't yet shipped this into info-typer or data-info, but it's in heuristomancer (moving it to the services will just be a release of heuristomancer and then update the dependencies).


ianmcorvidae:
Cool, thanks for the reviews!

PR #309, opened 2018-07-06
Core 9251 Add filtering to App Window

sri2k1us:


de-lib/src/main/java/org/iplantc/de/apps/client/views/list/widgets/AppTypeFilterCombo.java
```diff
@@ -0,0 +1,67 @@
...
+import com.sencha.gxt.widget.core.client.form.SimpleComboBox;
+
+import java.util.Arrays;
+
+/**
+ * Created by sriram on 7/3/18.
+ */
+public class AppTypeFilterCombo implements IsWidget {
+
+    private SimpleComboBox<org.iplantc.de.client.models.AppTypeFilter> filterTypeCombo;
```
ash3rz: I may have missed something - Is there a reason why this is expanded to `org.iplantc.de.client.models.AppTypeFilter` instead of just `AppTypeFilter`?

sri2k1us: Not Sure why it did that. But I fixed. Thanks for the review!

PR #310, opened 2018-07-19
Autocomplete component

ash3rz:
I've had this in my querybuilder branch for a while and figured I should probably go ahead and separate this out into its own PR. 
This is mostly just copied over from TagPanel.  I added the styled options (MenuItems), but couldn't figure out how to style the Input component to material-ui.  If you follow the example from material-ui, it looks right but the input component clears its state and loses focus if you receive new options/suggestions for some reason.

image

### DIFF
--- a/react-components/src/util/Autocomplete.js
+++ b/react-components/src/util/Autocomplete.js
@@ -1,0 +1,107 @@
+import MenuItem from "@material-ui/core/MenuItem";
+import PropTypes from "prop-types";
+import React, { Component } from "react";
+import Select from "react-select";
+import { withStyles } from "@material-ui/core/styles";
+
+/**
+ * @author aramsey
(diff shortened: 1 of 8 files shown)


react-components/src/util/Autocomplete.js
```diff
@@ -0,0 +1,107 @@
...
+    }
+});
+
+function Autocomplete(props) {
+    let {
+        variant,
+        ...custom
+    } = props;
+
+    let Component = null;
```
psarando: I might have named this to something that's not already in use (`SelectComponent`?), but since it's a local `let`, it's not a big deal either 🤷‍♂️

ash3rz: Sure, I'll rename it before merging 👍

ash3rz:
Thanks for the review!

PR #312, opened 2018-07-24
CORE-9262 Fix error when creating team with readOptin privilege

ash3rz:
The readOption privilege is fake and must be separated into 2 privileges: read and optIn


ash3rz:
Thanks for the review!

PR #315, opened 2018-08-03
Upgrade to react-select v2

ash3rz:
I tested this in GWT - the delete key works now!

image

The options appear above any elements below the Autocomplete instead of pushing those elements down - yay!
image

### DIFF
--- a/react-components/src/util/Autocomplete.js
+++ b/react-components/src/util/Autocomplete.js
@@ -1,107 +1,241 @@
+import styles from "./AutocompleteStyles";
+
+import Async from 'react-select/lib/Async';
+import AsyncCreatableSelect from 'react-select/lib/AsyncCreatable';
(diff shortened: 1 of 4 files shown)


react-components/src/util/Autocomplete.js
```diff
@@ -1,107 +1,244 @@
...
+            ValueContainer: CustomValueContainer,
+        };
 
         return (
-            <MenuItem onClick={this.handleClick}>
-                {children}
-            </MenuItem>
-        );
+            <div className={classes.root}>
+                <NoSsr>
```
sri2k1us: Do we need this tag ? We don't use server side rendering as far as I know...

ash3rz: Ah, good catch.  Left behind from some copy-paste 😊

ash3rz:
Thanks for the review!

PR #316, opened 2018-08-07
CORE-9210 Advanced Query Builder

ash3rz:
(sorry, another big PR)

image


react-components/src/collaborators/CollaboratorListIcon.js
```diff
@@ -0,0 +1,24 @@
+import messages from "./messages";
+import styles from "./styles";
+import withI18N, { getMessage } from "../util/I18NWrapper";
+
+import ListAlt from "@material-ui/icons/ListAlt";
+import React from "react";
+import ToolTip from "@material-ui/core/Tooltip";
+import { withStyles } from "@material-ui/core/styles";
```
ianmcorvidae: dangit, it lost my comment. are we planning to implement https://reactjs.org/docs/code-splitting.html#import or something at some point?

ash3rz: Ha... I think I posted early on like, "We should keep this in mind!" and then I forgot about it.  I'll start looking into it - thanks for the reminder! 😄

react-components/src/data/search/QueryBuilder.js
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    handleSaveSearch(values) {
+        let initialValues = this.props.initialValues;
+        let originalName = initialValues ? initialValues.label : null;
+        this.props.presenter.onSaveSearch(values, originalName);
+    }
+
+    render() {
+        let {
```
sri2k1us: Just being picky here but this declaration should be `const` instead of `let` because props are readonly. Mentioning only once here but applies all `let` declarations with props.

react-components/src/data/search/queryBuilder/Condition.js
```diff
@@ -0,0 +1,153 @@
...
+            obj = obj[targets[i]];
+        }
+    }
+    return obj;
+}
+
+function isGroup(condition) {
+    if (condition === "") return false;
+    return Conditions.componentMap[condition].isGroup
+}
```
sri2k1us: or ` return (condition) ? Conditions.componentMap[condition].isGroup : false;`

react-components/src/data/search/queryBuilder/Condition.js
```diff
@@ -0,0 +1,153 @@
...
+
+function ConditionComponent(condition) {
+    if (condition === "") return 'undefined';
+    return Conditions.componentMap[condition].component;
+}
```
sri2k1us: Same as above...

de-lib/src/main/java/org/iplantc/de/client/services/callbacks/ReactCallback.java
```diff
@@ -0,0 +1,14 @@
...
+import jsinterop.annotations.JsFunction;
+
+/**
+ * A generic callback function to talk between GWT and React
+ */
+@JsFunction
+@SuppressWarnings("unusable-by-js")
+public interface ReactCallback {
+    void onSuccess(Splittable data);
+}
```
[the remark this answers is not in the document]
ash3rz: Oh I knew you said you added one, but I couldn't find it so I added one 😄 I can remove mine

react-components/src/collaborators/CollaboratorListIcon.js
```diff
@@ -0,0 +1,24 @@
...
+function CollaboratorListIcon(props) {
+    let {classes} = props;
+    return (
+        <ToolTip title={getMessage('collaboratorListToolTip')}>
+            <ListAlt className={classes.searchIcon}/>
+        </ToolTip>
+    )
+}
+
+export default withStyles(styles)(withI18N(CollaboratorListIcon, messages));
```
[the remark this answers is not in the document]
sri2k1us: Ahh I missed that. Its all good then. I thought this was custom icon when I saw a custom class applied to it.

react-components/src/data/search/queryBuilder/Group.js
```diff
@@ -0,0 +1,45 @@
...
+    } = props;
+
+    let baseId = build(parentId, fields.name);
+
+    return (
+        <Fragment>
+            <AddBtn onClick={() => fields.push({})}
+                    id={build(baseId, ids.addConditionBtn)}/>
+            {!root && <DeleteBtn onClick={onRemove}
+                                 id={build(parentId, ids.deleteConditionBtn)}/>}
```
psarando: Minor suggestion: since the rows can have variable lengths, would the add/delete buttons line-up better if they were at the start of each row instead of at the end? 🤔

react-components/src/data/search/QueryBuilder.js
```diff
@@ -0,0 +1,106 @@
...
+            handleSubmit,
+            classes,
+            initialValues,
+            parentId
+        } = this.props;
+        let originalName = initialValues ? initialValues : null;
```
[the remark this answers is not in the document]
ash3rz: Oops, good catch!

react-components/src/data/search/queryBuilder/Group.js
ash3rz: If they were in the front, would we want them to still line up with each group? (It'd take a bit more effort to get the Add/Delete buttons by groups in front, so use your imagination for now 😉 )
image


I could also line up the Delete buttons at the end.  I thought it might be too easy for users to unintentionally click the Delete button on a group though and lose a bit of work, so maybe leave those where they are? 
image

de-lib/src/main/java/org/iplantc/de/diskResource/client/presenters/search/DataSearchPresenterImpl.java
```diff
@@ -364,6 +367,27 @@ public void onSuccess(Tag result) {
...
+
+            @Override
+            public void onSuccess(List<Subject> result) {
+                if (collaboratorCallback != null) {
+                    Splittable data = StringQuoter.createIndexed();
+                    result.forEach(subject -> {
+                        Splittable splSubject = AutoBeanCodex.encode(AutoBeanUtils.getAutoBean(subject));
+                        splSubject.assign(data, data.size());
+                    });
+                    collaboratorCallback.onSuccess(data);
```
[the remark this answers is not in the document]
ash3rz: Ah, I remember you making that suggestion and I completely spaced on it here.  I think it will actually only be a couple lines to change, so I'll put it in this PR so I don't forget 😉

react-components/src/data/search/queryBuilder/Group.js
sri2k1us: I think I mentioned earlier in my comment, but is it possible add more padding between the `add` and `delete` button. They are almost touching each other...

ash3rz: How does the padding look here?
image

de-lib/src/main/java/org/iplantc/de/client/services/callbacks/ReactCallback.java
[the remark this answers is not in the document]
ash3rz: I saw your generic callback is in the Notifications PR which is currently open. Since I'm ready to merge mine, I went ahead and renamed my callback to match yours so hopefully there won't be a merge conflict 🍀

ash3rz:
Thanks for the reviews!

PR #317, opened 2018-08-07
Desktop PR Updates

sri2k1us:
- Fix Bugs - CORE - 9265, 9266, 9267, 9268, 9271, 9272. 
- For CORE-9270, I have removed font-family style from body tag since, i couldn't find a way to force GXT based appearances to take the font-family set in the body tag.
- Address PR comments and refactors - #308


react-components/src/desktop/view/DesktopView.js
```diff
@@ -142,37 +144,46 @@ class DesktopView extends Component {
...
+            this.setState({unSeenCount: push_msg.total});
+        }
+
+        let message = push_msg.message;
+        let category = message.type;
+        if (category === "team") {
+            message.message.text = message.message.text + message.payload.team_name;  // attach team name to team notifications.
         }
-        if (msg.message) {
+        if (message) {
```
ash3rz: Here we're checking if there is a message, but on line 158 we're trying to get `message.type`.  Is there a possible NPE there?

react-components/src/desktop/view/DesktopView.js
```diff
@@ -189,13 +200,13 @@ class DesktopView extends Component {
...
-       this.props.presenter.doMarkAllSeen(true,(updatedUnSeenCount) => {
+    onMarkAllAsSeenClicked(shouldNotifyUser) {
+        this.props.presenter.doMarkAllSeen(shouldNotifyUser, (updatedUnSeenCount) => {
             this.setState({
-                unSeenCount: updatedUnSeenCount + "",       //count is always string
+                unSeenCount: updatedUnSeenCount,
             });
            const {notifications} = this.state;
-           notifications.forEach(function (n) {
+            notifications.messages.forEach(function (n) {
```
ash3rz: Looks like the formatting got a little offset here

react-components/src/desktop/view/Notifications.js
```diff
@@ -62,7 +64,47 @@ function NotificationFooter(props) {
...
+        notificationStyle.backgroundColor = '#99d9ea';
+        notificationStyle.borderBottom = 1;
+    }
+
+    return (
+        <span style={{outline: 'none'}}>
+                    <MenuItem
+                        id={notification.message.id}
+                        onClick={onClick}
+                        style={notificationStyle}>
```
ash3rz: Instead of defining the styles here, could we move these to the styles file?  Maybe we could have something like

```
seenNotification: {
    fontSize: 10
},
unseenNotification: {
   fontSize: 10,
   backgroundColor: '#99d9ea',
   borderBottom: 1
}
```

And then on this line, we could have:
```
<MenuItem
   id={notification.message.id}
   onClick={onClick}
   className={notification.seen ? classes.seenNotification : classes.unseenNotification}>
```

sri2k1us:
Thanks for the reviews.

PR #323, opened 2018-08-15
core-9215 fix static ids.

sri2k1us:


react-components/src/desktop/view/UserMenu.js
```diff
@@ -15,6 +15,7 @@ import tour from "../NewUserTourSteps";
 import { withStyles } from "@material-ui/core/styles";
 import PersonIcon from "@material-ui/icons/Person";
 import ReactDOM from "react-dom";
+import NotificationIcon from "../../../node_modules/@material-ui/icons/Notifications";
```
psarando: Also this `import`...

sri2k1us: 💯

PR #332, opened 2018-08-31
CORE-9136 Replace GXT Metadata Template View with a React view with nested AVU support.

psarando:
This PR will replace the GXT Metadata Template View with a React view that also has nested AVU/Attribute support.

**Note**: This PR probably should not be merged until `CORE-8699` is also ready.

This Metadata Template View uses Formik instead of `redux-form` for form creation and field validation. Formik has a similar API to `redux-form`, but I think Formik has some nicer features over `redux-form`.

For example, compare validation in Formik vs. `redux-form`, as used in https://github.com/cyverse-de/ui/blob/f7828b41731e7ced820a39ace950790fe6a9713d/react-components/src/metadata/admin/EditMetadataTemplate.js#L175-L250

In `redux-form`, each field is only passed an `error`, which comes from the validation function's `_error` key in its return object. So in order for a parent field to know if one of its nested child fields has an error, I practically had to duplicate the validation function's error object in the parent's `_error` key. Formik, on the other hand, passes the entire `error` object to each field, allow us to build it once and examine all of it in each field.

Likewise, `redux-form` does not pass the form's values to a field (only its own value). Formik will pass the entire form's values, errors, touched, setXXXX, handleXXXX, dirty, isValid, status, etc. to each field. This should make updating fields in reaction to a change in a sibling field easier. For example, in `redux-form`, in order to make only 1 Attribute enum selection marked as the default selection, the `change` function has to be passed from the top-level form all the way down to the nested fields where they need to be updated:
https://github.com/cyverse-de/ui/blob/d695edf0c6f29ed44156cb316822eaa02221b7b9/react-components/src/metadata/admin/EditMetadataTemplate.js#L153-L156
https://github.com/cyverse-de/ui/blob/d695edf0c6f29ed44156cb316822eaa02221b7b9/react-components/src/metadata/admin/AttributeEnumEditGrid.js#L34-L45

The biggest win for Formik is its [FastField](https://jaredpalmer.com/formik/docs/api/fastfield). Large forms with many nested attributes (such as DOI metadata templates) react very slowly in Firefox, since apparently a change in any field causes a change in the entire form in these libraries, causing every field to re-render. The `FastField` has its own `shouldComponentUpdate` that prevents a field from re-rendering if its own values/errors/etc. did not change.


ash3rz:
So far, doesn't seem too terribly different from redux-form.  Hopefully it proves to be faster though 👍 I've noticed that you can really tell how much slower redux-form is in FireFox when you're watching the Robot Framework tests run, so the new Advanced Search / Query Builder is suffering from some slow down as well.

react-components/src/metadata/MetadataTemplateView.js
```diff
@@ -0,0 +1,543 @@
...
+        const value = avu.value;
+
+        let attrTemplate = attributeMap[avu.attr];
+        if (!attrTemplate) {
+            return;
+        }
+
+        if (attrTemplate.required && value === "") {
+            avuErrors.value = getMessage("required");
+            avuErrors.error = true;
```
psarando: I ran into an odd feature(?) of Formik (or a bug in my code) that caused a nested child's error to always populate with an empty object `{}`, even when my validation function tried to explicitly set that field's `avuErrors` to `null`. So a simple `error && getMessage("error")` wouldn't work for displaying attribute errors. That's why this validation function sets an explicit `avuErrors.error` flag for the field to examine in its render method.

psarando:
Thanks for the review!
Any other reviews still welcome 👍

PR #333, opened 2018-09-07
CORE-9309 Add clickable icon to take users to VICE url

ash3rz:
The icon is only visible while the job is still running:
image


de-lib/src/main/java/org/iplantc/de/theme/base/client/analyses/AnalysesMessages.properties
```diff
@@ -52,3 +52,5 @@ supportRequestSuccess = Your support request is submitted. A support staff will
 userRequestingHelpSubject = requesting help with Analysis
 requestProcessing = Processing your request...
 stepInfoDialogHeader = Analysis Info
+interactiveAnalysisTooltip= Go to VICE analysis url
+
```
sri2k1us: May be `Access your interactive analyses` ? Might want to check with Science team or even Nirav. We haven't used `VICE` in app type filter.

ash3rz: Oh good point, I didn't think about the filter names.  I'll update it to match 👍

ash3rz:
Thanks for the review!

PR #334, opened 2018-09-12
CORE-9306: "Complete and Save Outputs"

ianmcorvidae:
I'm pretty sure it works, which is to say: review carefully because I have no idea what I'm doing?


de-lib/src/main/java/org/iplantc/de/analysis/client/presenter/AnalysesPresenterImpl.java
```diff
@@ -136,6 +138,28 @@ public void onSuccess(String result) {
...
+    private final class CompleteAnalysisServiceCallback extends AnalysisCallback<String> {
+        private final Analysis ae;
+
+        public CompleteAnalysisServiceCallback(final Analysis ae) {
+            this.ae = ae;
+        }
+
+        @Override
+        public void onFailure(Integer statusCode, Throwable caught) {
+            SafeHtml msg = SafeHtmlUtils.fromString(appearance.stopAnalysisError(ae.getName()));
```
ianmcorvidae: I decided this and the other string (which just refer to stopping the job, and don't use the word cancel or complete) were fine to reuse here.

de-lib/src/main/java/org/iplantc/de/client/services/impl/AnalysisServiceFacadeImpl.java
```diff
@@ -265,8 +265,8 @@ protected Void convertFrom(String object) {
     }
 
     @Override
-    public void stopAnalysis(Analysis analysis, DECallback<String> callback) {
-        String address = ANALYSES + "/" + analysis.getId() + "/stop";
+    public void stopAnalysis(Analysis analysis, DECallback<String> callback, String status) {
+        String address = ANALYSES + "/" + analysis.getId() + "/stop?job_status=" + status;
```
ianmcorvidae: I don't validate that it's a valid status here at all, but it's all hardcoded values for now, so I'm not sure if I need to?

de-lib/src/main/java/org/iplantc/de/theme/base/client/analyses/AnalysesViewDefaultAppearance.java
```diff
@@ -174,6 +179,11 @@ public String delete() {
         return iplantDisplayStrings.delete();
     }
 
+    @Override
+    public ImageResource completeIcon() {
+        return iplantResources.save();
```
ianmcorvidae: :floppy_disk: to confuse the younguns

de-lib/src/main/java/org/iplantc/de/analysis/client/presenter/AnalysesPresenterImpl.java
```diff
@@ -318,7 +344,16 @@ public void onCancelAnalysisSelected(CancelAnalysisSelected event) {
...
+            analysisService.stopAnalysis(analysis, new CancelAnalysisServiceCallback(analysis), "Canceled");
+        }
+    }
+
+    @Override
+    public void onCompleteAnalysisSelected(CompleteAnalysisSelected event) {
+        List<Analysis> analysesToComplete = event.getAnalysisList();
+
+        for (Analysis analysis : analysesToComplete) {
+            analysisService.stopAnalysis(analysis, new CompleteAnalysisServiceCallback(analysis), "Completed");
```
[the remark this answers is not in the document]
ianmcorvidae: Oh heh, that would probably make sense, assuming they turn into the same strings. I'll put a comment in and merge it for now though so we can get it into QA sooner

PR #340, opened 2018-10-05
UI-337 Create dropdown menu to replace DETabPanel in Apps

ash3rz:
This doesn't complete [UI-337](https://github.com/cyverse-de/ui/issues/337), but gets us ready to add the Communities option by switching to the dropdown.

image

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/apps/client/views/navigation/AppNavigationViewImpl.java
+++ b/de-lib/src/main/java/org/iplantc/de/apps/client/views/navigation/AppNavigationViewImpl.java
@@ -1,0 +1,120 @@
+package org.iplantc.de.apps.client.views.navigation;
+
+import org.iplantc.de.apps.client.AppNavigationView;
+
+import com.google.common.collect.Lists;
+import com.google.gwt.core.client.GWT;
+import com.google.gwt.event.logical.shared.SelectionEvent;
+import com.google.gwt.event.logical.shared.SelectionHandler;
+import com.google.gwt.event.shared.HandlerRegistration;
+import com.google.gwt.uibinder.client.UiBinder;
+import com.google.gwt.uibinder.client.UiFactory;
+import com.google.gwt.uibinder.client.UiField;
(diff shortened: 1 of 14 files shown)


ash3rz:
Thanks for the review!  
Which methods?  The ones in `AppNavigationViewImpl` I think are all overriding from the `AppNavigationView` interface so I believe those must be public.  I might have missed some others though

de-lib/src/main/java/org/iplantc/de/apps/client/views/navigation/AppNavigationViewImpl.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    @Override
+    public void add(Widget widget, String name) {
+        insert(widget, container.getWidgetCount(), name);
+    }
+
+    @Override
+    public void insert(Widget widget, int index, String name) {
+        if (!selectionMap.containsKey(widget)) {
```
psarando: I was wondering why not use a regular `ComboBox` over the `SimpleComboBox` + this `selectionMap`, but it does seem this widget lookup here is much simpler than trying to find the widget in a `ComboBox` list store 🤔

ash3rz: Well for one, I totally forgot about the regular `ComboBox` lol
But if I do remember correctly, I think `ComboBox` was better suited for a list of items of the same type which all had the same method/property that could be used to extract the display name. This is a combo of `Tree`s but they're not all of the same type (OntologyHierarchy, AppCategory, Group), so I'm not sure that would work.  I could be wrong though, it's been a while since I've used one.

ash3rz:
Thanks for the reviews!

Since the release is on Tuesday, I think I'll wait to see how far I get with the admin interface for Communities to determine if these should be merged with the release or merged afterwards.

ash3rz:
I'm going to close this since we decided to go a different direction for the communities listing in the DE - instead of having Communities be a separate tab/option, they'll be part of our current My Apps workspace tab

PR #343, opened 2018-10-31
Add static IDs to App Editor validation rule dialog

ash3rz:


ash3rz:
Thanks for the reviews!

PR #345, opened 2018-11-19
UI-336 Add community attr from de config

ash3rz:
I'm also stealing this commit from the current PR I have open for UI-336 (letting users pick communities when publishing an app). Since we're planning to release the admin interface first, then the DE side of things, I need this merged in before UI-336 gets merged in.


ash3rz:
This is a pretty small change and was already reviewed in the UI-336 PR, so I'm going to go ahead and merge.

PR #348, opened 2018-11-21
UI-337 Split Collaboration window into separate Collaborators and Teams windows

ash3rz:
This won't complete the work for UI-337 but gets us ready for adding a separate Communities window.  Note: there is still a `CollaborationView` and presenter that are used during sharing.  When you share an app, for example, there's an option to "Choose Collaborators".  That view will still have the tabbed view with Collaborators and Teams combined in tabs so users can choose multiple collaborator/group/subject types at once.

image


ash3rz:
Thanks for the review!

PR #349, opened 2018-11-29
Desktop bugs

sri2k1us:


react-components/src/desktop/style.js
```diff
@@ -231,16 +242,22 @@ export default {
     },
 
     notificationMenu: {
-        width: 400,
+        zIndex: 888888,
```
psarando: This shouldn't be required anymore, since we now set the `zIndex` for all MUI menus after the change in 3a914ac132ce25377db133973e5c03090e95b538:
https://github.com/cyverse-de/ui/blob/757ec084b73aed91592c1bce8a778caad389adf3/react-components/src/lib.js#L30-L39

sri2k1us: 👍 💯

PR #353, opened 2018-11-30
UI-346 Add retagging dialog and static IDs to communities tab

### LINKED ISSUE #346: Allow DE admins to manage communities

ash3rz:
This should resolve #346


ash3rz:
Yeah, I saw that option but wasn't sure if the admins would necessarily need it and we can always add it later if it doesn't become necessary.  I think it might come in handy for DE users to be able to force-rename when they're creating/modifying communities though.

Thanks for the review!

PR #354, opened 2018-11-30
Add missing fields to ToolAdmin

ash3rz:
New fields:
* Interactive
* Min Memory Limit
* Min/Max CPU Cores
* Skip Tmp Mount
* Min Disk Space
* OSG Image Path

Let me know if the fields could be rearranged better:
image


ash3rz:
Good questions!  

* For "skip tmp mount", I noticed in the very least `jupyter-lab-datascience-notebook` is using it, so I assume it's being used?
* Good point on the OSG image path, I'll look into possibly adding it conditionally
* Really good question, it's been 3 years since I touched this code 😄 .  We may have grayed it out because it needs to be set, but at the time we only wanted it to be set to `executable` for newly added tools.  I can try switching it to a dropdown with the types we discussed earlier.  Are any tools still set with `fAPI`?  If so, I might have to figure out a way to display `fAPI` if that's the current value but not let it be an option when creating tools.

ianmcorvidae:
I certainly hope nothing is set to fAPI, lol. The "display but don't choose" might be good for "internal" though. Of course, this is the admin interface, so if we just display all five and trust our admins in order to write less code, I personally think that's fine, lol

ianmcorvidae:
And yeah, you made me stop being lazy and look. Skip tmp mount is a thing interapps-runner knows to do, at least, so may as well have it here

ash3rz:
OK, I pushed a few updates
* The `Type` field is now a dropdown with all the types returned from `/apps/elements/tool-types`.  
* If you're creating/editing a tool, the `OSG Image Path` field will still be visible but disabled if you've not selected the OSG type.  I tried to make it not visible, but then if you change the type back to OSG, it doesn't quite render correctly and looks ugly.

ash3rz:
Thanks for the reviews!  I'll go ahead and merge this, but anyone is free to review the last couple of commits I pushed.

PR #357, opened 2018-12-18
Add container_ports to Tool Admin

ash3rz:
This should take care of the `container_ports` key.  I'll add the `interactive_apps` key next in a separate PR.  From what I could tell by our current tools, only the `container_port` value was required.

image

image


ash3rz:
Thanks for the review!

Issue #339, opened 2018-09-25
Allow tool integrators to select resource requirements


sri2k1us:

ianmcorvidae:
For the first pass, we'll add interface to select CPU and Memory limits only, with preset values. We'll also add a reminder that requesting higher amounts will result in longer waits, and perhaps recommend specifically that integrators choose the smallest working configuration (in a later pass, we can change this to properly use the minimum & maximum fields, and have the selection of a specific value move to the app interface within the range(s) set by the tool, instead).

psarando:
The functionality for this issue is now complete, but we still need some help text somewhere to remind integrators that requesting higher limits will result in longer waits.

Maybe we can put the help text in the form between the `UID` and `Max CPU Cores` fields?

PR #359, opened 2018-12-20
Update DE edit tool view with tool type and resource limit fields

psarando:
This PR will update the DE edit tool view with tool type and resource limit fields for #339.

I'm getting this PR up now before my winter break, and the max CPU cores field currently allows any double value and is not yet a selection list with preset values, so this PR could be considered a WIP until that field is also updated.

de tool resource limits screenshot_2018-12-20 discovery environment

de tool resource limits - restrictions - screenshot_2018-12-20 discovery environment


psarando:
I forgot to add `container_ports` settings for interactive tools, so this PR can be considered a WIP until that is added. Also, `working_directory` and `uid` settings should also be allowed, so I'll add those as well.

Since I'm running into issues with UI Binder and `UIFactory` when trying to create a different type of `ComboBox` for the max CPU cores selection list, I might leave that field as a double input field for this PR, then open a new PR with a React version of this private tool editor with the max CPU cores as a selection list.

ianmcorvidae:
If we merge this with just a simple long/double input, we should probably change our worker node `MODIFY_REQUEST_EXPR_` [1] settings so that they give us the benefits the "powers of 2" thing was going to as far as ensuring slots are a smaller number of relatively consistent sizes instead of a wild west array of random numbers. The defaults quantize memory to 128 bytes and cpu cores to whole numbers (and disk to multiples of 1024), but I think we can be more picky. Just let me know if that happens and I can work on it -- it won't affect *that* much, mostly just a nicety on the sys/ops side, but probably still worth doing.

[1] http://research.cs.wisc.edu/htcondor/manual/current/ConfigurationMacros.html#dx33-195180

psarando:
With the help of @sri2k1us and @ash3rz (thanks! 👍), I figured out how to get the MAX CPU values in a selection list along with selections lists for the data size limits, so this PR should be ready for review now.

de-lib/src/main/java/org/iplantc/de/client/models/tool/ToolAutoBeanFactory.java
```diff
@@ -58,4 +59,13 @@
...
     }
 
+    default void appendDefaultInteractiveAppValues(Tool tool, DEProperties deProperties) {
+        InteractiveApp interactiveApp = getInteractiveApp().as();
+        interactiveApp.setImage(deProperties.getDefaultViceImage());
+        interactiveApp.setName(deProperties.getDefaultViceName());
+        interactiveApp.setCasUrl(deProperties.getDefaultViceCasUrl());
+        interactiveApp.setCasValidate(deProperties.getDefaultViceCasValidate());
+
+        tool.getContainer().setInteractiveApps(interactiveApp);
```
ash3rz: 👍

PR #360, opened 2019-01-16
React-ified Error Handler

ash3rz:
image

### DIFF
--- a/react-components/src/util/ErrorHandler.js
+++ b/react-components/src/util/ErrorHandler.js
@@ -1,0 +1,94 @@
+/**
+ * @author aramsey
+ */
+import build from "./DebugIDUtil";
+import ids from "./ids";
+import messages from "./messages";
+import styles from "./style";
+import withI18N, { getMessage } from "./I18NWrapper";
+
+import Button from "@material-ui/core/Button";
+import Dialog from "@material-ui/core/Dialog";
+import DialogActions from "@material-ui/core/DialogActions";
+import DialogContent from "@material-ui/core/DialogContent";
+import DialogTitle from "@material-ui/core/DialogTitle";
+import ErrorIcon from "@material-ui/icons/Error";
+import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
+import ExpansionPanel from '@material-ui/core/ExpansionPanel';
+import ExpansionPanelSummary from '@material-ui/core/ExpansionPanelSummary';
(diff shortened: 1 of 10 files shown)


react-components/src/util/ErrorHandler.js
```diff
@@ -0,0 +1,94 @@
...
+            errorDetails,
+            classes,
+        } = this.props;
+
+        return (
+            <Dialog open={open}
+                    onClose={this.onClose}
+                    id={ids.errorHandlerDlg}
+            >
+                <DialogTitle>
```
sri2k1us: I mentioned this in Slack. Once Analyses PR is merged, we could use DEDialogHeader here to get uniform styling.

ash3rz:
@sri2k1us Which class were you mentioning?

sri2k1us:
> @sri2k1us Which class were you mentioning?

/de-lib/src/main/java/org/iplantc/de/commons/client/views/ReactUtil.java

ash3rz:
Ah ok, I was thinking eventually we'd have more added to `ReactUtil` but I can rename it since for now it is the only thing using it 👍

psarando:
> Ah ok, I was thinking eventually we'd have more added to `ReactUtil` but I can rename it since for now it is the only thing using it 👍

Actually I agree, I think it makes sense to keep it `ReactUtil` since this class is named for the namespace `CyVerseReactComponents.util`, and it won't have to be renamed if we export more classes from it in the future.

sri2k1us:
> > Ah ok, I was thinking eventually we'd have more added to `ReactUtil` but I can rename it since for now it is the only thing using it 👍
> 
> Actually I agree, I think it makes sense to keep it `ReactUtil` since this class is named for the namespace `CyVerseReactComponents.util`, and it won't have to be renamed if we export more classes from it in the future.
@psarando - Sorry am being picky here. If thats the case, I would rename it to ReactUtilComponents. A generic name like ReactUtil only confuses people to think it is React Utility class.

psarando:
> @psarando - Sorry am being picky here. If thats the case, I would rename it to ReactUtilComponents. A generic name like ReactUtil only confuses people to think it is React Utility class.

@sri2k1us Yeah, actually that makes sense. Something like `ReactUtilComponents` or `ReactViewUtils` would be more clear 👍

ash3rz:
I agree, I think I'll rename to `ReactUtilComponents` and then merge.

ash3rz:
Thanks for the reviews!

PR #361, opened 2019-01-17
Auto-expand MetadataTemplateView attribute panel when adding fields.

psarando:
This PR will update MetadataTemplateView attribute panels to auto-expand when new fields are added by the user.

Previously, if the attribute's ExpansionPanel was collapsed (as is the case by default if there are no AVUs matching that attribute name), and the user clicked the "Add" icon next to the attribute name, the panel would not expand on its own, and the user might not know to click the expand icon on the right side of the form to display the newly added AVU value field.

### DIFF
--- a/react-components/src/metadata/MetadataTemplateView.js
+++ b/react-components/src/metadata/MetadataTemplateView.js
@@ -78,12 +78,33 @@ const newAVU = attrTemplate => {
 };
 
 class MetadataTemplateAttributeView extends Component {
+    constructor(props) {
+        super(props);
+
+        const { avus } = props;
+        const expanded = {};
+
+        // Set all attrs as expanded in the state,
+        // even if there is no matching attribute name in the template
+        // and no expansion panel will be rendered for that attr.
+        avus && avus.forEach(avu => expanded[avu.attr] = true);
+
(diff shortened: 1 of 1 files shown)


react-components/src/metadata/MetadataTemplateView.js
```diff
@@ -236,7 +259,10 @@ class MetadataTemplateAttributeView extends Component {
                                 const hasAVUs = avuFields && avuFields.filter(avuField => avuField).length > 0;
 
                                 return ((writable || hasAVUs) &&
-                                    <ExpansionPanel key={attribute.name} defaultExpanded={hasAVUs}>
+                                    <ExpansionPanel key={attribute.name}
+                                                    expanded={!!expanded[attribute.name]}
```
psarando: The `!!expanded[attribute.name]` here forces `expanded` to a boolean value, even if a value as not yet been set for `attribute.name` in the `expanded` state map. Otherwise, if a `null` is passed as the `expanded` prop, then `ExpansionPanel` will assume it is not controlled and will manage its own internal `expanded` state.

psarando:
Thanks for the review! 👍

PR #364, opened 2019-01-23
Add class for http DELETE with a body attached

ash3rz:
I tested this with a DELETE request that contained an actual body and one that had a blank `{}` body like we've used in the past, and it seems to work.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/server/auth/BaseUrlConnector.java
+++ b/de-lib/src/main/java/org/iplantc/de/server/auth/BaseUrlConnector.java
@@ -79,8 +79,8 @@ abstract class BaseUrlConnector implements UrlConnector {
      * @param url the URL to connect to.
      * @return the request.
      */
-    protected HttpDelete createHttpDelete(String url) {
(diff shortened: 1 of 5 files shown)


ash3rz:
Yeah, that crossed my mind too.  It didn't take me long to figure this out, so if we'd rather fix it from that side, that's fine with me.  Not sure if others have thoughts on this though.  It seems to be a [topic](https://stackoverflow.com/questions/299628/is-an-entity-body-allowed-for-an-http-delete-request) for discussion, but generally seems like it should be ok?

ianmcorvidae:
It's a definite compromise. The RFCs allow it but specify there are no defined semantics: "A payload within a DELETE request message has no defined semantics;
   sending a payload body on a DELETE request might cause some existing
   implementations to reject the request." -- https://tools.ietf.org/html/rfc7231#section-4.3.5

REST in general doesn't seem to have very powerful semantics for bulk operations and personally I really dislike having to shove anything that touches more than one thing into a POST, so as long as this isn't too much of a burden to put this here I like keeping it.

PR #341, opened 2018-10-19
UI-337 Add communities to the DE

ash3rz:
NOTE: this branch is based on top of PR #340 

I think just the Communities commits should be [this](https://github.com/cyverse-de/ui/pull/341/files/0441213a1b726ceb2b358adaa2539e1e2a0c49dc..f52a7892f5f75fe1b2d717641e6576da8446cfb8)

PR #366, opened 2019-01-25
Add community listing to Apps window

ash3rz:
This is a branch on top of #365, so there are only actually [4 new commits](https://github.com/cyverse-de/ui/pull/366/files/24148141b0d34520703bd3b23071515c0eca201d..3c9d2bd87f323d4a34e66816d2cac28101955a06) here.  
Additionally, the first 3 commits of those 4 are from my first pass at adding Communities in a separate tab which was approved (#341).  So hopefully [this](https://github.com/cyverse-de/ui/pull/366/commits/3c9d2bd87f323d4a34e66816d2cac28101955a06) is the only commit that needs to be reviewed.

image

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/apps/client/views/WorkspaceViewImpl.java
+++ b/de-lib/src/main/java/org/iplantc/de/apps/client/views/WorkspaceViewImpl.java
@@ -1,0 +1,65 @@
+package org.iplantc.de.apps.client.views;
+
+import org.iplantc.de.apps.client.AppCategoriesView;
+import org.iplantc.de.apps.client.WorkspaceView;
+import org.iplantc.de.apps.shared.AppsModule;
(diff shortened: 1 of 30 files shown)


de-lib/src/main/java/org/iplantc/de/apps/client/views/WorkspaceViewImpl.java
```diff
@@ -0,0 +1,65 @@
...
+        });
+    }
+
+    @Override
+    public void selectFirstItem() {
+        if (trees != null) {
+            Tree firstTree = trees.get(0);
+            List rootItems = firstTree.getStore().getRootItems();
+            if (rootItems != null && rootItems.size() > 0) {
+                trees.get(0).getSelectionModel().select(rootItems.get(0), false);
```
psarando: Should also be `firstTree` here?

PR #371, opened 2019-02-01
Attempt to use `react-scripts` as build dependency.

psarando:
This PR is an attempt to replace deps such as `webpack` and `babel` with `react-scripts` in our package dependencies.
It will also update our webpack config with latest CRA config settings for `babel`.


sri2k1us:
LGTM: Fingers crossed!

ash3rz:
🚢 let's give it a go!

psarando:
Thanks for the reviews! 🤞

PR #374, opened 2019-02-05
9282

sri2k1us:


react-components/src/util/table/EnhancedTableHead.js
```diff
@@ -27,22 +27,23 @@ class EnhancedTableHead extends React.Component {
             order,
             orderBy,
             numSelected,
-            rowCount,
             columnData,
             selectable,
             classes,
             padding,
+            rowInPage
```
ash3rz: Changing this prop name might break any components currently using `rowCount` (not sure if there are any, but thought I'd mention it).  When I first read this, I thought `rowInPage` was referring to a particular row within the page, not the total number of rows in a page.  Maybe `rowsPerPage` would be clearer?

sri2k1us: Makes sense. I will update prop name. And yes, I am going update all places where this component being used.

sri2k1us: Actually, I will name it `rowsInPage` cos it is not the page size but the actual number of rows in the page. For example, you set `100` row per page from bottom pagination toolbar but the number of rows in that page be just 10. This prop refers to `10`.

ash3rz: Good point - `rowsInPage` sounds good

sri2k1us:
Closing this PR. Will re-open a new one.

PR #375, opened 2019-02-07
Webpack config updates

ash3rz:
I tried and couldn't figure out the minimal amount of changes to our configs in order to get source code back into dev tools.  This could be used more as a last resort, but I basically just copied over all the settings from create-react-app, and then added in our DE-specific changes.  It seems to work, I can see source code again, and `createProdWar` didn't crash, so I assume that will also work.


build.gradle
```diff
@@ -71,13 +71,20 @@ task npmTest(type:Exec, dependsOn:npmInstall) {
...
 
-task webpack(type:Exec, dependsOn:npmInstall) {
+task webpackDev(type:Exec, dependsOn:npmInstall) {
     group 'Build'
     description 'Builds the react-components directory.'
     workingDir 'react-components'
     commandLine 'npm', 'run', 'build'
 }
 
+task webpackProd(type:Exec, dependsOn:npmInstall) {
```
psarando: The changes to this file shouldn't be necessary (see comments below for `package.json`).

iplant/webapps.gradle
```diff
@@ -112,15 +112,15 @@ project(':de-webapp') {
     }
 
     task draftBootRepackage(type: BootRepackage, dependsOn: draftWar) {
-        dependsOn webpack
+        dependsOn webpackDev
```
psarando: The changes to this file shouldn't be necessary (see comments below for `package.json`).

react-components/package.json
```diff
@@ -43,7 +43,8 @@
   },
   "scripts": {
     "start": "npm run storybook",
-    "build": "node scripts/build.js",
+    "build": "REACT_APP_MODE=development node scripts/build.js",
+    "prod": "REACT_APP_MODE=production node scripts/build.js",
```
psarando: The changes to this file shouldn't be necessary, since the `development` env is used by CRA when running `npm start` and launching the react app in a local dev session. We don't need that because we use `storybook` for local dev instead.

If you create a React app and eject it, you'll see a `build.js` file (like the one in this PR) and a `start.js` script. That `start.js` script is where CRA sets `process.env.BABEL_ENV = 'development'` and `process.env.NODE_ENV = 'development'`.

They used to have separate `webpack.config.dev.js` and `webpack.config.prod.js` files for dev vs. prod, and that's why we are just using one named `webpack.config.prod.js` (from when we ejected a CRA v1.0.14 app). In CRA 2.x, it looks like they've combined these files into one, and are using those `isEnvX` vars to tweak the configs. We only need the prod configs, though our time is probably best spent on something other than refactoring all those config changes for a prod-only config 😅

react-components/scripts/build.js
```diff
@@ -1,8 +1,8 @@
 'use strict';
 
 // Do this as the first thing so that any code reading it knows the right env.
-process.env.BABEL_ENV = 'production';
-process.env.NODE_ENV = 'production';
+process.env.BABEL_ENV = process.env.REACT_APP_MODE;
+process.env.NODE_ENV = process.env.REACT_APP_MODE;
```
psarando: We should be able to leave these set to `production`.

react-components/config/webpack.config.prod.js
```diff
@@ -1,386 +1,612 @@
...
-// To have this structure working with relative paths, we have to use custom options.
-const miniCssExtractPluginOptions = shouldUseRelativeAssetPaths
-  ? // Making sure that the publicPath goes back to to build folder.
-    { publicPath: Array(cssFilename.split('/').length).join('../') }
-  : {};
+// This is the production and development configuration.
+// It is focused on developer experience, fast rebuilds, and a minimal bundle.
+module.exports = function(webpackEnv) {
+    const isEnvDevelopment = webpackEnv === 'development';
+    const isEnvProduction = webpackEnv === 'production';
```
psarando: We should only be building for `production`, since we use `storybook` for dev, but our time is probably best spent on something else than refactoring all these changes for a prod-only config 😅

ash3rz:
Ah, yeah, I was mostly thinking if we couldn't figure anything else out, here's a solution that seems to work.  I like that we're more closely matched to CRA, but I also would prefer less changes.  Now that I understand the history between the dev and prod configs and why we've been using prod, I'll poke at it a bit more to see if I can narrow it down.

PR #367, opened 2019-01-28
React update

sri2k1us:
- Update `react-gwt` to latest version
- Update to `webpack` 4.x
- Update to `React 16.7.0`
- Update `GWT` to `2.8.2`
- Update `storybook` (and related) for react to `4.1.7`
- Update `react-scripts` to latest version

And many more version updates. See `package.json` for full list.

PR #376, opened 2019-02-08
Remove UglifyJsPlugin - causing invalid sourcemaps

ash3rz:
This is the minimal amount of changes to get JS code viewable in dev tools again

### DIFF
--- a/react-components/config/webpack.config.prod.js
+++ b/react-components/config/webpack.config.prod.js
@@ -14,7 +14,6 @@ const eslintFormatter = require('react-dev-utils/eslintFormatter');
 const ModuleScopePlugin = require('react-dev-utils/ModuleScopePlugin');
 const paths = require('./paths');
 const getClientEnvironment = require('./env');
-const UglifyJsPlugin = require("uglifyjs-webpack-plugin");
 
 // Webpack uses `publicPath` to determine where the app is being served from.
 // It requires a trailing slash, or the file assets will get an incorrect path.
@@ -313,16 +312,6 @@ module.exports = {
     // It is absolutely essential that NODE_ENV was set to production here.
     // Otherwise React will be compiled in the very slow development mode.
     new webpack.DefinePlugin(env.stringified),
-    // Minify the code.
-    new UglifyJsPlugin({
-      uglifyOptions: {
-        warnings: false,
-        ie8: false,
-        output: {
-          comments: false
-        }
-      }
-    }),
     new MiniCssExtractPlugin({
       // Options similar to the same options in webpackOptions.output
       // both options are optional


react-components/config/webpack.config.prod.js
```diff
@@ -313,16 +312,6 @@ module.exports = {
...
-    new UglifyJsPlugin({
-      uglifyOptions: {
-        warnings: false,
-        ie8: false,
-        output: {
-          comments: false
-        }
-      }
-    }),
     new MiniCssExtractPlugin({
```
sri2k1us: Does this mean our production build size will increase ? Because we are removing `Minify`

psarando: Maybe not, since we added this `TerserPlugin` in #367:
https://github.com/cyverse-de/ui/blob/299c97c9a4478a4b66a89912ec9ad408eca6a62c/react-components/config/webpack.config.prod.js#L80-L84

There's also minify options in the `HtmlWebpackPlugin` config?
https://github.com/cyverse-de/ui/blob/299c97c9a4478a4b66a89912ec9ad408eca6a62c/react-components/config/webpack.config.prod.js#L293-L310

sri2k1us: 👍. Anyway it will be good to compare the resulting war size.

ash3rz: It is slightly bigger.  Doing a comparison, it looks like the latest CRA has some other things added in that make the size smaller.

Before (with uglify):
```
File sizes after gzip:

  273.78 KB     react_components/static/js/1.4b0fddba.chunk.js
  63.72 KB      react_components/static/js/main.6be962e7.chunk.js
  12.62 KB      react_components/static/css/1.4b0fddba.chunk.css
  745 B         react_components/static/js/runtime~main.6172aa93.js
  67 B          react_components/static/media/index.83c8bca1.css
```

After (uglify removed):
```
File sizes after gzip:

  275.67 KB (+1.89 KB)  react_components/static/js/1.4b0fddba.chunk.js
  63.98 KB (+263 B)     react_components/static/js/main.6be962e7.chunk.js
  12.62 KB              react_components/static/css/1.4b0fddba.chunk.css
  793 B (+48 B)         react_components/static/js/runtime~main.6172aa93.js
  67 B                  react_components/static/media/index.83c8bca1.css
```

With CRA Prod config:
```
File sizes after gzip:

  271.86 KB (-3.8 KB)  react_components/static/js/1.d1089dff.chunk.js
  63.92 KB (-62 B)     react_components/static/js/main.ae2727bd.chunk.js
  12.62 KB             react_components/static/css/1.e8b05bf8.chunk.css
  793 B                react_components/static/js/runtime~main.6172aa93.js
```

ash3rz: Also yes, looks like CRA switched from Uglify to Terser back in September

PR #377, opened 2019-02-08
CORE-25 Replace redux-form with Formik in EditMetadataTemplate

psarando:
This PR will replace `redux-form` fields with `Formik` fields in the admin `EditMetadataTemplate` dialog and in `util/FormField`.

Only the `data/search/QueryBuilder` forms now use `redux-form`, but I think we can convert those in a later PR.

### DIFF
--- a/react-components/src/metadata/admin/AttributeEnumEditGrid.js
+++ b/react-components/src/metadata/admin/AttributeEnumEditGrid.js
@@ -2,7 +2,8 @@
  * @author psarando
  */
 import React, { Component } from "react";
-import { Field } from "redux-form";
+import { FastField, Field, getIn } from "formik";
+import PropTypes from "prop-types";
 import { injectIntl } from "react-intl";
 
 import build from "../../util/DebugIDUtil";
@@ -11,7 +12,7 @@ import intlData from "../messages";
 import styles from "../style";
 import ids from "./ids";
 
-import { FormCheckbox, FormCheckboxTableCell, FormTextField } from "../../util/FormField";
+import { FormCheckbox, FormCheckboxTableCell, FormTextField, getFormError } from "../../util/FormField";
 import OrderedGridToolbar from "./OrderedGridToolbar";
 
 import Button from "@material-ui/core/Button";
@@ -29,14 +30,16 @@ import { withStyles } from "@material-ui/core/styles";
 
 import ContentRemove from "@material-ui/icons/Delete";
 import ContentEdit from "@material-ui/icons/Edit";
-import PropTypes from "prop-types";
 
-const normalizeDefault = (fields, change, is_default) => {
+const normalizeDefault = (enumValues, replace) => event => {
+    event.stopPropagation();
+
+    const is_default = event.target.checked;
+
     if (is_default) {
-        fields.forEach((field, index) => {
-            let attrEnum = fields.get(index);
+        enumValues.forEach((attrEnum, index) => {
             if (attrEnum.is_default) {
-                change(field, {...attrEnum, is_default: false});
+                replace(index, {...attrEnum, is_default: false});
             }
         });
     }
@@ -45,25 +48,12 @@ const normalizeDefault = (fields, change, is_default) => {
 };
 
 class AttributeEnumEditDialog extends Component {
-    constructor(props) {
-        super(props);
-
-        [
-            "normalizeDefaultField",
-        ].forEach(methodName => (this[methodName] = this[methodName].bind(this)));
-    }
-
     static propTypes = {
         onClose: PropTypes.func.isRequired,
     };
 
-    normalizeDefaultField(is_default) {
-        const { fields, change } = this.props;
-        return normalizeDefault(fields, change, is_default);
-    }
-
     render() {
-        const { open, intl, field, error, onClose } = this.props;
+        const { open, intl, field, error, replace, enumValues, onClose } = this.props;
 
         const dialogID = build(ids.METADATA_TEMPLATE_FORM, field, ids.DIALOG);
         const dialogTitleID = build(dialogID, ids.TITLE);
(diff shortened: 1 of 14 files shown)


react-components/src/metadata/admin/EditMetadataTemplate.js
```diff
@@ -85,8 +63,8 @@ class EditMetadataTemplate extends Component {
             classes,
             intl,
             open,
-            // from redux-form
-            handleSubmit, pristine, submitting, error, change,
```
psarando: 🎉 We no longer have to pass a `change` function from the top-level form down to each attribute and enum editor, since Formik provides each field access to top-level form functions like `setFieldValue `.

react-components/src/metadata/admin/EditMetadataTemplate.js
```diff
@@ -235,25 +212,57 @@ const validate = values => {
...
+        },
+        resolve,
+        errorCallback,
+    );
+};
+
+const mapPropsToValues = props => {
+    const { template } = props;
+
+    return { ...template };
```
psarando: The `initialValues` prop is not required by a Formik form (though it will populate its own `initialValues` field in its form props). This `mapPropsToValues` function allows us to populate the form's `values` prop with anything given in the initial props passed to our form component.

For another example:
https://github.com/cyverse-de/ui/blob/299c97c9a4478a4b66a89912ec9ad408eca6a62c/react-components/src/metadata/MetadataTemplateView.js#L421-L468

react-components/src/metadata/admin/EditMetadataTemplate.js
```diff
@@ -174,38 +151,42 @@ class EditMetadataTemplate extends Component {
...
 
         // Validate Enum values
         if (attr.type === "Enum") {
             if (!attr.values || attr.values.length < 1) {
-                // Setting an attrErrors.values["_error"] message allows the error to be displayed in the table header.
-                attrErrors.values = [];
-                attrErrors.values["_error"] = getMessage("errEnumValueRequired");
+                // Setting an attrErrors.values message allows the error to be displayed in the table header,
+                // but that component will have to check first if attrErrors.values is an array or not.
+                attrErrors.values = getMessage("errEnumValueRequired");
```
psarando: Here's an example of how I think Formik's `validate` function is more flexible than `redux-form`, which requires an `_error` to be set in each field's validation object if you want to access that custom error in the field's component.

On the other hand, Formik passes the entire `errors` object to each field, so they can access whatever is set in this function.

react-components/src/metadata/admin/EditMetadataTemplate.js
```diff
@@ -235,25 +212,57 @@ const validate = values => {
...
+    const errorCallback = (httpStatusCode, errorMessage) => {
+        setSubmitting(false);
+        setStatus({ success: false, errorMessage });
+    };
+
+    const { id } = props.template;
+
+    props.presenter.onSaveTemplate(
         {
-            form: ids.METADATA_TEMPLATE_FORM,
```
psarando: Since Formik doesn't use redux, there's no need for a global form ID.

react-components/src/metadata/admin/EditAttributeFormList.js
```diff
@@ -175,39 +176,48 @@ class EditAttribute extends Component {
...
+                            <FastField name={`${field}.type`}
+                                       label={getMessage("attrTypeLabel")}
+                                       id={build(formID, ids.ATTR_TYPE)}
+                                       render={({ field: { onChange, ...field }, ...props }) => (
+                                           <FormSelectField
+                                               {...props}
+                                               field={field}
+                                               onChange={event => {
+                                                   this.normalizeType(event.target.value);
+                                                   onChange(event);
```
psarando: I didn't see any post-change `normalize` support in Formik fields, so manually calling the Formik field's `onChange` function in our component's custom `onChange` function, which can do its own normalization, is one way to implement this now.

react-components/src/metadata/admin/EditAttributeFormList.js
```diff
@@ -223,9 +233,12 @@ class EditAttribute extends Component {
                                 </ExpansionPanelSummary>
                                 <ExpansionPanelDetails>
                                     <FieldArray name={`${field}.values`}
-                                                component={AttributeEnumEditGrid}
-                                                parentID={formID}
-                                                change={change}
+                                                render={arrayHelpers => (
+                                                    <AttributeEnumEditGrid
+                                                        {...arrayHelpers}
+                                                        parentID={formID}
```
psarando: Unfortunately Formik's `FieldArray` won't pass extra props down to its given `component`, so in order to pass things like this `parentID` prop to the inner `component`, I had to use the FieldArray's `render` prop instead of simply `component`.

react-components/src/metadata/admin/AttributeEnumEditGrid.js
```diff
@@ -77,20 +67,20 @@ class AttributeEnumEditDialog extends Component {
...
+                               margin="dense"
+                               component={FormTextField}
                     />
                     <Field name={`${field}.is_default`}
                            label={formatMessage(intl, "enumDefaultLabel")}
                            id={build(dialogID, ids.ENUM_VALUE_DEFAULT)}
                            color="primary"
                            component={FormCheckbox}
-                           normalize={this.normalizeDefaultField}
+                           onClick={normalizeDefault(enumValues, replace)}
```
psarando: Unfortunately an attribute enum `is_default` checkbox (here and in the enum listing grid) can't be a `FastField`, otherwise this enum values list passed to the `normalizeDefault` function can become stale, and the `normalizeDefault` function won't work properly (if a value becomes checked, this function needs to uncheck any other value with the FieldArray's `replace` function, since there can be only one default checked).

react-components/src/metadata/admin/EditAttributeFormList.js
sri2k1us: 👍

react-components/src/metadata/admin/EditMetadataTemplate.js
sri2k1us: 👍

react-components/src/metadata/admin/EditMetadataTemplate.js
```diff
@@ -105,9 +83,9 @@ class EditMetadataTemplate extends Component {
                         <IconButton id={build(ids.METADATA_TEMPLATE_FORM, ids.BUTTONS.CLOSE)}
                                     aria-label={formatMessage(intl, "close")}
                                     onClick={() => (
-                                        pristine ?
-                                            this.props.presenter.closeMetadataTemplateDialog() :
-                                            this.setState({showConfirmationDialog: true})
+                                        dirty ?
```
ash3rz: Should this be `!dirty ? ...`  Maybe I'm misunderstanding the change from `pristine` to `dirty` though.  Below you have a similar change from `pristine` to `!dirty` in a button

psarando: Ah, no, because the `then/else` lines have also swapped here 😉

psarando:
Thanks for the reviews! 👍

PR #378, opened 2019-02-12
9282

sri2k1us:
I want to hear thoughts on the Kludge i put together to reconfigure analyses window when an analyses notification clicked from desktop. 

Refactoring Analyses window to manage state from the GWT presenter and pass down analyses listing to react (as suggested in slack) is a big effort and will at least take 2 weeks plus testing.

Going forward we need agree upon a set methodology on how we are going to handle global state management in GWT + React.

This PR also includes fixes for the following:
* Pagination / Selection issues
* Double scrollbars in Analyses / Notification windows

### DIFF
--- a/react-components/package.json
+++ b/react-components/package.json
@@ -56,5 +56,11 @@
   },
   "eslintConfig": {
     "extends": "react-app"
-  }
+  },
+  "browserslist": [
+    ">0.2%",
+    "not dead",
+    "not ie <= 11",
+    "not op_mini all"
+  ]
 }
--- a/react-components/src/analysis/view/AnalysesView.js
+++ b/react-components/src/analysis/view/AnalysesView.js
@@ -362,11 +362,13 @@ class AnalysesView extends Component {
 
     handleChangePage(event, page) {
         const {rowsPerPage} = this.state;
-        this.setState({page: page, offset: rowsPerPage * page});
+        //reset selection between pages
+        this.setState({page: page, offset: rowsPerPage * page, selected: []},
+            () => this.fetchAnalyses());
     }
 
     handleChangeRowsPerPage(event) {
-        this.setState({rowsPerPage: event.target.value});
+        this.setState({rowsPerPage: event.target.value}, () => this.fetchAnalyses());
     }
 
     handleSelectAllClick(event, checked) {
@@ -826,11 +828,11 @@ class AnalysesView extends Component {
                                 orderBy={orderBy}
                                 onSelectAllClick={this.handleSelectAllClick}
                                 onRequestSort={this.handleRequestSort}
-                                rowCount={total}
                                 columnData={columnData}
                                 baseId={baseId}
                                 ids={ids}
                                 padding="none"
+                                rowInPage={data.length}
                             />
                             <TableBody>
                                 {data.map(analysis => {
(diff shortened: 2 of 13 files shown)


react-components/src/analysis/view/AnalysesView.js
```diff
@@ -225,10 +225,33 @@ class AnalysesView extends Component {
...
-    fetchAnalyses() {
+    componentDidUpdate(prevProps) {
+        //this is bad. react recommends enclosing this in conditional statement.
+        //But if i add
+        //prevProps.selectedAnalysisId !== this.props.selectedAnalysisId
+        //check, i cannot repeat the same operation of clicking the notification to reconfigure
+        // AnalysesView. For example: Open Analyses Window, click on a notification, Analyses window will
+        // reconfigure. Repeat the same step after clearing search field, it won't work since the prop
+        // hasn't changed.
+        //Kludge - Pass a boolean to presenter which in turn resets selected analyses.
```
psarando: I spoke too hastily in slack before reviewing this entire component, and I don't think we have to refactor the `fetchAnalyses` method back into the GWT presenter, but I think part of the complication here comes from copying these listing filter props into this component's state.

If all of these filter values were passed in as props, then all of them could be checked here for changes before calling `fetchAnalyses`.

When the search field or a drop-down changes a filter value, then a callback into the presenter can update the props from the GWT side, and the `AnalysesViewImpl` can update its props passed to this component. The `AnalysesViewImpl` can also keep a static copy of the props and only update where required, similar to the dialog `open` prop in the `MetadataTemplateView`:
https://github.com/cyverse-de/ui/blob/11127b4aec59c268c7fc6c488bd5436fa30b9da0/de-lib/src/main/java/org/iplantc/de/diskResource/client/views/metadata/MetadataTemplateView.java#L22

All the filter objects can still be built by `fetchAnalyses` before passing them to the presenter's service call, but instead of building them from values in state, they can be created with values from props.

This would mean refactoring the logic in `handleBatchIconClick`, `onPermissionsFilterChange`, `onTypeFilterChange`, and especially `handleSearch` into the GWT Analyses presenter (and these methods would callback into the presenter instead of setting state, but all other methods that update state and fetch a new listing can be left as is). Then the presenter would figure out the correct values of these filter props (as `handleBatchIconClick` and `handleSearch` are doing now), then call new methods in `AnalysesViewImpl`, which can then update its copy of the props and pass them down to this component. Then this `componentDidUpdate` can detect if any filters have changed and call `fetchAnalyses` with new filter objects.

react-components/package.json
```diff
@@ -56,5 +56,11 @@
   },
   "eslintConfig": {
     "extends": "react-app"
-  }
+  },
+  "browserslist": [
```
psarando: 👍

react-components/src/analysis/view/AnalysesView.js
sri2k1us: This sounds like a plan. Let me give this a try.

ash3rz: Sriram, it sounds like you have a path forward.  Do we need a design discussion though to talk about methodology going forward?

sri2k1us: Yes. A DD wouldn't hurt. We should all be in the same page with the approach we take 👍

PR #379, opened 2019-02-19
CORE-103 Enhance EnhancedTableHead

psarando:
This PR will enhance `EnhancedTableHead` with changes that will be used for CORE-7743, plus some minor refactoring for code readability: 

* Added `id` field to `columnData` props, instead of auto generating based on name/key and the `ids` prop, which also fixed a bug in `RequestHistoryDialog`.
* Added an optional `padding` field to `columnData` props, which can override the overall `padding` prop.
* Check `onRequestSort` prop before setting a column sort handler (fixes NPE in `analysis` stories).

### DIFF
--- a/react-components/src/analysis/view/AnalysesView.js
+++ b/react-components/src/analysis/view/AnalysesView.js
@@ -122,13 +122,53 @@ function Status(props) {
 }
 
 const columnData = [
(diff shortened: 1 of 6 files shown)


psarando:
Thanks for the reviews! 👍

PR #382, opened 2019-02-19
Communities with buttons moved to listing

ash3rz:
### **BEFORE**

Listing:
image

Admin View:
image

Detail View:
image

### **AFTER**

Listing:
image

Admin View:
image

Detail View:
image


react-components/src/communities/view/CommunityListing.js
```diff
@@ -44,10 +45,19 @@ class CommunityListing extends Component {
         this.setState({order, orderBy});
     }
 
+    handleRowClick(community) {
+        this.setState({
+            selectedCommunity: community
+        });
+
+        this.props.onCommunitySelected(community);
```
psarando: Since we're passing the selected community back up to the `ManageCommunitiesView`, we could also have that view pass the selected community back down to this component as a prop, rather than also storing it in this component's state.

I'm not sure if there's a benefit of storing the selected community in this component's state vs. getting it from props, but I get the impression that React prefers controlling components with props over state (where it makes sense). In this case, it could also open up the possibility of telling this listing component that a community has been selected in some way other than the user clicking a row (though I don't know if we'll ever need that functionality 🤔 ).

ash3rz:
Thanks for the review!  So I guess the final question is which way do we want to go?  I have the BEFORE and AFTER pictures in the PR description.  

Visually, I could go either way at this point - I think it makes sense to open up the details and then decide to join/leave/delete the community.  I think it also can make sense to have those buttons in the listing, and leave the details dialog clutter-free of those buttons.

Functionally, it does add extra time to query the permissions and membership status for each community in order to have that data in the listing to enable/disable those buttons appropriately.  As more communities get added, the longer it will take to fetch the listing.  If a user joins a lot of communities, that could potentially also increase how long it takes to open the Apps window unless there's something different we want to do there as well?  Right now, the app category listing takes longer (at least for me) so there's no difference in time.

react-components/src/communities/view/CommunityListing.js
ash3rz: Sure, I'll refactor!  Did we still want a design discussion on our state management strategy going forward?  With communities, I'm definitely managing most of it in the React components instead of in GWT, similar to what we had in the Analysis window, so I may need more refactors depending on what we decide.

sri2k1us:
The idea behind communities is to have a focused set of apps, tools and data relevant to the users area of research. I am not saying users wouldn't join a lot communities but I think we can wait until that happens.

cyverse-org:
It’s actually a little more insidious than that. The lookup time is proportional to the number of privileges that the user has to anything in Grouper. Since being able to see something in Grouper is also defined as a privilege, the lookup time will increase with the number of communities defined, rather than the number of communities that the user belongs to. We may not have to change it right away, but this is definitely something to keep in mind if any performance issue emerge.




On February 26, 2019 at 1:35:58 PM, Sriram Srinivasan (notifications@github.com) wrote:

The idea behind communities is to have a focused set of apps, tools and data relevant to the users area of research. I am not saying users wouldn't join a lot communities but I think we can wait until that happens.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub, or mute the thread.

_______________________________________________  
Core-sw mailing list  
Core-sw@mailgate.cyverse.org  
https://maillist.cyverse.org/mailman/listinfo/core-sw

psarando:
I'm OK with the old functionality, where you join/leave/delete from the details view. Especially if this helps the listings performance. Though there's something a bit awkward to me about having those action buttons down by the dialog's OK/Cancel buttons.

Maybe having those actions above the community's name and description would help (similar to the `+APPS` button above the apps listing)? Then we could also shove the `Delete` action all the way to the right side of the dialog, so it's not so close to any other actions. It could also have a red background, when hovering over it, to indicate the `Danger Zone`.

react-components/src/communities/view/CommunityListing.js
psarando: We can. I'm not sure we can nail down an overarching strategy (beyond what I just mentioned above with props over state where reasonable), but if we have questions or want further clarifications about it, let's discuss it!

sri2k1us:
Lets go with having join/leave/delete from details view given there are concerns about performance. I agree with @psarando suggestions to move the actions to top of the details screen (may be a toolbar) and just have `OK` and `Cancel` at the bottom. A few other minor things to consider 

- Add `textTransformation` css so that button text is not all caps
-  Add some tool tip to community name in the listing that says something like `Click to view details, join, leave or delete the community` 

This is for future - We can consider caching communities on the client side (browser local storage) like pre-fetching them right after login. This cache can be updated whenever an error happens or after every login. For example, User request to delete a community but suddenly he / she has lost that privilege because another community admin has removed this user from admin privileges. In this case, the UI can refetch the communities and update the cache.

ash3rz:
Sounds good - I'll close this PR and reopen a new one with the refactors for the details view.  It sounds like @slr71 can revert the changes to add privileges and membership status to community listing then in that case.

PR #385, opened 2019-02-26
CORE-120 Fix EnhancedTableHead rendering with FAB and Checkboxes.

psarando:
Something about MUI's CSS for `Fab` and `Checkbox` components make them float over our stickied `EnhancedTableHead` when scrolling through tables. Adding the `EnhancedTableHead` to a `Table` after a `TableBody` prevents this odd behavior.

Also updated `EnhancedTableHead` and `NotificationView` checkbox column padding.


sri2k1us:
LGTM. As much this solution works but i can't prevent a 🤦‍♂️ reaction.

psarando:
Thanks for the reviews! 👍

PR #386, opened 2019-02-27
Make sure query parameter windows are on top

ash3rz:
Basically:
1. If there's a DATA parameter, only the Data window will open regardless of saved session (this is the original functionality for "data only" viewer that I'm maintaining)
1. The user's session is fetched
1. Any other parameters are then parsed (apps, error, state)


ash3rz:
Thanks for the review!

PR #388, opened 2019-02-27
Move join, leave, delete community buttons to top of Edit dialog

ash3rz:
**Admin View**
image

**Detail View**
image


ash3rz:
Thanks for the review!  I think I'll update the buttons to more match [material's dialog action](https://material.io/design/components/dialogs.html#actions) spec, and then merge.

react-components/src/util/dialog/DEConfirmationDialog.js
```diff
@@ -17,10 +17,20 @@ import build from "../DebugIDUtil";
...
+            onOkBtnClick,
+            onCancelBtnClick,
+            debugId,
+            okLabel,
+            messages
+        } = this.props;
         return (
             <Dialog open={dialogOpen} id={debugId}>
                 <DEDialogHeader
+                    messages={messages}
```
psarando: I'm guessing this `messages` prop is passed in from a parent component so that `getMessage` can find the appropriate intl messages?

In order to accomplish this in my metadata branch, I was creating a wrapped `DEConfirmationDialog` with the `intlData` that had my messages:
```jsx
const WrappedDEConfirmationDialog = withI18N(DEConfirmationDialog, intlData);
```
Then in my component's `render` method:
```jsx
                <WrappedDEConfirmationDialog dialogOpen={showImportConfirmationDialog}
                                             heading={formatMessage(intl, "importIRODSMetadata")}
                                             message={formatHTMLMessage("importIRODSMetadataMsg")}
                                             ...
                />
```

Which method should we prefer going forward? Passing `messages={intlData.messages}` seems simple enough, but then again, I think you'd have to know what props `withI18N` is injecting (and what `getMessage` expects to find). I haven't reviewed their API docs in a while, but are we expected to know they inject a `messages` prop?

Would it be safer to use a wrapped component as I did above, in case these injected props change for some reason in a future version?

ash3rz: This is neat!  I was looking for a way to do this without having to manage the `messages` prop too actually.  This seems to work with messages from parent -> child, but I can't figure out how to send a message from parent (EditCommunity) -> child (DEConfirmationDialog) -> sub-child (DEDialogHeader).  Right now the text from EditCommunity's messages for the title for the dialog header can't be found with this approach.  Maybe there's a different approach to take in this particular scenario or something I'm missing?

psarando: Odd, because that's the same setup I'm using in my metadata components. I've even changed my example above from `heading={formatMessage(intl, "importIRODSMetadata")}` to `heading={getMessage("importIRODSMetadata")}`.

In other words, my metadata component looks something like this:
```jsx
import intlData from "./messages";
// ...
const ConfirmationDialog = withI18N(DEConfirmationDialog, intlData);

class EditMetadata extends Component {
// ...
    render() {
// ...
        return (
            <Dialog open={open}>
// ...
                <ConfirmationDialog dialogOpen={showImportConfirmationDialog}
                                    debugId={ids.EDIT_METADATA_FORM}
                                    heading={getMessage("importIRODSMetadata")}
                                    message={formatHTMLMessage("importIRODSMetadataMsg")}
                                    onOkBtnClick={...}
                                    onCancelBtnClick={...}
                />
            </Dialog>
        );
    }
}
// ...
export default withFormik(
    {
        mapPropsToValues: ({ metadata }) => ({...metadata}),
        validate,
        handleSubmit,
    }
)(withStyles(styles)(withI18N(injectIntl(EditMetadata), intlData)));
```

Also note that the `intlData.messages` that are injected into the wrapped `DEConfirmationDialog` has both the confirmation dialog body and header messages.

Is the message for the `DEDialogHeader` in the `EditCommunity`'s messages?  And those same messages are passed to the wrapped `withI18N(DEConfirmationDialog, messages)` in `EditCommunity`?

ash3rz: Do you have any additional changes to `DEConfirmationDialog`?

I think I have mine set up like yours:
```
import messages from "../messages";
...
                    <ConfirmationDialog dialogOpen={this.state.deleteCommunity}
                                        ...
                                        heading={getMessage('confirmDeleteCommunityTitle', {values: {name: communityName}})}
                                        message={
                                            <div>
                                                {loading &&
                                                <CircularProgress size={30}
                                                                  classes={{root: classes.loading}}
                                                                  thickness={7}/>
                                                }
                                                {getMessage('confirmDeleteCommunity')}
                                            </div>
                                        }/>
...
const ConfirmationDialog = withI18N(DEConfirmationDialog, messages);
export default withStyles(styles)(withI18N(EditCommunity, messages));
```

I see you're also using `injectIntl` and I'm not, but I don't think that would have an effect here?

`DEConfirmationDialog` has all my messages:
image

But `DEDialogHeader` only has its own messages for me:
image

ash3rz: Oh, actually I'm the one with the changes to `DEConfirmationDialog`!  

I removed the `messages={messages}` bit, that's in `master` right now.  Without that line, the title text won't be found.  Does it still work for you in that case?

psarando: Oh, I missed that! 😩 

When I remove `messages={messages}` from the `DEDialogHeader` tag in `DEConfirmationDialog`, then it no longer works for me either.

I guess in this 1 spot we could continue to pass down the `messages` prop. It's too bad the API isn't a little more straight forward for cases like this 🤔

ash3rz: Yeah, I've been trying to dig into their docs a little more. They do have a nested example, but when I try to run it locally to mess around with it, it doesn't build for me 🤦‍♀️ I'm not 100% sure, but I get the feeling we're not using it the way they intended. I only looked briefly at the docs this morning and then moved on to some other things though, but I can take a look again maybe tomorrow.

PR #390, opened 2019-03-04
CORE-126 Add static IDs for analysis name icons

ash3rz:


ash3rz:
Since this is small, I'll go ahead and merge.  Reviews still welcome though.

PR #391, opened 2019-03-05
Fix annoying typography variant deprecated warning.

sri2k1us:
See https://material-ui.com/style/typography/#migration-to-typography-v2

P.S: I may have missed few spots. If so, please update them in your next PR. 🙏

### DIFF
--- a/react-components/src/lib.js
+++ b/react-components/src/lib.js
@@ -32,7 +32,9 @@ const getDefaultTheme = () => createMuiTheme({
     typography: {
         button: {
             textTransform: "none",
-        }
+        },
+        useNextVariants: true,
+
     },
     zIndex: {
         mobileStepper: 1000 + adjustZIndexForGXT,
--- a/react-components/src/analysis/view/AnalysesView.js
+++ b/react-components/src/analysis/view/AnalysesView.js
@@ -1006,7 +1006,7 @@ class AnalysesView extends Component {
                             !hasData &&
                             <Typography style={{margin: '0, auto, 0, auto', width: 600}}
                                         align="center"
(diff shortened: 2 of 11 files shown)


react-components/src/lib.js
```diff
@@ -32,7 +32,9 @@ const getDefaultTheme = () => createMuiTheme({
     typography: {
         button: {
             textTransform: "none",
-        }
+        },
+        useNextVariants: true,
```
psarando: I'm wondering if we should hold off on these updates.

By just setting `useNextVariants: true`, the delete icons in the `EditMetadataTemplate` expansion panel headers go from looking like this
screen shot 2019-02-27 at 12 54 10 pm
to looking like this
screen shot 2019-03-05 at 4 52 06 pm
This happens in both Chrome and Firefox, and it doesn't matter which `Typography` variants are used in `EditMetadataTemplate`.

sri2k1us: Weird. Looks Ok for me...
delete-btn

Google chrome - Version 72.0.3626.121 (Official Build) (64-bit)

sri2k1us: It does look like your screenshot in storybook 😕

sri2k1us: delete-with-margin


Fixed it with margin. Looks good in Belphe as well.

psarando: I think the reason it looked OK in your 1st screenshot is there is only an attribute name header, but no attribute description subheading.

Adding the margin appears to fix it, though.

Nice find! 👍

sri2k1us:
Thanks for reviews.

PR #392, opened 2019-03-06
Upgrade to SB version 5.

sri2k1us:


react-components/package.json
```diff
@@ -7,27 +7,29 @@
...
-    "@storybook/addon-actions": "^4.1.7",
+    "@storybook/addon-actions": "5.0.0",
     "@storybook/addon-console": "^1.1.0",
-    "@storybook/addon-links": "^4.1.7",
-    "@storybook/addons": "^4.1.7",
-    "@storybook/react": "^4.1.7",
+    "@storybook/addon-links": "5.0.0",
+    "@storybook/addon-knobs": "^5.0.0",
+    "@storybook/addons": "5.0.0",
+    "@storybook/react": "5.0.0",
```
psarando: Rather than an exact version `5.0.0` for these `storybook` deps, shouldn't we use the "compatible version" syntax: `^5.0.0`? Same as `@storybook/addon-knobs` here.

sri2k1us: Yup good catch!. Oversight! let me fix it.

sri2k1us:
Thanks for reviews!

PR #393, opened 2019-03-06
core-100 Add default input option for folder and file input fields.

sri2k1us:


de-lib/src/main/java/org/iplantc/de/resources/client/uiapps/widgets/AppsWidgetsContextualHelpMessages.properties
```diff
@@ -47,4 +47,9 @@ infoLabelHelp = This field accepts the following html tags: &lt;p&gt;, &lt;br&gt
 
 fileInputIsImplicit = <p>Selecting this option allows the app user to select input files without those files being included in the command line.</p>
 fileInputRepeatOptionFlag = <p>Selecting this option allows the "Argument option" defined for this parameter to be included on the command line before each filename selected by the user.</p>
-multiFilePrompt = Select multiple input files. Tip: You can also drag and drop files from the Data window.
\ No newline at end of file
+multiFilePrompt = Select multiple input files. Tip: You can also drag and drop files from the Data window.
+fileInputDefaultText = <p>The app will use this value by default. App will need to be updated if this file moved, deleted or renamed!</p>
```
ash3rz: Maybe some slight changes:
```
The app will use this value by default. Note that the app will need to be updated if this file is moved, deleted, or renamed.
```

de-lib/src/main/java/org/iplantc/de/apps/widgets/client/view/editors/arguments/converters/SplittableToFileConverter.java
```diff
@@ -31,10 +32,11 @@ public File convertModelValue(Splittable object) {
         if (object == null) {
             return null;
           }
-        if(object.isString()){
+        if (object != null && object.get("path") != null && !Strings.isNullOrEmpty(object.get("path")
```
psarando: We don't have to check `object != null` here, because there is a `null` check right above this. It might also be slightly more readable to refactor `object.get("path").asString()` into a variable, which could also be reused in the `file.setPath` below:
```java
        String path = object.get("path") != null ? object.get("path").asString() : "";
        if (!Strings.isNullOrEmpty(path)) {
            File file = factory.file().as();
            file.setPath(path);
            return file;
        }
        return null;
```

de-lib/src/main/java/org/iplantc/de/resources/client/uiapps/widgets/AppsWidgetsContextualHelpMessages.properties
```diff
@@ -47,4 +47,9 @@ infoLabelHelp = This field accepts the following html tags: &lt;p&gt;, &lt;br&gt
...
+folderInputDefaultText = <p>The app will use this value by default. App will need to be updated if this folder moved, deleted or renamed!</p>
```
psarando: Just copying @ash3rz 's suggestion here too:
```suggestion
folderInputDefaultText = <p>The app will use this value by default. Note that the app will need to be updated if this folder is moved, deleted, or renamed.</p>
```

de-lib/src/main/java/org/iplantc/de/apps/widgets/client/view/editors/arguments/converters/SplittableToFileConverter.java
sri2k1us: 👍 . Actually i just realized i missed this fix for `SplittableToFolderConverter`. Will add that too.

sri2k1us:
Thanks for the reviews.

PR #394, opened 2019-03-07
CORE-103: Allow DE metadata dialogs to attach AVUs to other AVUs

psarando:
This PR will replace the GXT view for the Edit Metadata dialog with a React view that supports nested metadata:

DE Edit Metadata Screenshot_2019-03-19 Discovery Environment

Note that hovering over the `Delete` icon in the listing grid will highlight the icon a red color: 
screen shot 2019-02-27 at 12 54 10 pm

After the dialog title are the `View in Template`, `Save to File`, `?` (Help), and `X` (Close) icons:
de edit metadata - view in template screen shot 2019-03-07 at 1 45 05 pm de metadata - save to file screen shot 2019-03-07 at 1 47 49 pm

Selecting the `?` (Help) icon will display a help-text popup:
DE Metadata Help popup Screenshot_2019-03-19 Discovery Environment

The `Save` button in the bottom-right of the dialog will only become active if the metadata has been updated in some way, which will also disable the `Save to File` icon. The `Save to File` icon will be disabled since the service endpoint that exports the metadata to a file will only use metadata that is saved in the database. So disabling this `Save to File` icon is new in this PR, since I feel it's misleading to the user if they make some changes to the metadata, then decide to export it without saving the changes first.

---

Selecting the `X` icon or `Close` button when there are metadata changes will display a confirmation dialog:
de metadata discard changes confirmation screenshot_2019-03-07 discovery environment

---

The form for editing an AVU and its children:
de edit nested metadata screenshot_2019-03-07 discovery environment

---

The form for editing a child (nested) AVU:
de edit leaf metadata screenshot_2019-03-07 discovery environment

---

If there are also iRODS metadata, then the tabs `User Metadata` and `Additional Metadata` will appear under the dialog's header (otherwise the tabs are hidden if there are no iRODS metadata). The `View in Template` icon is replaced with an `Import` icon when clicking on the `Additional Metadata` tab (unless the data is read-only). The `Import` icon is only active if at least 1 row of iRODS metadata is selected:
DE Import iRODS Metadata tooltip Screen Shot 2019-03-19 at 5 19 02 PM

Note that iRODS metadata is read-only, so the `Edit` and `Delete` icons in the listing grid are replaced with a single `View` button.

---

Selecting the `Import` icon will display a confirmation dialog before attempting to import the metadata:
de import irods metadata confirmation screen shot 2019-03-07

---

For read-only files/folders (such as public data), the Metadata dialog has a read-only `View` mode (there are no iRODS metadata in the following example):
DE read-only Metadata Screenshot_2019-03-19 Discovery Environment

Note that the `Save` button is hidden, and just as in the iRODS listing grid, the `Edit` and `Delete` icons in the listing grid are replaced with a single `View` button.

---

The form for viewing a read-only AVU and its children:
DE read-only Nested Metadata Screenshot_2019-03-19 Discovery Environment

---

The form for viewing a child (nested) AVU (without any child AVUs under it):
DE read-only Leaf Metadata Screenshot_2019-03-19 Discovery Environment

---

Finally, the view for a read-only file/folder with iRODS metadata:
DE read-only iRODS Metadata Screenshot_2019-03-19 Discovery Environment


react-components/src/util/dialog/DEConfirmationDialog.js
```diff
@@ -27,33 +27,33 @@ class DEConfirmationDialog extends Component {
...
-                    }}/>
+            <Dialog id={debugId}
+                    open={dialogOpen}
+                    aria-labelledby={dialogTitleID}
+            >
+                <DEDialogHeader id={dialogTitleID}
+                                messages={messages}
+                                heading={heading}
+                                onClose={onCancelBtnClick}
+                />
```
sri2k1us: `onCancelBtnClick` is not a required prop. Should we add back the check to see if the method is `undefined`? Actually let's make it required.

psarando: We don't have to check for `null`/`undefined`, since this is being passed directly as the `onClick` prop to `IconButton`, and it handles `null`/`undefined` for its `onClick` just fine.

Since `onClose` is a required prop for `DEDialogHeader`, then I agree that it should be required for `DEConfirmationDialog` as well, so I'll update that in this PR.

Let's also consider updating `DEConfirmationDialog`, in a future PR, to make the `OK` or `Cancel` button optional, so we can show a dialog with only 1 "confirm" or "cancel" button (we may have already discussed this before in elsewhere 🤔 ).

psarando:
> However there is one sticking point I would like to bring up is the placement of `close` button on the left side rather than the usual right side for the DE.

Oh right, I was originally modelling these dialogs off the full-screen dialog example, but these dialogs are no longer full-screen. I like the idea of having the cancel/save buttons at the bottom-right, as we do in all other normal-size-dialogs. So I can add those bottom-right buttons and move the close button to the top-right in this dialog, and in the metadata template dialogs as well.

react-components/src/util/dialog/DEConfirmationDialog.js
sri2k1us: Weird...I do get 
`analysis/view/with test analyses/error: ["Uncaught TypeError: onCancelBtnClick is not a function", "http://localhost:6006/vendors~main.369c8f96462db18bc2f3.bundle.js", 107711, 5, TypeError]
analysis/view/with test analyses/error: ["Uncaught TypeError: onCancelBtnClick is not a function", "http://localhost:6006/main.369c8f96462db18bc2f3.hot-update.js", 116, 11, TypeError]` when I don't pass the prop in SB. 🤔 

Anyway I agree this should be a required prop. 

WRT `DEConfirmationDialog` we agreed to make `OK` and `Cancel` text as prop. But yes we can make just one button mandatory allowing us to display the dialog with 1 button.

react-components/src/metadata/MetadataList.js
```diff
@@ -0,0 +1,295 @@
...
+        };
+
+        [
+            "onAddAVU",
+            "handleRequestSort",
+        ].forEach(methodName => (this[methodName] = this[methodName].bind(this)));
+    }
+
+    newAttrName() {
+        return `New attribute ${this.newAttrCount++}`;
```
ash3rz: Could this be i18n-ized?

react-components/src/metadata/MetadataList.js
```diff
@@ -0,0 +1,295 @@
...
+                aVal = aVal ? aVal.length : 0;
+                bVal = bVal ? bVal.length : 0;
+            }
+
+            return order === 'desc' ? comparator(bVal, aVal) : comparator(aVal, bVal);
+        });
+
+        this.setState({ order, orderBy });
+
+        setFieldValue(name, avus);
```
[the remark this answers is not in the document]
ash3rz: I added some [helper methods](https://github.com/cyverse-de/ui/blob/master/react-components/src/util/table/TableSort.js) for sorting, not sure if that would be helpful in this case or not

react-components/src/util/dialog/DEConfirmationDialog.js
```diff
@@ -63,15 +63,21 @@ class DEConfirmationDialog extends Component {
...
-    message: PropTypes.string.isRequired,
-    heading: PropTypes.string.isRequired,
+    message: PropTypes.oneOfType([
+        PropTypes.string,
+        PropTypes.object,
+    ]).isRequired,
+    heading: PropTypes.oneOfType([
+        PropTypes.string,
+        PropTypes.object,
+    ]).isRequired,
```
ash3rz: 👍

react-components/src/metadata/MetadataList.js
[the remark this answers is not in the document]
psarando: Thanks for the reminder 👍 
I had written this initially (with redux-form) before you pushed those util methods, but I forgot to make use of them after I rebased this PR on top of those changes.

I might timebox sorting in the `render` method with your util functions, and maybe I can get that updated before I merge. At the very least I can replace my sorting logic with your stable sort.

react-components/src/metadata/MetadataList.js
psarando: Good catch. Looks like I already created an `intl` message for use in `EditAttributeFormList`, but I forgot to come back to this component and add it here too 😆 :
https://github.com/cyverse-de/ui/blob/7bc4e878f27fd90295a0ef0aad5f88aee3aae52e/react-components/src/metadata/admin/EditAttributeFormList.js#L288-L290

sri2k1us:
Sorting changes LGTM.

psarando:
Thanks for the reviews! 👍 
I'll go ahead and merge this now that the prod hotfix is deployed.

PR #395, opened 2019-03-13
Communities hotfix

ash3rz:
This will:
* In the Apps window, deselect communities after the user has submitted a search
* In the Communities window, fix issues regarding searching for apps to add to a community

### DIFF
--- a/react-components/src/communities/view/EditCommunityDialog.js
+++ b/react-components/src/communities/view/EditCommunityDialog.js
@@ -24,13 +24,15 @@ class EditCommunityDialog extends Component {
         super(props);
 
         this.state = {
-            saveCommunity: false
+            saveCommunity: false,
+            isSelectAppsDlgOpen: false,
         };
 
         [
             'onDialogBtnClicked',
             'handleCloseEditDialog',
             'cancelSave',
+            'isSelectAppsDlgOpen'
         ].forEach((fn) => this[fn] = this[fn].bind(this));
     }
 
@@ -52,6 +54,10 @@ class EditCommunityDialog extends Component {
         });
     }
 
+    isSelectAppsDlgOpen(open) {
+        this.setState({isSelectAppsDlgOpen: open});
+    }
+
     render() {
         const {
             community,
@@ -63,10 +69,12 @@ class EditCommunityDialog extends Component {
             presenter,
             currentUser,
             onCommunitySaved,
+            classes,
         } = this.props;
 
         const {
-            saveCommunity
+            saveCommunity,
+            isSelectAppsDlgOpen,
         } = this.state;
 
         const communityName = community !== null ? collaboratorsUtil.getSubjectDisplayName(community) : null;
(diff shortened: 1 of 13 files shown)


react-components/src/communities/view/EditCommunityDialog.js
```diff
@@ -76,6 +84,7 @@ class EditCommunityDialog extends Component {
                     fullWidth={true}
                     maxWidth='lg'
                     onClose={onClose}
+                    classes={{root: isSelectAppsDlgOpen ? classes.hidden : null}}
```
psarando: In my metadata dialog PR (#394), I set the dialog's `open` prop to `false` to hide it temporarily, and when it needs to come back, resetting its `open` prop to `true` shows it again, but it retains its state while it's hidden: https://github.com/cyverse-de/ui/pull/394/files#diff-80f7a61599153f473b4b30c76a63f831R74

This also has the advantage that it will also use its slide up/down animation when hiding and re-displaying. I'm not sure if a transition animation is used when setting this CSS `hidden` attribute.

sri2k1us: Actually I remember now, I did the same thing in `AnalysesView` as well - https://github.com/cyverse-de/ui/blob/8fb3917c94916dd2c5c3e3767acc41bf9e05f473/react-components/src/analysis/view/AnalysesView.js#L561

ash3rz: It doesn't seem to be working for me, probably because I have state inside `EditCommunity` and that seems to automatically get cleared when `EditCommunityDialog` isn't open.  The constructor gets called again when the dialog reopens.  I could take this opportunity to refactor the state up the hierarchy if you guys don't mind waiting for the hotfix.

psarando: That's unfortunate 😩 I noticed the `EditCommunityDialog` uses the `EditCommunity` component as its dialog contents, which is actually recommended for performance:
https://material-ui.com/demos/dialogs/#performance

I was also planning on refactoring the metadata dialogs this way to see if that will help the form's performance, but if hiding the dialog also removes its children (resetting their state), then that could make the refactoring a trickier task.

Pulling state up to the dialog does seem like it might help, but I'm not sure if that makes sense (e.g. maybe `EditCommunity` should be able to display in something besides a dialog). It also might be too much work for this hotfix, so I'm fine if we :shipit: as is 👍

ash3rz: I know 😩 There is a "fix" (https://github.com/mui-org/material-ui/issues/10572), and I tried that, but then it keeps those children in the DOM forever.  So in my case, if I opened the dialog, closed it, closed the Communities window too, and then reopened everything, I could add an infinite number of dialogs to the DOM that just stay there forever (even though I don't see them) and create duplicate static IDs.

ash3rz:
Thanks for the reviews!

PR #397, opened 2019-03-16
CORE-123 and 124 - Update DE's error page with log out btn and error details

ash3rz:
With beautification help from Mariah ✨ 

**Before**
image

**After**
image

**After, expanded**
image

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/DesktopPresenterImpl.java
+++ b/de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/DesktopPresenterImpl.java
@@ -23,6 +23,7 @@ import org.iplantc.de.client.services.UserSessionServiceFacade;
 import org.iplantc.de.client.services.callbacks.ReactErrorCallback;
 import org.iplantc.de.client.util.CommonModelUtils;
 import org.iplantc.de.client.util.DiskResourceUtil;
+import org.iplantc.de.client.util.JsonUtil;
 import org.iplantc.de.client.util.WebStorageUtil;
 import org.iplantc.de.commons.client.CommonUiConstants;
 import org.iplantc.de.commons.client.ErrorHandler;
@@ -97,6 +98,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 
 /**
  * @author jstroot
@@ -154,6 +156,7 @@ public class DesktopPresenterImpl implements DesktopView.Presenter,
     @Inject UserSettings userSettings;
     @Inject NotifyInfo notifyInfo;
     @Inject DiskResourceUtil diskResourceUtil;
+    @Inject JsonUtil jsonUtil;
     @Inject NotificationUtil notificationUtil;
     @Inject AsyncProviderWrapper<PreferencesDialog> preferencesDialogProvider;
     private DesktopPresenterAppearance appearance;
@@ -281,10 +284,17 @@ public class DesktopPresenterImpl implements DesktopView.Presenter,
         }
     }
 
-    public void onBootstrapError(Integer statusCode) {
+    public void onBootstrapError(Integer statusCode, String errorMessage) {
         String redirectUrl = GWT.getHostPageBaseURL() + deClientConstants.errorUrl();
         if (statusCode != null) {
             redirectUrl += "-" + statusCode.toString();
+        }
+        if (errorMessage != null) {
+            JSONObject json = jsonUtil.getObject(errorMessage);
+            String errorString = json.keySet().stream()
+                                     .map(key -> URL.encode(key) + "=" + URL.encode(json.get(key).toString()))
+                                     .collect(Collectors.joining("&"));
+            redirectUrl += "?" + URL.encode(errorString);
         }
         Window.Location.assign(redirectUrl);
     }
--- a/de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/InitializationCallbacks.java
+++ b/de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/InitializationCallbacks.java
@@ -68,7 +68,7 @@ class InitializationCallbacks {
             if (caught instanceof HttpException) {
                 statusCode = ((HttpException)caught).getStatusCode();
             }
-            presenter.onBootstrapError(statusCode);
+            presenter.onBootstrapError(statusCode, caught.getMessage());
         }
 
         @Override
@@ -76,7 +76,7 @@ class InitializationCallbacks {
             userInfo.init(result);
 
             if (userInfo.hasUserProfileError()) {
-                presenter.onBootstrapError(null);
+                presenter.onBootstrapError(null, "{\"reason\":\"User profile contains errors\"}");
                 return;
             }
 
@@ -196,7 +196,7 @@ class InitializationCallbacks {
             if (caught instanceof HttpException) {
                 statusCode = ((HttpException)caught).getStatusCode();
             }
-            presenter.onBootstrapError(statusCode);
+            presenter.onBootstrapError(statusCode, caught.getMessage());
         }
 
         @Override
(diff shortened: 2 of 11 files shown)


de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/DesktopPresenterImpl.java
```diff
@@ -281,11 +284,18 @@ public void doLogout(boolean sessionTimeout) {
...
         if (statusCode != null) {
             redirectUrl += "-" + statusCode.toString();
         }
+        if (errorMessage != null) {
+            JSONObject json = jsonUtil.getObject(errorMessage);
+            String errorString = json.keySet().stream()
+                                     .map(key -> URL.encode(key) + "=" + URL.encode(json.get(key).toString()))
+                                     .collect(Collectors.joining("&"));
+            redirectUrl += "?" + URL.encode(errorString);
+        }
```
sri2k1us: ✨

ash3rz: That's a really good question. I was trying to maintain some feedback I got from Kathleen when we first put the error page together about how "ugly" and confusing it is to see error details and that they should be hidden unless the user wants to see them.  At least back then, it was kind of in line with the rest of the DE because we were leaning more heavily on friendly error messages in the `IplantAnnouncer` instead of displaying the raw error strings in `ErrorHandler`, but we've been utilizing the `ErrorHandler` more recently.  I could go either way - I like the cleaner look, but maybe it's more efficient to have the details right there?

sri2k1us: The reason why i asked this question is that we are recommending users to include the details when they contact the support. If we want to encourage users to include them, we should try to remove all additional steps (like clicking on the button to view details).

psarando: On the other hand, if the error details are always showing, then I'd be worried that users would never read the last part of the error page, about checking our maintenance page, etc.

The information is segregated nicely with the error details inside the expansion panel. Maybe we could at least have the expansion panel expanded by default, so the user can collapse it if they want to only read the user-friendly messages below it?

sri2k1us: I am ok with having them expanded by default. Speaking of maintenance, we should link our public maintenance calendar in this page - https://www.cyverse.org/maintenance-calendar

de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/InitializationCallbacks.java
```diff
@@ -68,15 +68,15 @@ public void onFailure(Throwable caught) {
...
+            presenter.onBootstrapError(statusCode, caught.getMessage());
         }
 
         @Override
         public void onSuccess(String result) {
             userInfo.init(result);
 
             if (userInfo.hasUserProfileError()) {
-                presenter.onBootstrapError(null);
+                presenter.onBootstrapError(null, "{\"reason\":\"User profile contains errors\"}");
```
psarando: 👍

de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/DesktopPresenterImpl.java
ash3rz: OK, should I reword the bottom blurb to mention the maintenance calendar separately or maybe leave it the same but make "...as well as `scheduled maintenance` events...` become a link to the calendar?

sri2k1us: I am ok with `scheduled maintenance` as link

de-webapp/src/main/webapp/WEB-INF/jsp/bootstrapError.jsp
```diff
@@ -173,10 +193,38 @@ a:hover {
...
+    if (sessionStorage.getItem("is_reloaded")) {
+        window.location.replace('${login_url}');
+        sessionStorage.removeItem("is_reloaded");
+    } else {
+        sessionStorage.setItem("is_reloaded", 1);
+    }
+
+    var acc = document.getElementById("errorBtn");
+
+    acc.addEventListener("click", function() {
```
psarando: This looks good for this PR, but I just want to point out that we should be able to use React components here, if we use the following JSP include that our main `de.jsp` page uses:
https://github.com/cyverse-de/ui/blob/7bc4e878f27fd90295a0ef0aad5f88aee3aae52e/de-webapp/src/main/webapp/WEB-INF/jsp/de.jsp#L68

Then we could use material-ui components (and our custom CyVerse theme) and display a custom component with the error message here.

For example:
```javascript
        ReactDOM.render(
            React.createElement(
                CyVerseReactComponents.util.ErrorExpansionPanel,
                {
                    errMsg: "${error_message}",
                    username: "${username}",
                    userAgent: "${user_agent}",
                    date: "${date}",
                    host: "${request_url}"
                }
            ),
            document.getElementById(acc)
        );
```
By the way, since our `CyVerseReactComponents` javascript library is a global object on the page, it can be examined from the browser's console when the DE is loaded.

ash3rz: Oh, neat - I'll give that a go!  I was thinking material would be better, but wasn't sure how to put it directly into the JSP.

de-webapp/src/main/webapp/WEB-INF/jsp/bootstrapError.jsp
```diff
@@ -1,4 +1,5 @@
 <%@ taglib prefix="c" uri="http://java.sun.com/jsp/jstl/core" %>
+<%@ include file="react_include.jsp" %>
```
psarando: This `react_include.jsp` file contains `<link>` and `<script>` tags that are intended to go inside the `<header>` tag of the HTML doc.

Notice the placement of this `include` in the `de.jsp` page is after most other `<script>` tags:
https://github.com/cyverse-de/ui/blob/a30f4c9914ee589a54731ce503a97b857bd7e10a/de-webapp/src/main/webapp/WEB-INF/jsp/de.jsp#L65-L68
Also in `belphegor.jsp`:
https://github.com/cyverse-de/ui/blob/295a657052f10fbcb6400c9e8f02151ff7201828/de-webapp/src/main/webapp/WEB-INF/jsp/belphegor.jsp#L10-L18

react-components/stories/util/ErrorExpansionPanel.stories.js
```diff
@@ -0,0 +1,23 @@
+import ErrorExpansionPanel from '../../src/util/ErrorExpansionPanel';
+
+import React, { Component } from 'react';
+
+class ErrorExpansionPanelTest extends Component {
+    render() {
+        const errMsg = "Error: ERR_ILLEGAL_ARGUMENT<br>Real Error: Everything is broken forever, run!";
```
psarando: 😆

de-webapp/src/main/webapp/WEB-INF/jsp/bootstrapError.jsp
ash3rz: I moved where I'm adding the include, but I also had to then copy over another line from the `de.jsp` and `belphegor.jsp` files:
```
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN" "http://www.w3.org/TR/html4/strict.dtd">
```
Without that, I would get errors along the lines of "...not a supported stylesheet MIME type, and strict MIME checking is enabled."

de-webapp/src/main/webapp/WEB-INF/jsp/bootstrapError.jsp
```diff
@@ -1,6 +1,5 @@
 <%@ taglib prefix="c" uri="http://java.sun.com/jsp/jstl/core" %>
-<%@ include file="react_include.jsp" %>
-<!DOCTYPE html>
+<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN" "http://www.w3.org/TR/html4/strict.dtd">
```
psarando: 👍

ash3rz:
Thanks for the reviews!

PR #399, opened 2019-03-28
Update react-scripts. Add prettier.

sri2k1us:


sri2k1us:
Should we run `npm run prettier` on our `master` branch once so that we all have a formatted code to start with ?

PR #406, opened 2019-04-11
core-188 Fix relaunch for apps with multi-input.

sri2k1us:


sri2k1us:
Thanks for the reviews.

PR #408, opened 2019-04-22
CORE-234 Fixes for Metadata Template checkboxes

psarando:
This PR will add a `FormCheckboxStringValue` component that saves and parses existing metadata
values as a "true" or "false" string, since metadata AVU values must be strings, and the service will throw the validation error `ERR_ILLEGAL_ARGUMENT` for non-string values set in the AVU's `value` field.

We have existing metadata AVUs with "true" or "false" values for some template attributes with `Boolean` types, since the previous GXT component saved these attribute values as "true" or "false".

Also included a fix for read-only checkboxes in templates.


psarando:
Thanks for the reviews! 👍

PR #409, opened 2019-04-24
CORE-242 Only provide name and desc in group updates

ash3rz:
This is what the schema in terrain is expecting now


ash3rz:
Thanks for the review!

PR #411, opened 2019-04-26
Quick launch

sri2k1us:
- Add ability to create quick launch from App launch window.
- List quick launches in App Details dialog
- Add ability use, embed, share via link for quick launch.
- Update packages.

image

image


react-components/src/apps/quickLaunch/CreateQuickLaunchDialog.js
```diff
@@ -0,0 +1,107 @@
...
+    const { dialogOpen, appName, presenter, intl } = props;
+
+    const handleSubmit = (values, actions) => {
+        console.log("submit clicked->" + values.name);
+        actions.setSubmitting(true);
+        const { name, description, is_public } = values;
+        presenter.createQuickLaunch(
+            name,
+            description ? description : "",
+            is_public ? is_public : false,
```
ash3rz: I'm not familiar with Formik yet, but are the `description` and `is_public` checks necessary?  For example, if `is_public` is tied to a CheckBox, I would think it would automatically be either true or false.

de-lib/src/main/java/org/iplantc/de/client/util/AnalysisSubmissionUtil.java
```diff
@@ -0,0 +1,105 @@
...
+import com.google.common.base.Strings;
+import com.google.web.bindery.autobean.shared.AutoBeanCodex;
+import com.google.web.bindery.autobean.shared.AutoBeanUtils;
+import com.google.web.bindery.autobean.shared.Splittable;
+import com.google.web.bindery.autobean.shared.impl.StringQuoter;
+
+
+public class AnalysisSubmissionUtil {
+
+    public static Splittable assembleLaunchAnalysisPayload(final AppTemplateUtils appTemplateUtils,
```
ash3rz: 🌟 Nice refactor!  I think eventually we might want to create models/autobeans for this stuff, but probably not for this PR

react-components/src/util/dialog/DEAlertDialog.js
```diff
@@ -0,0 +1,49 @@
...
+    const [dialogOpen, heading, alertMessage, handleClose] = props;
+    return (
+        <Dialog
+            open={dialogOpen}
+            onClose={handleClose}
+            aria-describedby="alert-dialog-description"
+        >
+            <DEDialogHeader heading={heading} onClose={handleClose} />
+            <DialogContent>
+                <DialogContentText id="alert-dialog-description">
```
ash3rz: Is this id meant to be a static ID?  If so, I would imagine we could have multiple alert dialogs open, in which case, this ID wouldn't be unique and could cause issues.

ash3rz:
Oh also, I'm not sure if `prettier` has been run?

de-lib/src/main/java/org/iplantc/de/apps/widgets/client/view/AppLaunchPreviewViewImpl.java
```diff
@@ -63,6 +72,12 @@ public void edit(AppTemplate appTemplate, JobExecution je) {
         editorDriver.edit(appTemplate);
     }
 
+    @Override
+    public void showOrHideCreateQuickLaunchView(ReactQuickLaunch.CreateQLProps props) {
+        throw new UnsupportedOperationException(
+                "App Launch preview does not support  Create Quick " + "Launch view");
```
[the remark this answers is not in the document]
sri2k1us: Yes. I don't remember breaking the string like this. May be some weird setting in intellij ? 🤔

de-lib/src/main/java/org/iplantc/de/client/util/AnalysisSubmissionUtil.java
sri2k1us: I didn't create autobeans for this stuff because the splittable created here is going to be sent as body to an API call (and i didn't want to write more GWT code )

react-components/src/apps/details/AppInfoDialog.js
```diff
@@ -115,18 +136,52 @@ function AppInfoDialog(props) {
...
+        onClose();
+    };
+
+    const doDeleteQuickLaunch = (qLaunch) => {
+        setLoading(true);
+        presenter.deleteQuickLaunch(
+            qLaunch.id,
+            () => {
+                let pos = quickLaunches.indexOf(qLaunch);
+                if (pos !== -1) {
```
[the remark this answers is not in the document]
sri2k1us: I wrote the check inside the callback because `doDeleteQuickLaunch` cannot be invoked without a quick launch in the first place. So i am making sure the quick launch is still there after the remote call is executed.

react-components/src/apps/quickLaunch/CreateQuickLaunchDialog.js
```diff
@@ -0,0 +1,107 @@
...
+} from "../../util/FormField";
+import Button from "@material-ui/core/Button";
+import DialogActions from "@material-ui/core/DialogActions";
+import { Tooltip } from "@material-ui/core";
+
+function CreateQuickLaunchDialog(props) {
+    const { dialogOpen, appName, presenter, intl } = props;
+
+    const handleSubmit = (values, actions) => {
+        console.log("submit clicked->" + values.name);
```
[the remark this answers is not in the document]
sri2k1us: nope. I will remove them.

react-components/src/apps/quickLaunch/CreateQuickLaunchDialog.js
sri2k1us: As far as my understanding, if a field is not filled by user, that field will not show up under `value` object. So we have to manually set those values to "" or false depending on the type of field. For example, if the user didn't check `Public` checkbox, then `values.is_public` will be `undefined` but i can't send that to java so I have to manually set it to `false`. May be @psarando has a different idea 🤔

react-components/src/util/dialog/DEAlertDialog.js
sri2k1us: There will be one alert dialog at a time because this is Modal but I agree, that `id` needs to be more unique. I will add a `parent-debug-id` prop from which a unique id can built.

react-components/src/apps/details/AppInfoDialog.js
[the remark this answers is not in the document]
sri2k1us: > I had a few minor comments that aren't necessary for this PR, but I don't think I saw many static IDs for testing. Also, I think a few classes were missing prop-types. Other than that, LGTM 👍

I totally forgot about `ids`. Will add them 👍

sri2k1us:
> Oh also, I'm not sure if `prettier` has been run?

It should be. I ran it again just to double check, there was no change except for a small formatting change in `index.stories.js`

react-components/src/apps/quickLaunch/CreateQuickLaunchDialog.js
sri2k1us: Ok after reading formik docs, it looks like, if i set initial values for `description` and `is_public`, then i don't have to set them here.

sri2k1us:
Also @psarando @ash3rz  - When you get a chance, can you checkout this branch and try build and run npm test to see if it succeeds ?

react-components/src/apps/quickLaunch/CreateQuickLaunchDialog.js
```diff
@@ -68,18 +64,6 @@ function CreateQuickLaunchDialog(props) {
...
-                                    baseDebugId,
-                                    ids.QUICK_LAUNCH.description
-                                )}
-                                name="description"
-                                label={getMessage("descriptionLabel")}
-                                required={false}
-                                margin="dense"
-                                component={FormMultilineTextField}
-                            />
-                            <ErrorMessage name="description" component="div" />
```
sri2k1us: Currently, there is no UI to display the `description` associated with a `Quick Launch`. Removing `description` field for the first pass of `Quick Launch`. I will add them back when we start displaying stats for `Quick Launch`.

react-components/src/apps/quickLaunch/CreateQuickLaunchDialog.js
```diff
@@ -0,0 +1,110 @@
...
+                        <DialogContent>
+                            <Field
+                                id={build(baseDebugId, ids.QUICK_LAUNCH.name)}
+                                name="name"
+                                label={getMessage("quickLaunchNameLabel")}
+                                required={true}
+                                margin="dense"
+                                component={FormTextField}
+                            />
+                            <ErrorMessage name="name" component="div" />
```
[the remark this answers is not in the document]
sri2k1us: Ahh Good to know!

react-components/src/apps/quickLaunch/QuickLaunchListing.js
```diff
@@ -0,0 +1,339 @@
...
+            "http//:" +
+            window.location.host +
+            "/" +
+            constants.QUICK_LAUNCH_EMBED_ICON;
+        const embed =
+            "<a href='" +
+            getShareUrl(selected.id) +
+            "' target='_blank'><img src='" +
+            img_src +
+            "'></a>";
```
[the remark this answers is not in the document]
sri2k1us: Yes! This is much better. I still haven't gotten used this syntax. That why i didn't think of doing it this way. And some `java` hangover too 😉 .

de-lib/src/main/java/org/iplantc/de/client/util/AnalysisSubmissionUtil.java
```diff
@@ -0,0 +1,105 @@
...
+                    value.assign(configSplit, arg.getId());
+                } else if (diskResourceArgumentType && !arg.getType()
+                                                           .equals(ArgumentType.MultiFileSelector)) {
+                    if (value.get("path") != null) {
+                        value.get("path").assign(configSplit, arg.getId());
+                    }
+                } else if (arg.getType().equals(ArgumentType.MultiFileSelector) && value.isIndexed()) {
+                    value.assign(configSplit, arg.getId());
+                } else if (arg.getType().equals(ArgumentType.TreeSelection) && (arg.getSelectionItems()
+                                                                                != null) && (
```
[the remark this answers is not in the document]
sri2k1us: Well...IntelliJ doen't want to format this in any other 😕

react-components/src/apps/details/AppInfoDialog.js
```diff
@@ -115,18 +137,52 @@ function AppInfoDialog(props) {
         presenter.onClose();
     };
 
+    const onCreateQuickLaunch = () => {
+        presenter.onRequestToCreateQuickLaunch(app.id);
+        onClose();
+    };
+
+    const useQuickLaunch = (qLaunch) => {
+        presenter.getAppInfoForQuickLaunch(qLaunch.id, qLaunch.app_id);
```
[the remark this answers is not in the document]
sri2k1us: 🆗

react-components/src/apps/messages.js
```diff
@@ -61,20 +61,20 @@ const intlData = {
...
-        qLaunchToolTip: "Use this quick launch",
+        publicQLTooltip: "Quick Launch visible to all users",
+        qLaunchToolTip: "Use this Quick Launch",
         qLaunchEmbedToolTip: "Copy embed code",
         quickLaunchLabel: "Quick Launch",
         embedLbl: "Embed",
         shareLbl: "Share",
         deleteLbl: "Delete",
-        qLaunchShareToolTip: "Share quick launch URL",
+        qLaunchShareToolTip: "Share Quick launch URL",
```
psarando: Missed an `L` 😉

sri2k1us:
Thanks for the reviews.

PR #412, opened 2019-04-26
Edit comments dialog

jack-mitt:
editcomments1

This is the edit comments dialog written in react. This pull request only contains the UI portion of this element. Currently the "Retract" and "+" buttons do nothing, but in the future will make API requests to retract or add a comment respectively. The comments can be sorted by recency (dates will all have the same format once api is implemented) and by owner, these sorts can be seen after clicking the dot menu.

editcommentsdd

The comment section itself is in a Dialog Content element and is therefore scrollable, there is a max width so longer comments will wrap.

### DIFF
--- a/react-components/src/data/edit/EditComments.js
+++ b/react-components/src/data/edit/EditComments.js
@@ -1,0 +1,207 @@
+import React, {Component} from "react";
+import Dialog from '@material-ui/core/Dialog';
+import Button from '@material-ui/core/Button';
+import DeleteIcon from '@material-ui/icons/Delete';
+import { withStyles } from '@material-ui/core/styles';
+import DEDialogHeader from  "./../../util/dialog/DEDialogHeader.js"
+import DialogContent from '@material-ui/core/DialogContent';
+import exStyles from "./style.js"
+import DotMenu from "./DotMenu.js"
+import Comment from "./Comment.js"
+import TextField from '@material-ui/core/TextField'
+import Fab from '@material-ui/core/Fab'
+import AddIcon from '@material-ui/icons/Add'
+import CyVersePalette from "../../util/CyVersePalette";
+import List from '@material-ui/core/List';
+
+
+class EditComments extends Component{
+    state = {
+        anchorEl: null,
+    };
+    constructor(props){
+        super(props);
+        this.state = {
+            open: true,
+            commentList : null,
+        };
+        this.handleClose = this.handleClose.bind(this);
+        this.handleSortMostRecent = this.handleSortMostRecent.bind(this);
+        this.handleSortLeastRecent = this.handleSortLeastRecent.bind(this);
+        this.handleSortOwner = this.handleSortOwner.bind(this);
+        this.handleChange = this.handleChange.bind(this);
+        this.createComment = this.createComment.bind(this);
+    }
+
+    componentDidMount(){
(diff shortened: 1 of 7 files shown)


react-components/src/data/edit/EditComments.js
```diff
@@ -0,0 +1,207 @@
...
+                    <DialogContent id="editCommentsCommentList" className={classes.dContent}>
+                        <List component="nav">
+                            {commentItems}
+                        </List>
+                    </DialogContent>
+                    <TextField
+                        value={this.state.multiline}
+                        onChange={this.handleChange('multiline')}
+                        className={classes.addCommentTextField}
+                        label="Add a Comment"
```
sri2k1us: I will comment here once. But all the Strings need to be externalized. Please see https://github.com/cyverse-de/ui/blob/master/react-components/src/analysis/view/AnalysesView.js for example.

react-components/src/data/edit/EditComments.js
```diff
@@ -0,0 +1,207 @@
...
+                        />
+
+                    <DialogContent id="editCommentsCommentList" className={classes.dContent}>
+                        <List component="nav">
+                            {commentItems}
+                        </List>
+                    </DialogContent>
+                    <TextField
+                        value={this.state.multiline}
+                        onChange={this.handleChange('multiline')}
```
sri2k1us: this will immediately invoke the function `handleChange`. If you want to invoke a function onChange event, the you need to do something like `onChange={(parameterTopass)=>this.handleChange(parameterTopass)}`

react-components/src/data/edit/EditComments.js
```diff
@@ -0,0 +1,207 @@
...
+                    date={comment.date} owner={comment.owner}/>
+    ) : [];
+        return(
+            <div className={classes.root}>
+                <Dialog
+                    open={this.state.open}
+                    onClose={this.handleClose}
+                    id={"edit-comments-dialog"}
+                    className={classes.main}
+                >
```
sri2k1us: From you screenshot, the dialog looks very narrow. You can use some `maxWidth` and `fullWidth` props to make it little wider (according to the screen size). See https://material-ui.com/api/dialog/

react-components/src/data/edit/EditComments.js
```diff
@@ -0,0 +1,207 @@
...
+                            className={this.props.classes.dropDownDots}
+                        />
+
+                    <DialogContent id="editCommentsCommentList" className={classes.dContent}>
+                        <List component="nav">
+                            {commentItems}
+                        </List>
+                    </DialogContent>
+                    <TextField
+                        value={this.state.multiline}
```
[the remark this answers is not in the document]
jack-mitt: I don't think it had a use. I have removed it, I simply wrote it based on a material UI example.

react-components/src/data/edit/EditComments.js
```diff
@@ -0,0 +1,207 @@
...
+
+        }
+        this.setState({commentList : temp});
+    };
+
+    handleChange = name => event => {
+        this.setState({
+            [name]: event.target.value,
+        });
+    };
```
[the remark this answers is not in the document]
jack-mitt: Also stolen from material ui and due to my lack of state understanding, forgot to remove.

PR #415, opened 2019-05-06
Enable any JSON obj to show in ErrorHandler messages

ash3rz:
The logic I wrote assumed the messages would be strings, but they're sometimes json objects.  Should fix issues like this:

image

image


ash3rz:
Thanks for the review!

PR #417, opened 2019-05-10
CORE-17 Bypass user-info call on empty list during sharing

ash3rz:
This was causing schema validation errors to pop up


ash3rz:
Ah, even better!  Getting a little rusty with GWT already 😅

ash3rz:
@psarando I did a force push with a new commit

ash3rz:
Thanks for the review and the cleaner suggestion!

PR #419, opened 2019-05-14
Add static IDs to app info and metadata dialogs

ash3rz:
These should help me patch up some failing GUI tests


ash3rz:
Thanks for the review!

PR #422, opened 2019-05-22
core-287 Fix analyses selection bug.

sri2k1us:
Reset selection when a filter is applied.


sri2k1us:
Thanks for the reviews.

PR #425, opened 2019-06-04
Core 289 Fix Belphegor App Stats

sri2k1us:
- Fix date parameter being sent the service.


sri2k1us:
Thanks for the reviews.

PR #428, opened 2019-06-24
CORE-308 Fix for using public Quick Launches.

psarando:
This PR will move the `ActionsPopper` components out of the `ListQuickLaunches` Grid list and replace them with a single `ActionsPopper`.

Otherwise an `ActionsPopper` will be created for every Quick Launch item in the list, and they will all show at the same time when a public `qLaunch` is clicked, with the popover for the last `qLaunch` in the list always on top.

This caused the `useQuickLaunch` button to always open the last `qLaunch` in that list.

Also refactored the `ActionsPopper` debug IDs so they are associated with
the buttons that also have the `onClick` handlers.

### DIFF
--- a/react-components/src/apps/quickLaunch/QuickLaunchListing.js
+++ b/react-components/src/apps/quickLaunch/QuickLaunchListing.js
@@ -40,7 +40,6 @@ import Share from "@material-ui/icons/Share";
 
 function ActionsPopper(props) {
     const {
-        qLaunch,
         anchorEl,
         intl,
         useQuickLaunchClickHandler,
@@ -49,7 +48,7 @@ function ActionsPopper(props) {
         onActionPopperClose,
         baseDebugId,
     } = props;
-    const actionsBaseId = build(baseDebugId, qLaunch.id);
+
     return (
         <Popover
             open={Boolean(anchorEl)}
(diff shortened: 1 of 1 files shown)


react-components/src/apps/quickLaunch/QuickLaunchListing.js
```diff
@@ -236,12 +239,11 @@ function ListQuickLaunches(props) {
                                             }
                                         />
                                         <ActionsPopper
```
ash3rz: I think the heart of the issue is here, which was kinda fun to spot!

This is a popover that will open only if `anchorEl` is set and anchor to whatever the current `anchorEl` is, but this is within the `map` function.  So whenever you click a quick launch, the `anchorEl` gets set to that element, then N number of popovers all render at the same time, all anchored to the same element which you just clicked. The last element is the one you see on top and thus the only one you can interact with.

My guess is if you find a way to move the `ActionsPopper` rendering out of `map`, it should work.

psarando: I think you're right! Good catch! 🔍

psarando:
Thanks for the review! 👍

PR #429, opened 2019-06-25
CORE-294 Refresh Data windows after upload completed

ash3rz:


ash3rz:
Thanks for the review!

PR #430, opened 2019-06-25
CORE-9 Remove includefiles param from directory endpoint calls.

psarando:
This PR will remove the `includefiles` param from `/filesystem/directory` endpoint calls.

The `data-info` directory endpoint's `includefiles` param is obsolete, so it will no longer be supported in terrain's API after cyverse-de/terrain#127 (it's already ignored anyway).


psarando:
I'll merge this now, but reviews still welcome 👍

PR #431, opened 2019-07-02
Update React ErrorHandler component reference

ash3rz:
The ErrorHandler component moved from the CyVerseReactComponent's util directory to within ui-lib


ash3rz:
Thanks for the review!

PR #432, opened 2019-07-09
CORE-320 Open App Launch with default settings if QL has been deleted

ash3rz:
The App Launch window will instead open like normal with no params set and there will be an error announcement to let the user know the QL was deleted.  
I made this error announcement one that doesn't automatically get dismissed to try and avoid any confusion - for example, if the announcement gets dismissed before the user sees it, they might wonder if the QL is broken instead of deleted.  I could change it back to being auto-dismissed if others think it might be more annoying than helpful.


de-lib/src/main/java/org/iplantc/de/theme/base/client/apps/widgets/AppLaunchViewDisplayStrings.properties
```diff
@@ -4,4 +4,5 @@ dontShow = Do not show me this message again.
 hpcAppWaitTimes = <p> This application runs on XSEDE, the U.S. National Supercomputing Network. Analyses run on this system enter a queue that includes many non-CyVerse users. Wait times for your analyses (jobs) to submit and run may be up to several days for large-memory applications. Please do not resubmit your analysis. If the status of your analysis has not changed from Submitted to Running after several days, contact support@cyverse.org. To learn more about HPC apps, see <a href='https://pods.iplantcollaborative.org/wiki/x/MYOdAQ' target='_blank'> Using HPC Apps in the DE </a>.</p> <p>Apply for your own XSEDE allocation to gain access to additional resources at <a href="https://portal.xsede.org/allocation-request-steps" target="_blank"> XRAS - Submit Allocation Request </a>. Access to this powerful system is made available to CyVerse users at no cost through a grant from the National Science Foundation. </p>
 launchPreviewHeader=Preview of {0}
 createQuickLaunchSuccess = {0} quick launch created successfully.
+quickLaunchDeleted= The selected Quick Launch has been deleted. The selected app will be opened with cleared settings.
```
psarando: Since it's possible for apps to have some default params set by the app integrator:
```suggestion
quickLaunchDeleted= The selected Quick Launch has been deleted. The selected app will be opened with its default settings.
```

de-lib/src/main/java/org/iplantc/de/apps/widgets/client/presenter/AppLaunchPresenterImpl.java
```diff
@@ -82,6 +82,21 @@ public void onSuccess(AppTemplate result) {
...
+    private class QuickLaunchTemplateCallback extends AppTemplateCallback {
+        private AppWizardConfig config;
+
+        public QuickLaunchTemplateCallback(AppWizardConfig config) {
+            this.config = config;
+        }
+
+        @Override
+        public void onFailure(Integer statusCode, Throwable caught) {
+            announcer.schedule(new ErrorAnnouncementConfig(appearance.quickLaunchDeleted(), true, 0));
```
psarando: 👍 I think it's a good idea to leave this error message displayed until dismissed.

Especially if someone is clicking a QL link from some other site, and they are still getting their bearings in the DE after the auto-dismiss timeout occurs. 😵

de-lib/src/main/java/org/iplantc/de/theme/base/client/apps/widgets/AppLaunchViewDisplayStrings.properties
ash3rz: Ah good point, I'd forgotten about default params!

ash3rz:
Thanks for the review!

PR #434, opened 2019-07-12
CORE-318 cont... Belphegor

ash3rz:
This covers the conversion to React for Belphegor's Tool Admin details view and Tool Request make public view.

image


react-components/src/tools/EditTool.js
```diff
@@ -167,12 +202,46 @@ function EditToolForm(props) {
                     </FormSelectField>
                 )}
             />
+            {isAdmin && (
+                <Field
+                    name="interactive"
```
psarando: Since this flag should always be set to `true` when the tool type is `interactive`, maybe we can have the [ToolAdminPresenterImpl](https://github.com/cyverse-de/ui/blob/a8b15fe363e8ece4b32e798d0d7576cd53f9de0b/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/presenter/ToolAdminPresenterImpl.java#L224-L228) auto-set this field, instead of forcing the admin to remember to set or unset this for `interactive` types. We do this for the (non-admin) DE form already:
https://github.com/cyverse-de/ui/blob/4fbf931cbf72bcb69c71379e2c30b4f088011de7/de-lib/src/main/java/org/iplantc/de/tools/client/presenter/ManageToolsViewPresenter.java#L204-L222
Same for the `skip_tmp_mount` flag below.

Maybe we can also auto-set the `network_mode` to `bridge` for `interactive` tools in the admin form, but we should still allow admins to freely change this field for any tool type (DE users are restricted from setting this flag in private tools).

react-components/src/tools/EditTool.js
```diff
@@ -255,21 +387,36 @@ const DEFAULT_TOOL = {
...
     }
 }
 
 EditToolDialog.propTypes = {
     open: PropTypes.bool.isRequired,
+    isAdmin: PropTypes.bool.isRequired,
+    isAdminPublishing: PropTypes.bool.isRequired,
     presenter: PropTypes.shape({
-        addTool: PropTypes.func.isRequired,
-        updateTool: PropTypes.func.isRequired,
```
psarando: We don't want these functions to be required?

react-components/src/tools/ToolImplementation.js
```diff
@@ -0,0 +1,166 @@
...
+                            parentId,
+                            ids.EDIT_TOOL_DLG.SAMPLE_INPUT_FILES
+                        )}
+                    />
+                )}
+            />
+            <FieldArray
+                name={`${name}.test.output_files`}
+                render={({ ...props }) => (
+                    <TestFiles
```
psarando: 👍

react-components/src/tools/ToolRestrictions.js
```diff
@@ -108,28 +143,44 @@ function Restrictions(props) {
...
                 name="container.network_mode"
                 label={getMessage("networkMode")}
                 id={build(parentId, ids.EDIT_TOOL_DLG.NETWORK_MODE)}
-                disabled
-                component={FormTextField}
-            />
+                disabled={!isAdmin}
+                component={FormSelectField}
+            >
+                {NETWORK_MODES.map((mode, index) => (
```
psarando: 👍

react-components/src/tools/ToolRestrictions.js
```diff
@@ -108,28 +143,44 @@ function Restrictions(props) {
...
+                <Typography variant="caption">
+                    {getMessage("restrictionsSupport")}
+                </Typography>
+            )}
             <Field
                 name="container.pids_limit"
                 label={getMessage("pidsLimit")}
                 id={build(parentId, ids.EDIT_TOOL_DLG.PIDS_LIMIT)}
-                disabled
+                disabled={!isAdmin}
```
psarando: 👍

react-components/src/tools/EditTool.js
ash3rz: So for Manage Tools and Tool Admin, they require `addTool` and `updateTool`.  But for Tool Request, it only has `onPublish`. I realized you can add custom validation functions, so I'll add that for this.

psarando: I forgot (or hadn't realized yet) that this form is also reused for Tool Requests. 👌

react-components/src/tools/ContainerPorts.js
```diff
@@ -87,6 +108,38 @@ function ContainerPorts(props) {
                                         component={FormNumberField}
                                     />
                                 </TableCell>
+                                {isAdmin && (
```
psarando: Oh yeah, 👍 for the original purpose of CORE-318 😆

react-components/src/tools/EditTool.js
ash3rz: Ah, good catch.
Do we want these to be in ToolAdminPresenter or do we want these to happen directly in the form while the admins are changing things?

psarando: I think we can hide the flags from the UI and let the Presenter manage them. We can add them back to the form later, if admins need to manually set these flags for some reason in the future.

The `network_mode` setting can stay in the form, though.

ash3rz: One more question - ManageToolsPresenter also adds this stuff automatically.  Do we want that here as well?
https://github.com/cyverse-de/ui/blob/0cd2f34f508340b9afe79358d04140e8fc7f4983/de-lib/src/main/java/org/iplantc/de/client/models/tool/ToolAutoBeanFactory.java#L62-L69

psarando: It looks like it already does: https://github.com/cyverse-de/ui/blob/a8b15fe363e8ece4b32e798d0d7576cd53f9de0b/de-lib/src/main/java/org/iplantc/de/admin/desktop/client/toolAdmin/presenter/ToolAdminPresenterImpl.java#L224-L228

ash3rz: Ah, you're right.  Was looking in the wrong spot. Cool, I'll push out an update

react-components/src/tools/EditTool.js
```diff
@@ -230,13 +344,20 @@ function resetOnTypeChange(currentType, form) {
     }
     if (currentType !== "interactive") {
         form.setFieldValue("container.container_ports", null);
+        form.setFieldValue("container.network_mode", "none");
+    } else {
+        form.setFieldValue("container.network_mode", "bridge");
```
psarando: 👍

ash3rz:
Thanks for the review!!

PR #435, opened 2019-07-12
Clean up unusable-by-js warnings

ash3rz:


de-lib/src/main/java/org/iplantc/de/apps/client/AppDetailsView.java
```diff
@@ -32,24 +32,29 @@
     @JsType
     interface Presenter {
 
+        @SuppressWarnings("unusable-by-js")
         void go(App app,
                 String searchRegexPattern);
```
psarando: I wouldn't think these `Presenter.go` methods would be used by our javascript code 🤔 
Should these `go` methods be annotated with `@JsIgnore` instead?

ash3rz: Oh whoops 😊 you're probably right

ash3rz:
Thanks for the review!

PR #436, opened 2019-07-17
CORE-326 Add icons and tooltips to the desktop User menu items

ash3rz:
Here's an example:
image

Very much open to suggestions on how to make the tooltip wordings more clear

### DIFF
--- a/react-components/src/desktop/messages.js
+++ b/react-components/src/desktop/messages.js
@@ -30,6 +30,11 @@ var intlData = {
         emailSupportTooltip: "support@cyverse.org",
         supportChatLink: "Chat With Support",
         supportChatTooltip: "Chat may be disabled by your browser's settings",
+        collaboratorsTooltip: "Manage your private collaborator lists",
+        teamsTooltip:
+            "Manage public or private groups of collaborators. Collaborators in a team can all share resources with that team.",
+        communitiesTooltip:
+            "Manage public collections of apps for display in Community categories in the Apps window",
     },
 };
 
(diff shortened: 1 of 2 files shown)


react-components/src/desktop/messages.js
```diff
@@ -30,6 +30,11 @@ var intlData = {
         emailSupportTooltip: "support@cyverse.org",
         supportChatLink: "Chat With Support",
         supportChatTooltip: "Chat may be disabled by your browser's settings",
+        collaboratorsTooltip: "Manage your private set of collaborators",
```
psarando: May as well get the `List` terminology in here too:
```suggestion
        collaboratorsTooltip: "Manage your private collaborator lists",
```

react-components/src/desktop/messages.js
```diff
@@ -30,6 +30,11 @@ var intlData = {
...
+        teamsTooltip:
+            "Manage public or private groups of collaborators. Collaborators in a group can all share resources within the group",
```
psarando: It might be better to stick with the `Team` terminology here:
```suggestion
            "Manage public or private groups of collaborators. Collaborators in a team can all share resources with that team.",
```

react-components/src/desktop/messages.js
```diff
@@ -30,6 +30,11 @@ var intlData = {
...
+        communitiesTooltip:
+            "Manage public collections of apps for displaying in the Apps window",
```
psarando:
```suggestion
            "Manage public collections of apps for display in Community Categories in the Apps window",
```
This might make it more clear what Communities are for currently, though I'm sure we'll have to change this text if Communities are expanded for other things in the future 🤔

ash3rz: Agreed - right now we have wording in the help buttons for Collaborator and Team windows that indicates those groups only work with apps and tools right now as well.

ash3rz:
Thanks for the review!

PR #437, opened 2019-07-20
CORE-329 Convert App Details dialog in Belphegor to React

ash3rz:
image


react-components/src/apps/admin/AdminAppDetails.js
```diff
@@ -0,0 +1,314 @@
...
+            setSubmitting(false);
+        });
+};
+
+function mapPropsToValues(props) {
+    const { app, documentation } = props;
+    return {
+        app: app,
+        documentation: documentation,
+    };
```
psarando: Similar to destructuring, the key doesn't have to be specified here if it's the same as the variable name:
```javascript
     return {
        app,
        documentation,
    };
```
And to get really terse, we can do away with this function definition all together and use an `=>` function below:
```javascript
 export default withFormik({
    enableReinitialize: true,
    mapPropsToValues: ({ app, documentation }) => ({ app, documentation }),
    handleSubmit,
})(withI18N(AdminAppDetailsDialog, messages));
```

react-components/stories/apps/admin/AdminAppDetails.stories.js
```diff
@@ -0,0 +1,108 @@
...
+
+class AdminAppDetailsTest extends Component {
+    render() {
+        const presenter = {
+            closeAppDetailsDlg: () => console.log("close"),
+
+            onSaveAppSelected: (app, resolve, reject) => {
+                setTimeout(() => {
+                    console.log("Save app selected", app);
+                    resolve();
```
psarando: 👍

react-components/src/apps/admin/AdminAppDetails.js
```diff
@@ -0,0 +1,314 @@
...
+                    onClick={handleSubmit}
+                >
+                    {getMessage("save")}
+                </Button>
+            </DialogActions>
+        </Dialog>
+    );
+}
+
+const StyledAdminAppDetailsFrom = withStyles(styles)(AdminAppDetailsForm);
```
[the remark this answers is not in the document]
ash3rz: Oops!  Good catch

react-components/src/apps/admin/AdminAppDetails.js
ash3rz: Ah, terse is nicer!
I end up needing to modify this a bit in #438, so I'll save the update for that PR

ash3rz:
Thanks for the review!

PR #438, opened 2019-07-22
CORE-327 Add htcondor extra requirements to Belphegor

ash3rz:
image

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/client/models/apps/App.java
+++ b/de-lib/src/main/java/org/iplantc/de/client/models/apps/App.java
@@ -149,4 +149,10 @@ public interface App extends HasQualifiedId,
 
     @PropertyName("job_stats")
     AppStats getAppStats();
+
+    AppExtra getExtra();
+    void setExtra(AppExtra extra);
+
+    AppDoc getDocumentation();
+    void setDocumentation(AppDoc appDoc);
 }
--- a/react-components/src/apps/admin/AdminAppDetails.js
+++ b/react-components/src/apps/admin/AdminAppDetails.js
@@ -293,11 +293,6 @@ const handleSubmit = (values, { props, setSubmitting }) => {
         });
 };
 
-function mapPropsToValues(props) {
-    const { app } = props;
-    return { ...app };
-}
-
 AdminAppDetailsDialog.propTypes = {
     open: PropTypes.bool.isRequired,
     presenter: PropTypes.shape({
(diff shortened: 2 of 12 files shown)


de-lib/src/main/java/org/iplantc/de/client/models/apps/App.java
```diff
@@ -149,4 +149,10 @@
 
     @PropertyName("job_stats")
     AppStats getAppStats();
+
+    AppExtra getExtra();
+    void setExtra(AppExtra extra);
+
+    AppDoc getDocumentation();
+    void setDocumentation(AppDoc appDoc);
```
psarando: 👍 
This is slightly confusing later, when we have to access `app.documentation.documentation` to get to the doc text.
Would it look better to use `app.docsinfo.documentation` / `getDocsInfo` / `setDocsInfo`, or something like that?

react-components/src/apps/admin/AdminAppDetails.js
```diff
@@ -309,6 +312,6 @@ AdminAppDetailsDialog.propTypes = {
 
 export default withFormik({
     enableReinitialize: true,
-    mapPropsToValues,
+    mapPropsToValues: ({ app }) => ({ ...app }),
```
psarando: 👍

de-lib/src/main/java/org/iplantc/de/client/models/apps/App.java
psarando: Either way, 👍 for bundling this with the other app info.

react-components/src/apps/messages.js
```diff
@@ -86,6 +86,7 @@ const intlData = {
             "the wiki URL field once you fill this field. Replace everything inside \\{\\{\\}\\}.",
         appDocumentation: "DE App Documentation",
         documentationTemplate: "View Documentation Template",
+        htcondorExtraRequirements: "HT Condor Extra Requirements",
```
ianmcorvidae: Extremely tiny but should probably be "HTCondor" (without the space), that's what their stuff uses.

de-lib/src/main/java/org/iplantc/de/client/models/apps/App.java
ash3rz: 🤔 The keys back from the service are 
```
documentation: {
    app_id: "someId", 
    documentation: "myDocs",
    ...
}
```
So I could change the function name here in `App.java` to `getDocsInfo` and add the `@PropertyName` annotation to properly access `documentation`, but in javascript, it will still be `app.documentation.documentation` I thought because those are the real keys.

psarando: Oh, oops. I didn't realize this already matched the service response. I thought we were using a different endpoint for docs and just bundling in the UI for convenience.

In that case, let's leave it this way so they match.

ash3rz:
Thanks for the reviews!

PR #439, opened 2019-07-23
added the Internet2 InCommon intermediate CA certificate

slr71:
NCSU uses Internet2's InCommon certificate authority. Adding it as a trusted certificate authority shouldn't pose much of a risk, and will make it easier for us to deploy to sites that use it.


slr71:
Thanks for the review!

PR #441, opened 2019-07-23
CORE-334 Disable VICE launch button from opening Data window

ash3rz:


ash3rz:
I'm gonna close this - I realized there are more icons we have in the Analysis window that probably should also not open the Data window when clicked, but I need to test them out.

ash3rz:
It seems the other icons in the Analysis window don't trigger the Data window to open.  Reopening this PR.

react-components/src/analysis/view/AnalysesView.js
```diff
@@ -92,20 +92,22 @@ function AnalysisName(props) {
...
-                onClick={() => handleGoToOutputFolder(analysis)}
-            >
+            <Fragment>
                 <LaunchIcon
                     onClick={() =>
                         handleInteractiveUrlClick(interactiveUrls[0])
                     }
                     id={build(baseId, ids.ICONS.INTERACTIVE)}
-                    style={{ color: palette.darkBlue }}
+                    className={interactiveStyle}
```
psarando: ✨

ash3rz:
Thanks for the review!

PR #443, opened 2019-08-05
4.0

sri2k1us:


react-components/config/webpack.config.prod.js
```diff
@@ -260,12 +266,13 @@ module.exports = function(webpackEnv) {
...
-                .map(ext => `.${ext}`)
-                .filter(ext => useTypeScript || !ext.includes('ts')),
+                .map((ext) => `.${ext}`)
+                .filter((ext) => useTypeScript || !ext.includes("ts")),
             alias: {
                 // Support React Native Web
                 // https://www.smashingmagazine.com/2016/08/a-glimpse-into-the-future-with-react-native-for-web/
-                'react-native': 'react-native-web',
+                "react-native": "react-native-web",
+                react: path.resolve("./node_modules/react"),
```
ash3rz: For anyone else reviewing, I believe this is the only change in this file.

react-components/package.json
```diff
@@ -20,9 +20,9 @@
         "prop-types": "^15.6.2"
     },
     "dependencies": {
-        "@cyverse-de/ui-lib": "0.3.4",
-        "@material-ui/core": "^3.9.3",
-        "@material-ui/icons": "^3.0.2",
+        "@cyverse-de/ui-lib": "file:../../ui-lib",
```
ash3rz: This should probably be the version pushed instead of a path

sri2k1us: Correct. I have changes to merge in https://github.com/cyverse-de/ui-lib/pull/12. Once that is approved and merged, I will update this.

react-components/config/webpack.config.prod.js
psarando: @sri2k1us I have pushed some prettier settings so our IDEs can preserve the styles from the original CRA ejected files.

If you rebase this PR on top of master now and reformat this file with the new settings, it will help clean-up this PR's diff.

sri2k1us: @psarando - Thanks. That helped with getting the original formatting back but it still changed all double quotes to single :(

sri2k1us: Updated configs to use double quotes.

psarando: @sri2k1us: Yes, the configs were intentionally set to use single quotes, because that's how the ejected file was originally formatted by CRA, and it's also how it was formatted before the changes in this PR.

Keeping the original formatting, including the single quotes, will not only clean up this file's diff in this PR, it will also allow us to run an easy diff with the original CRA ejected scripts to help us quickly find our custom changes.

sri2k1us: 🆗 Undid my last commit. Back to single to quotes.

PR #446, opened 2019-08-24
CORE-501: allow multiple IP addresses in the X-Forwarded-For header

slr71:


de-lib/src/main/java/org/iplantc/de/server/auth/BaseUrlConnector.java
```diff
@@ -161,9 +161,9 @@ protected String addIpAddress(String uriString, HttpServletRequest request) thro
                                             realIp);
         }
         if(!Strings.isNullOrEmpty(forwardedForIp)){
-            ret = appLoggerUtil.addIpHeader(c,
-                                            FWDED_FOR_IP_HEADER_NAME,
-                                            forwardedForIp);
+            ret = appLoggerUtil.addIpListHeader(c,
+                                                FWDED_FOR_IP_HEADER_NAME,
+                                                forwardedForIp);
```
ianmcorvidae: Is there something that'll verify if this is/isn't a list, or is that unnecessary? Looking at the prior code it looks like it should just be a single IP most of the time (but I'm guessing I don't totally understand the need for the change -- maybe we've usually been getting just the frontend nginx's x-forwarded-for but sometimes people are using proxies that also add stuff and it's making this code fail?).

de-lib/src/main/java/org/iplantc/de/server/AppLoggerUtil.java
```diff
@@ -64,6 +64,28 @@ public static AppLoggerUtil getInstance(){
...
+     * @param ipAddresses the list of IP addresses to use as the header value.
+     * @param <T> Basic request object
+     * @return the request with the new header added.
+     */
+    public <T extends HttpRequestBase> T addIpListHeader(T request,
+                                                         final String headerName,
+                                                         final String ipAddresses) {
+        for (String addr : ipAddresses.split("\\s*,\\s*")) {
+            Preconditions.checkArgument(InetAddresses.isInetAddress(addr),
+                                        "Invalid IP address: " + addr);
```
ianmcorvidae: I worry a little about stripping off the X-Forwarded-For entirely (or erroring out) if we're passed something that's malformed -- I guess the prior code would already do that though, so maybe this is still an improvement. If we know what's consuming the `X-Forwarded-For` information, would it maybe be possible to push this checking to there, to ensure we're not likely to toss out any useful information (even if possibly malformed)?

de-lib/src/main/java/org/iplantc/de/server/auth/BaseUrlConnector.java
slr71: From what I've read, it appears that `X-Forwarded-For` can contain any number of IP addresses. I'd originally planned to modify the deployment so that the header would contain only a single IP address, but it seemed more reasonable to support any valid value for this header.

I'm not quite sure I understand why we'd need to determine if the value contains a list or not. Since this is a delimited string, a header value containing a single IP address can be treated as a single-element list, and the validation should work fine.

The specific case that this is addressing is where `X-Forwarded-for` contains a value that looks like `<original-ip-address>, <intermediate-ip-address>` and calls to the DE services are failing because the entire header value doesn't represent a single IP address.

psarando: I don't think we need to verify if the string is a comma-separated list or not. I think the `split` call in `addIpListHeader` will return an array of 1 item if the string contains only 1 IP address.

In fact, we could probably replace the `addIpHeader` method call everywhere else with this new `addIpListHeader` method call instead.

de-lib/src/main/java/org/iplantc/de/server/AppLoggerUtil.java
slr71: The request will currently error out if the header is formatted badly. For the time being, I was hoping to modify the code as little as possible while still getting it to work in the Austria deployment. I wouldn't be opposed to making this sort of improvement, but I'd want to take more time to test a bigger change.

de-lib/src/main/java/org/iplantc/de/server/auth/BaseUrlConnector.java
ianmcorvidae: Ah, yup, I was misunderstanding the need. This looks reasonable to me now! It's a super nitpick but `forwardedForIp` might not be the best name now that we know this may or may not be a single IP address. Of course, I'm not sure I have an alternate suggestion, so if you ignore me I won't mind :laughing:

de-lib/src/main/java/org/iplantc/de/server/AppLoggerUtil.java
ianmcorvidae: That sounds sensible to me, thanks for answering.

de-lib/src/main/java/org/iplantc/de/server/auth/BaseUrlConnector.java
slr71: I briefly considered using `addIpListHeader` for both, but the documentation that I found on `X-Real-Ip` seemed to indicate that it should only contain a single IP address if it's used. It's probably not that critical, but I thought I'd leave the validation for that header as-is to avoid allowing invalid header values through.

psarando: Ah, good point. I didn't bother checking where else `addIpHeader` was getting called, so leaving that one alone sounds good to me 👍

slr71:
Thanks for the reviews!

PR #449, opened 2019-08-28
CORE-116, CORE-150, CORE-239

ash3rz:
Mostly just adds some static IDs that were requested by QA, and also fixes a selection bug in the Data window while searching


ash3rz:
Thanks for the review!

PR #451, opened 2019-09-06
Timelimit

sri2k1us:
image

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/client/services/impl/AnalysisServiceFacadeImpl.java
+++ b/de-lib/src/main/java/org/iplantc/de/client/services/impl/AnalysisServiceFacadeImpl.java
@@ -317,5 +317,20 @@ public class AnalysisServiceFacadeImpl implements AnalysisServiceFacade {
 
     }
 
+    @Override
+    public void getViceTimeLimit(String id, DECallback<String> callback) {
+        String address = ANALYSES + "/" + id + "/time-limit";
+        ServiceCallWrapper wrapper = new ServiceCallWrapper(GET, address);
+        deServiceFacade.getServiceData(wrapper, callback);
+    }
+
+    @Override
+    public void setViceTimeLimit(String id, DECallback<String> callback) {
+        String address = ANALYSES + "/" + id + "/time-limit";
+        ServiceCallWrapper wrapper =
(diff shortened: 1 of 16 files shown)


react-components/src/analysis/view/dialogs/ViceLogsViewer.js
```diff
@@ -115,7 +115,7 @@ function ViceLogsViewer(props) {
                 </div>
                 <DialogContent dividers={true}>
                     <DialogContentText className={classes.content}>
-                        {logs}
+                        <pre>{logs}</pre>
```
sri2k1us: I sneaked this change in this PR.

de-lib/src/main/java/org/iplantc/de/client/services/impl/AnalysisServiceFacadeImpl.java
```diff
@@ -317,5 +317,20 @@ public void getVICELogs(String id,
...
 
+    @Override
+    public void getViceTimeLimit(String id, DECallback<String> callback) {
+        String address = ANALYSES + "/" + id + "/time-limit";
+        ServiceCallWrapper wrapper = new ServiceCallWrapper(GET, address);
+        deServiceFacade.getServiceData(wrapper, callback);
+    }
+
+    @Override
+    public void setViceTimeLimit(String id, DECallback<String> callback) {
```
psarando: Not sure about naming this method `setViceTimeLimit`, since it doesn't actually accept a time limit value to set with this endpoint call 🤔 

Maybe `extendViceTimeLimit` is more accurate?

react-components/src/analysis/view/AnalysesView.js
```diff
@@ -94,13 +96,15 @@ function AnalysisName(props) {
...
         return (
             <Fragment>
-                <LaunchIcon
-                    onClick={() =>
-                        handleInteractiveUrlClick(interactiveUrls[0])
-                    }
-                    id={build(baseId, ids.ICONS.INTERACTIVE)}
-                    className={interactiveStyle}
-                />
+                <ToolTip title={getMessage("goToVice")}>
```
psarando: 👍

sri2k1us:
Thanks for the review.

PR #452, opened 2019-09-26
CORE-506 Convert Teams window to React

ash3rz:
This is just the window, the dialogs will come next.

image


de-lib/src/main/java/org/iplantc/de/client/services/impl/GroupServiceFacadeImpl.java
```diff
@@ -410,15 +421,14 @@ public void denyRequestToJoinTeam(Group team,
...
-        deService.getServiceData(wrapper, new AsyncCallbackConverter<String, List<Group>>(callback) {
+        deService.getServiceData(wrapper, new AsyncCallbackConverter<String, Splittable>(callback) {
             @Override
-            protected List<Group> convertFrom(String object) {
-                GroupList groupList = AutoBeanCodex.decode(factory, GroupList.class, object).as();
-                return groupList.getGroups();
+            protected Splittable convertFrom(String object) {
+                return StringQuoter.split(object);
             }
         });
```
sri2k1us: This is matter of preference but all these 3 methods can be condensed into  1 method if we do
```java
public void getTeams(boolean myTeam, String searchTerm, AsyncCallback<Splittable> callback)
```
and then add relevant query strings to the service call. Once we do that, we can collapse `getMyTeams`, `getTeams` and `SearchTeams` methods in `TeamsPresenterImpl`  into 1 method like
```java
public void getTeams(String Filter, String search Term)
```
This will reduce the code duplication across these 3 methods in `TeamsPresenterImpl`.

And once that is done, in `Teams.js`, you can just use `useEffect` like this:
```jsx

useEffect(() => {
        presenter.getMyTeams(teamFilter, searchTerm);
    }, [teamFilter, searchTerm]);
```
and eliminate the manual call to `presenter` methods from `handleTeamFilterChange` and `handleSearch `

react-components/src/teams/Teams.js
```diff
@@ -0,0 +1,344 @@
...
+
+const TABLE_COLUMNS = [
+    { name: "Name", numeric: false, enableSorting: true },
+    {
+        name: "Creator Name",
+        numeric: false,
+        enableSorting: false,
+        key: "creator",
+    },
+    { name: "Description", numeric: false, enableSorting: false },
```
sri2k1us: We can skip `numeric` flag and use `align` if needed.

react-components/src/teams/styles.js
```diff
@@ -0,0 +1,27 @@
...
+    table: {
+        height:
+            "calc(100% - " +
+            theme.mixins.toolbar["@media (min-width:600px)"].minHeight +
+            "px)",
+        overflow: "auto",
+    },
+});
+
+export default styles;
```
sri2k1us: Since you have all the code related to `Teams` in 1 file `Teams.js`, we can move this style into the `js` file and use the new `useStyles` hook. This way we can remove the wrapping. This can help with `debugging` when using react browser plugin. Again just a suggestion.

de-lib/src/main/java/org/iplantc/de/client/services/impl/GroupServiceFacadeImpl.java
ash3rz: Ah, really good catch!  Will refactor.

ash3rz:
Thanks for the review!  I'll rebase real quick before merging since this branch is a bit behind.

PR #456, opened 2019-10-08
CORE-128 Disable Delete option for running jobs

ash3rz:
Since deleting a job does not cancel it, this prevents users from accidentally deleting running jobs.


ash3rz:
One of my smaller PRs 😄 Thanks for the review!

PR #457, opened 2019-10-09
Add docs on react dev and standards

ash3rz:
Now that we'll have more devs working on UI, figured it'd be good to get this all documented.  I'm sure I forgot things, but here's a start.


ash3rz:
Cool - thanks for the review!

PR #458, opened 2019-10-11
Update ManageTools table to use DETableRow

ash3rz:
Selected in blue, hover in gray:
image

I'm now leaning towards not adding the checkboxes back because we also use this view in the App Editor to select a tool for the app.  Having the checkboxes in that case might be confusing, as if you can create a workflow of tools.  We could add a flag to toggle having the checkboxes, but, as Sriram said, probably it's unlikely users are trying to select/deal with multiple tools in the Manage Tools window anyway.


ash3rz:
Agreed - will open a new PR with the rest of the tables updated.  
Thanks for the reviews!

PR #459, opened 2019-10-15
Add failure message when updating community members

ash3rz:


de-lib/src/main/java/org/iplantc/de/theme/base/client/communities/CommunitiesDisplayStrings.java
```diff
@@ -10,4 +12,12 @@
...
 
     String agaveAppsNotSupportedToolTip();
+
+    String joinCommunityFailure();
+
+    String leaveCommunityFailure();
+
+    String removeCommunityAdminFailure(@PluralCount List<String> failedUsers);
+
+    String addCommunityAdminFailure(@PluralCount List<String> failedUsers);
```
sri2k1us: 👍

de-lib/src/main/java/org/iplantc/de/communities/client/presenter/ManageCommunitiesPresenterImpl.java
```diff
@@ -383,6 +397,11 @@ public void onFailure(Throwable caught) {
 
             @Override
             public void onSuccess(List<UpdateMemberResult> result) {
+                List<String> failedUsers = getFailedMemberUpdateNames(result);
+
+                if (failedUsers.size() > 0) {
+                    announcer.schedule(new ErrorAnnouncementConfig(appearance.joinCommunityFailure()));
+                }
                 callback.onSuccess(null);
```
sri2k1us: Should we call error callback instead when the request fails for all the users ? 🤔

ash3rz: Ooh, really good catch - I think in all these cases, it actually makes more sense to call the `errorCallback`.  I'll refactor

ash3rz:
Thanks for the reviews!

PR #460, opened 2019-10-16
CORE-469 Add resource requirements to Analysis Launch dialogs

psarando:
This PR will add user adjustable resource requirements to Analysis Launch dialogs.

Resource Requirement panels will always be appended last in the current accordion panels view, after the analysis details and parameter group panels defined by the app integrator.

If there is only 1 step in the analysis, then one `Resource Requirements` panel will be appended to the dialog:
image

If there is more than 1 step in the analysis, then each step will append its own `Resource Requirements` panel:
image

The drop-downs for each requirement are populated with the same values available to tool integrators in the DE's add/edit private tool dialog.

* 0,1,2,4,8 minimum CPU
* 0,2,4,8,16 GiB minimum memory
* 0,1,2,4,8, and doubling each additional step up to 512GiB minimum disk space

If a step defines minimum CPU, RAM, or disk requirements, those values will be populated by default in that step's `Resource Requirements` panel. Additionally, min and max values set for these resources are used to limit which options are available in that resource's drop-down.

For example, the following tool has 3.4 min CPU, 4GiB max memory, and 4GiB min disk space defined. So the `Resource Requirements` panel in the launch dialog for an app using this tool will set 3.4 min CPU cores and 4GiB min disk space by default:
image

Additionally, if the user wishes to adjust the minimum CPU requirements, then only the options 0, 4, and 8 will be available. Similarly for memory, only 0, 2, and 4GiB options will be available; and for disk, the options for 1 and 2 GiB will not be available (since those are below the 4GiB minimum defined by the tool).

Selecting a `0` in any of these drop-downs has the effect of requesting no requirement for that resource in the Condor submission file.

As another example, our internal docs for creating VICE tools used to instruct admins to use 0.1 min CPUs, 2 max CPUs, and 4,000,000,000 max memory. So the `Resource Requirements` panel for the following `jupyter-lab` app will use 0.1 min CPUs by default, and limit the CPU options to 0, 1, or 2 CPUs. Minimum memory will not be auto-filled by default, but will be limited to only 0 or 2GiB for the user, since the 4,000,000,000 max memory limit is just less than the next 4GiB option:
image

The admin `Edit Tool` form has also been updated with validations to ensure the minimum requirements are not larger than their maximums. The RAM and Disk select list values have also been updated to use size formatting from the [numeral](http://numeraljs.com/#format-bytes) package.

### Before:
image
---
image
---

### After:
image
image


ash3rz:
Haven't reviewed yet - but thought this was a really good writeup/explanation that we might want to preserve in the Jira ticket if it's not already there.

de-lib/src/main/java/org/iplantc/de/apps/widgets/client/view/editors/AppStepResourcesViewImpl.java
```diff
@@ -0,0 +1,169 @@
...
+                limitSelectionList.add(resourceLimit);
+            }
+
+            resourceLimit *= 2;
+        }
+    }
+
+    private SimpleComboBox<Long> createDataSizeComboBox() {
+        final SimpleComboBox<Long> resourceSizeSimpleComboBox =
+                new SimpleComboBox<>(size -> Math.round((float)(size * 10 / ONE_GB)) / 10 + " GB");
```
ash3rz: I saw a similar equation like this elsewhere, but I'm not sure we need `* 10` and then also later `/ 10`?

psarando: This is to ensure the label can display a decimal for partial GB values with `0.1` precision.

For example, if the size were 1,632,087,572 bytes, it should display as "1.5 GB". If we didn't multiply the integer before rounding (which requires dividing after rounding), then `Math.round(1632087572 / ONE_GB) + " GB"` would give us `2 GB`. But `Math.round(1632087572 * 10 / ONE_GB) / 10` gives us the proper `1.5 GB`.

Similarly, if we wanted an extra precision, then we could do `Math.round(1632087572 * 100 / ONE_GB) / 100` to display `1.52 GB`.

ash3rz: I see - thanks for the explanation!

de-lib/src/main/java/org/iplantc/de/apps/widgets/client/view/editors/AppStepResourcesViewImpl.java
```diff
@@ -0,0 +1,169 @@
...
+        final long memoryEditorMax =
+                (memoryLimit != null && memoryLimit > 0) ? memoryLimit : deProperties.getToolsMaxMemLimit();
+        buildResourceSizeLimitList(memory, 2 * ONE_GB, memoryEditorMin, memoryEditorMax);
+
+        final Double maxCPUCores = limits.getMaxCPUCores();
+        final Double minCPUCores = limits.getMinCPUCores();
+
+        final double cpuEditorMin = (minCPUCores != null && minCPUCores > 1) ? Math.round(minCPUCores) : 1;
+        final double cpuEditorMax =
+                (maxCPUCores != null && maxCPUCores > 0) ? Math.round(maxCPUCores) : deProperties.getToolsMaxCPULimit();
```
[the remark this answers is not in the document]
psarando: This could be possible even if these checks were the same.

If a tool admin sets min CPUs to 2 and max CPUs to 1 (Belphegor still allows admins to set any arbitrary decimal for min CPUs, but the max CPUs are currently restricted to whole numbers starting with `1`), then in this launch dialog, min CPUs would be set to `2` by default, but only `0` would be available in the drop-down.

So we're counting on tool admins to set realistic min/max values, but `0` will always be available to clear the resource request in case an admin makes a mistake.

Although now I don't remember why I thought I needed to check for `min > 1` here. It must have been while I was debugging with VICE tools that have `0.1` min CPUs set, but checking for `min > 0` doesn't seem to be an issue with the current logic in `buildResourceDoubleLimitList` 🤔 

In fact, I don't think I even need to round the min/max limits passed to this method anymore. I'll update these lines before merging, to get rid of the rounding, and to make both checks the same, which I think makes the logic clearer.

de-lib/src/main/java/org/iplantc/de/apps/widgets/client/view/editors/AppStepResourcesViewImpl.java
ianmcorvidae: Heh, I wish `Math.round` just supported rounding to decimals. Maybe it would be good to separate out the scaling factor/precision into a constant? Could then do something like:

```
DISPLAY_PRECISION = 0.1
Math.round((float)(size /DISPLAY_PRECISION)) * DISPLAY_PRECISION + " GB")
```

or call it something like `SCALING_FACTOR` and set it to 10. Not sure which of those ways would be clearer/work better.

(something like `size * (1/DISPLAY_PRECISION)` could also work if dividing by 0.1 is too weird)

de-lib/src/main/java/org/iplantc/de/apps/widgets/client/view/editors/AppStepResourcesViewImpl.java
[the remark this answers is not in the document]
ash3rz: Hm... we could probably add a validation check to make sure on the Belphy side that the max is greater than the min.  Or I guess 0 would be a valid max too, so 0 or greater I guess?

sri2k1us:
LGTM. From the screen shot: The resource requirements for multi step app, each step is named `Resource Requirements for step n`. Is it possible to name it something line `Resource Requirements for step <stepname>`. This will make little bit more clear for the users.

psarando:
> Is it possible to name it something line `Resource Requirements for step <stepname>`.

This should be possible, but it will also require service updates for the https://qa.cyverse.org/terrain/docs/index.html#!/apps/get_terrain_apps_system_id_app_id endpoint, so that the step name is also returned along with the step number in each step's requirements. So I can open other PRs for those updates.

Then we could label multi-step resource requirements similar to the way the workflow editor labels them: `Resource Requirements for Step 1: <stepname>`.

de-lib/src/main/java/org/iplantc/de/apps/widgets/client/view/editors/AppStepResourcesViewImpl.java
psarando: > Heh, I wish `Math.round` just supported rounding to decimals. Maybe it would be good to separate out the scaling factor/precision into a constant?

I realized that I was casting ints to floats in the wrong places, so this wasn't even working as inteded before 😖
I've since fixed it for this PR: `(float)Math.round((float)(size * 10) / ONE_GB) / 10`

I've seen a few examples of formatting integers this way for javascript code, so I didn't think to clarify what it was doing before, but now I've also added a comment explaining what this conversion is doing.

When this is converted into React, then I think we can use the [numeral](http://numeraljs.com/#format-bytes) package to format these display labels.

psarando:
> thought this was a really good writeup/explanation that we might want to preserve in the Jira ticket if it's not already there

I didn't want to duplicate this in Jira, though, since that issue already auto-links to this PR 😅 
I did want to add these docs to our learning center docs, however, so I'll try to open a PR there that can be merged once this feature is released in prod.

de-lib/src/main/java/org/iplantc/de/apps/widgets/client/view/editors/AppStepResourcesViewImpl.java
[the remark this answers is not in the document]
psarando: The last commit of this PR adds these validation checks 👍

react-components/src/tools/ToolRestrictions.js
```diff
@@ -8,21 +8,20 @@ import {
...
 const NETWORK_MODES = ["bridge", "none"];
 
 const ONE_GB = 1024 * 1024 * 1024;
 
-function convertToGB(size) {
-    let gbValue = Math.round(size / ONE_GB);
-    return gbValue + " GB";
-}
+const formatGBListItem = (size) => size && numeral(size).format("0 ib");
+const formatGBValue = (size) => size && numeral(size).format("0.0 ib");
```
ash3rz: So simple! ✨ 
At least in the screenshots, the DE shows, for example, `GB` and Belphy will now show `GiB`.  Should we keep these units consistent to eliminate any confusion?  Not sure if one is actually calculating `GiB` and the other `GB`, or if the units are labeled incorrectly though.

psarando: Good point. They're both actually calculating `GiB`, so I should probably update the GXT code to use that unit label as well.

sri2k1us: Is there a reason behind using `GiB` format instead of `GB` ?

sri2k1us: Honestly I haven't seen `GiB` formatting commonly used. And We still have data window showing file sizes in `GB` as far as I know.

ianmcorvidae: Just because I was curious: the data window seems to also use the binary equivalents. My file `/home/mian/analyses/gzip__for_pipeline__giant_disk_analysis1-2019-01-30-01-20-34.2/logs/docker-compose.yml` shows a size of 2469 bytes, which in KB would be about 2.47 and in KiB would be about 2.41. The data window shows "2.41 KB" (which is to say, we're labeling KiB as KB).

ianmcorvidae: (of course, some stuff just labels the binary ones without the `i`, but it seems like if we're more specific it's better :smile: )

ash3rz: I've probably contributed to this in the DE 😊  I realize now I've had `*B` and `*iB` flipped in my head.

sri2k1us: Now am torn 🤔 should we change Data window to display `KiB` or update formatting here ...

psarando: It's the unit that k8s is using for setting memory resource requirements: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#meaning-of-memory

I imagine it's also the units that Condor is using as well.

The `numeral` package will only calculate the correct binary value if using this unit.
Using this example again: 4,000,000,000 bytes is 4GB, but only 3.7GiB.

sri2k1us: https://github.com/cyverse-de/ui/blob/f437d7c55d19e35c80924f63f7144834bbc1c7ba/de-lib/src/main/java/org/iplantc/de/client/util/DiskResourceUtil.java#L334 - @ianmcorvidae you are right!

sri2k1us: Sounds good to me then. We can fix Data window to show sizes `iB`

ianmcorvidae: Yeah, condor uses the binary versions. https://htcondor.readthedocs.io/en/v8_9_2/man-pages/condor_submit.html#submit-description-file-commands (it, mildly annoyingly, doesn't let you actually say MiB or similar, it just treats M and MB as MiB).

My experience is that most people are more used to using/prefer the binary units, and most of the time things labeled KB/MB/GB/etc. are actually KiB/MiB/GiB/etc. I sorta wish they'd changed the unit names the other way, but I do also understand that K/M/G/T/etc. have pre-existing meanings based on powers of 10. So my thought is that using the binary units and labeling them accurately is the clearest way we can communicate, even if it might mean some users have to look up why exactly that `i` is in the unit.

psarando:
Thanks for the reviews! 👍

PR #462, opened 2019-10-17
Convert all tables to use DETableRow and DECheckbox

ash3rz:
I updated some imports too since I was modifying the files anyway


ash3rz:
Thanks for the review!

PR #463, opened 2019-10-17
Add logic to convert search string to regex before 

sri2k1us:
passing to Highlighter.
- Updated few prop-types that were throwing validation errors in dev console.

### DIFF
--- a/react-components/package.json
+++ b/react-components/package.json
@@ -20,7 +20,7 @@
         "prop-types": "^15.7.2"
     },
     "dependencies": {
-        "@cyverse-de/ui-lib": "^0.3.13",
+        "@cyverse-de/ui-lib": "file:../../ui-lib/cyverse-de-ui-lib-0.3.12.tgz",
         "@material-ui/core": "^4.5.1",
         "@material-ui/icons": "^4.5.1",
         "@material-ui/lab": "^4.0.0-alpha.29",
--- a/react-components/package.json
+++ b/react-components/package.json
@@ -20,7 +20,7 @@
         "prop-types": "^15.7.2"
     },
     "dependencies": {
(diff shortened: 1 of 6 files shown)


react-components/src/apps/details/AppDetails.js
```diff
@@ -366,7 +366,10 @@ class AppDetails extends Component {
 
 AppDetails.propTypes = {
     details: PropTypes.object.isRequired,
-    searchText: PropTypes.string.isRequired,
+    searchText: PropTypes.oneOfType([
+        PropTypes.string,
+        PropTypes.instanceOf(RegExp),
```
ash3rz: 👍

react-components/src/apps/listing/AppGridListing.js
```diff
@@ -279,7 +280,7 @@ AppGridListing.defaultProps = {
 
 AppGridListing.propTypes = {
     parentId: PropTypes.string.isRequired,
-    apps: PropTypes.array.isRequired,
+    apps: PropTypes.array,
```
ash3rz: Did we not want this required?  I think in other components, I've initialized my props with an empty array instead of null in order to get around the warning (not sure if that's the case here, but thought I'd mention it).

react-components/package.json
```diff
@@ -20,7 +20,7 @@
         "prop-types": "^15.7.2"
     },
     "dependencies": {
-        "@cyverse-de/ui-lib": "^0.3.13",
+        "@cyverse-de/ui-lib": "file:../../ui-lib/cyverse-de-ui-lib-0.3.12.tgz",
```
ash3rz: Reminder to update before merging 😉

react-components/src/apps/listing/AppGridListing.js
sri2k1us: 🤔 The `apps` props in my case is a `splittable` and I had to initialize to null.

PR #464, opened 2019-10-21
core-608 Add help text for image name field.

sri2k1us:


react-components/src/tools/EditTool.js
```diff
@@ -218,6 +223,7 @@ function EditToolForm(props) {
                 parentId={parentId}
                 isOSGTool={isOSGTool}
                 component={ContainerImage}
+                intl={intl}
```
ash3rz: Super trivial, but I wonder if `intl` needs to be piped through all the parent components leading up to `ContainerImage` or if `ContainerImage` could be wrapped in `injectIntl` directly?

psarando: I think I'd prefer to wrap the components that need `intl` with `injectIntl`. It seems like that's the point of the HOC wrapper, and it would keep the props cleaner for the parent components. This is minor enough that we probably don't need to change it for now (maybe the next time this file needs an update).

sri2k1us: But I am ok with having each components individually wrapped as well.

PR #467, opened 2019-10-28
Fix weird transition to Boolean objects in Apps window

ash3rz:
I have no idea why converting the AppList splittable doesn't modify any booleans to Boolean objects while converting each App object in an array does in `onAppSelectionChanged`, but it seems to be working for me.

The other change is sending a shallow copy of an app via destructuring to the presenter whenever the presenter ends up modifying the app object. 

For the App Info dialog, there's a modification here: 
https://github.com/cyverse-de/ui/blob/master/de-lib/src/main/java/org/iplantc/de/client/services/impl/AppUserServiceFacadeImpl.java#L244

For app favorites, there's a modification here (though I'm not sure it's necessary since the window does a refresh of the listing afterwards anyway):
https://github.com/cyverse-de/ui/blob/master/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/list/AppsListPresenterImpl.java#L338

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/list/AppsListPresenterImpl.java
+++ b/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/list/AppsListPresenterImpl.java
@@ -24,6 +24,7 @@ import org.iplantc.de.client.models.UserInfo;
 import org.iplantc.de.client.models.apps.App;
 import org.iplantc.de.client.models.apps.AppAutoBeanFactory;
 import org.iplantc.de.client.models.apps.AppCategory;
+import org.iplantc.de.client.models.apps.AppList;
 import org.iplantc.de.client.models.apps.proxy.AppListLoadResult;
 import org.iplantc.de.client.models.avu.Avu;
 import org.iplantc.de.client.models.groups.Group;
@@ -178,11 +179,7 @@ public class AppsListPresenterImpl implements AppsListView.Presenter,
 
     @Override
     public void onAppSelectionChanged(Splittable splittableSelectedApps) {
-        selectedApps.clear();
-        for (int i = 0; i < splittableSelectedApps.size(); i++) {
-            App app = splittableToApp(splittableSelectedApps.get(i));
-            selectedApps.add(app);
-        }
+        selectedApps = AutoBeanCodex.decode(factory, AppList.class, splittableSelectedApps.getPayload()).as().getApps();
         AppSelectionChangedEvent event = new AppSelectionChangedEvent(selectedApps);
         fireEvent(event);
     }
--- a/react-components/src/apps/listing/AppListingView.js
+++ b/react-components/src/apps/listing/AppListingView.js
@@ -66,7 +66,7 @@ export default function AppListingView(props) {
     };
 
     const onAppInfoClick = (app) => {
-        presenter.onAppInfoSelected(app, false);
+        presenter.onAppInfoSelected({ ...app }, false);
     };
 
     const onCommentsClick = (app) => {
@@ -74,7 +74,7 @@ export default function AppListingView(props) {
     };
 
     const onFavoriteClick = (app) => {
-        presenter.onAppFavoriteSelected(app);
+        presenter.onAppFavoriteSelected({ ...app });
     };
 
     const onAppNameClick = (app) => {
@@ -90,7 +90,7 @@ export default function AppListingView(props) {
     };
 
     const onQuickLaunchClick = (app) => {
-        presenter.onAppInfoSelected(app, true);
+        presenter.onAppInfoSelected({ ...app }, true);
     };
 
     if (viewType === view.TILE) {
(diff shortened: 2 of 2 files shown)


sri2k1us:
If you don't I would like to investigate little bit more before we merge this.

sri2k1us:
Thanks for all you time :)

ash3rz:
Yeah, no problem!  I'll leave the branch and PR open for anyone to experiment with.

psarando:
> The other change is sending a shallow copy of an app via destructuring to the presenter whenever the presenter ends up modifying the app object.

This sounds correct to me. We should always treat GWT models used by our React code as immutable, especially if those objects are used in React state.

The tricky part will be finding where the GWT code is modifying a model object 😖 

So maybe we should always treat everything passed between GWT and React as immutable and always make a copy before transferring from one to the other? 🤔 Or continue to copy models on a case-by-case basis, as we find the bugs 🐛 

I thought encoding/decoding Autobeans to/from JSON meant it was copying the data from one to the other, but I guess this bug means that's not the case.

sri2k1us:
> For the App Info dialog, there's a modification here:
> https://github.com/cyverse-de/ui/blob/master/de-lib/src/main/java/org/iplantc/de/client/services/impl/AppUserServiceFacadeImpl.java#L244

I think the core of the issue here where we try to merge details into an existing `App`. So why does the details end-point only returns partial information ? 

> 
> For app favorites, there's a modification here (though I'm not sure it's necessary since the window does a refresh of the listing afterwards anyway):
> https://github.com/cyverse-de/ui/blob/master/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/list/AppsListPresenterImpl.java#L338

Thats correct. We don't need this code. I just cleaned this up this morning.

sri2k1us:
What I have doing is entirely avoiding Splittable to Model and Model to Splittable conversion in the first place. But I had to do it for since some parts of the code needs the model. The other thing I try to do is not pass the entire object to when calling presenter. Instead I try to send `id` and few other fields as required. This prevents modification all together. Hopefully, when other parts of Apps window is `react`fied, we can refactor a bit. 

> > The other change is sending a shallow copy of an app via destructuring to the presenter whenever the presenter ends up modifying the app object.
> 
> This sounds correct to me. We should always treat GWT models used by our React code as immutable, especially if those objects are used in React state.
> 
> The tricky part will be finding where the GWT code is modifying a model object 😖
> 
> So maybe we should always treat everything passed between GWT and React as immutable and always make a copy before transferring from one to the other? 🤔 Or continue to copy models on a case-by-case basis, as we find the bugs 🐛
> 
> I thought encoding/decoding Autobeans to/from JSON meant it was copying the data from one to the other, but I guess this bug means that's not the case.

psarando:
> I think the core of the issue here where we try to merge details into an existing `App`. So why does the details end-point only returns partial information ?

I originally thought that this code was updating the model from the [apps listing endpoint](https://de.cyverse.org/terrain/docs/index.html#!/apps/get_terrain_apps) with additional info returned by the [app details endpoint](https://de.cyverse.org/terrain/docs/index.html#!/apps/get_terrain_apps_system_id_app_id_details), but after reviewing their docs, I see that the `listing` endpoint does return info not included by the `details` endpoint, such as `pipeline_eligibility`, `is_favorite`, and `beta` flags.

I don't recall the limitations on why the `details` endpoint doesn't include this info, but after reviewing the `listing` endpoint's code, it could be simply because this extra info was requested in the listings for use by the UI, so this endpoint does some [extra lookups and formatting for listings](https://github.com/cyverse-de/apps/blob/ba8d236768198eb9e96e13eef4b3952b37e90d68/src/apps/service/apps/de/listings.clj#L298-L309) that weren't duplicated in the `details` endpoint, since it wasn't requested for that endpoint.

sri2k1us:
Thanks for looking into it 👍 

> > I think the core of the issue here where we try to merge details into an existing `App`. So why does the details end-point only returns partial information ?
> 
> I originally thought that this code was updating the model from the [apps listing endpoint](https://de.cyverse.org/terrain/docs/index.html#!/apps/get_terrain_apps) with additional info returned by the [app details endpoint](https://de.cyverse.org/terrain/docs/index.html#!/apps/get_terrain_apps_system_id_app_id_details), but after reviewing their docs, I see that the `listing` endpoint does return info not included by the `details` endpoint, such as `pipeline_eligibility`, `is_favorite`, and `beta` flags.
> 
> I don't recall the limitations on why the `details` endpoint doesn't include this info, but after reviewing the `listing` endpoint's code, it could be simply because this extra info was requested in the listings for use by the UI, so this endpoint does some [extra lookups and formatting for listings](https://github.com/cyverse-de/apps/blob/ba8d236768198eb9e96e13eef4b3952b37e90d68/src/apps/service/apps/de/listings.clj#L298-L309) that weren't duplicated in the `details` endpoint, since it wasn't requested for that endpoint.

ash3rz:
Since we have #468 going into a deeper dive for fixes, I'll close this PR

PR #468, opened 2019-10-29
Apps bugs

sri2k1us:
Overall, this is a much bigger change but the underlying logic for this big change is that these are changes that is needed eventually when reactify other parts of apps window. The additional `app` parameter that is passed around can be removed eventually. If you feel this is big change to do just before release, then I can cherry-pick `b5107d5` and `3bf2f3f` alone to prod which will fix most of the bugs.  We can merge #467 for now and then we can merge `8548c7b` after  release.


de-lib/src/main/java/org/iplantc/de/admin/desktop/client/services/impl/AppAdminServiceFacadeImpl.java
```diff
@@ -110,18 +111,12 @@ public void deleteApp(final HasQualifiedId app,
...
+        String address = APPS_ADMIN + "/" + systemId + "/" + appId + "/details";
         ServiceCallWrapper wrapper = new ServiceCallWrapper(GET, address);
-        deService.getServiceData(wrapper, new AsyncCallbackConverter<String, App>(callback) {
-            @Override
-            protected App convertFrom(String object) {
-                AutoBean<App> appAutoBean = AutoBeanCodex.decode(factory, App.class, object);
-                return appAutoBean.as();
-            }
-        });
+        deService.getServiceData(wrapper, new SplittableCallbackConverter(callback));
```
sri2k1us: Since Admin app details view already in react, I updated this service to return splittable.

de-lib/src/main/java/org/iplantc/de/apps/client/AppDetailsView.java
```diff
@@ -82,7 +82,11 @@ void useQuickLaunch(String quickLaunchId,
 
     }
 
-    void load(Presenter presenter);
+    void load(Presenter presenter,
+              final Splittable app,
+              final Splittable appDetails,
+              boolean showQuickLaunchFirst,
+              final String searchText);
```
sri2k1us: Instead of merging the details fetched from service into existing `App` model,  I have updated to code to work with an `App` splittable and separate `details` splittable.

de-lib/src/main/java/org/iplantc/de/apps/client/AppsListView.java
```diff
@@ -105,8 +109,11 @@
         void onAppNameSelected(Splittable appSplittable);
 
         @SuppressWarnings("unusable-by-js")
-        void onAppInfoSelected(Splittable appSplittable,
-                               boolean showQuickLaunchFirst);
+        void onAppInfoSelected(String appId,
+                               String systemId,
+                               boolean isPublic,
+                               boolean showQuickLaunchFirst,
+                               Splittable app);
```
sri2k1us: The required parameters for details to work properly are `appId`, `systemId` and `isPublic`. But we need to pass in the `app` splittable for now since the details view doesn't have access to selected `app` in react. Once all parts of the Apps window is in react, we can remove this parameter. Details view will get the selected `app` from the the state.

de-lib/src/main/java/org/iplantc/de/apps/client/events/selection/AppInfoSelectedEvent.java
```diff
@@ -21,23 +21,47 @@
     }
 
     public static final Type<AppInfoSelectedEventHandler> TYPE = new Type<>();
-    private final App app;
+    private final Splittable app;
```
sri2k1us: This parameter will be removed eventually as mentioned in my previous commit.

de-lib/src/main/java/org/iplantc/de/apps/client/presenter/list/AppsListPresenterImpl.java
```diff
@@ -277,6 +281,7 @@ protected void getAppsWithSelectedCategory() {
     public void onOntologyHierarchySelectionChanged(OntologyHierarchySelectionChangedEvent event) {
         listView.setLoadingMask(true);
         selectedHierarchy = event.getSelectedHierarchy();
+        appCategory = null;
```
sri2k1us: resetting `selectedHierarchy` when `appCategory` is selected and vice-versa fixes refresh logic in most cases. I think there is still issue when your search and refresh...

de-lib/src/main/java/org/iplantc/de/apps/client/presenter/list/AppsListPresenterImpl.java
```diff
@@ -335,9 +340,12 @@ public void onFailure(Integer statusCode, Throwable caught) {
...
             @Override
             public void onSuccess(Void result) {
-                app.setFavorite(!app.isFavorite());
-                eventBus.fireEvent(new AppFavoritedEvent(app));
                 eventBus.fireEvent(new AppUpdatedEvent(app));
+                if (!app.isFavorite()) {
+                    announcer.schedule(new SuccessAnnouncementConfig(appearance.favSuccess(app.getName())));
+                } else {
+                    announcer.schedule(new SuccessAnnouncementConfig(appearance.unFavSuccess(app.getName())));
+                }
```
sri2k1us: There was no feedback to user when an app was marked or removed as favorites. Added user feedback.

react-components/src/apps/listing/AppGridListing.js
```diff
@@ -250,7 +249,7 @@ function getTableColumns(deletable, enableMenu) {
         },
         {
             name: AppFields.SYSTEM.fieldName,
-            enableSorting: true,
+            enableSorting: false,
```
sri2k1us: disabled sorting by `system ID` field since there is a bug in app listing service when sorting using this field.

react-components/src/apps/listing/AppListingView.js
```diff
@@ -82,15 +88,21 @@ export default function AppListingView(props) {
     };
 
     const onRatingClick = (event, app, score) => {
-        presenter.onAppRatingSelected(app, score);
+        presenter.onAppRatingSelected(app, Math.ceil(score));
```
sri2k1us: Service accepts only int for rating score.

react-components/src/apps/listing/AppListingView.js
```diff
@@ -130,6 +142,7 @@ export default function AppListingView(props) {
                 handleSelectAll={handleSelectAll}
                 onFavoriteClick={onFavoriteClick}
                 onQuickLaunchClick={onQuickLaunchClick}
+                onCommentsClick={onCommentsClick}
```
sri2k1us: Added missing callback prop for Comments.

react-components/src/apps/listing/AppTileListing.js
```diff
@@ -92,7 +91,7 @@ function AppTileListing(props) {
                                         name={app.name}
                                         creator={app.integrator_name}
                                         rating={app.rating}
-                                        type={app.app_type}
+                                        type={app.system_id}
```
sri2k1us: Fixed type display discrepancy between grid and tile view.

de-lib/src/main/java/org/iplantc/de/client/services/impl/AppUserServiceFacadeImpl.java
```diff
@@ -230,25 +230,19 @@ protected Void convertFrom(String object) {
...
 
-        deServiceFacade.getServiceData(wrapper, new DECallbackConverter<String, App>(callback) {
-            @Override
-            protected App convertFrom(String object) {
-                Splittable split = StringQuoter.split(object);
-                AutoBean<App> appAutoBean = AutoBeanUtils.getAutoBean(app);
-                AutoBeanCodex.decodeInto(split, appAutoBean);
-                return appAutoBean.as();
-            }
-        });
```
sri2k1us: As @ash3rz figured, this is the place where we trying to mutate the selected `app`. Removed this code and modified all it return a splittable.

react-components/src/apps/admin/AdminAppDetails.js
```diff
@@ -312,6 +314,6 @@ AdminAppDetailsDialog.propTypes = {
 
 export default withFormik({
     enableReinitialize: true,
-    mapPropsToValues: ({ app }) => ({ ...app }),
+    mapPropsToValues: ({ details }) => ({ ...details }),
```
ash3rz: I think this will break a few things in the AdminAppDetails view.  `details` doesn't have the `beta` key, for example, and the form is looking for that.  Also, in the `handleSubmit` method I added a lot of logic checking the `props` (initial) values vs the form's (updated) `values` to try and minimize the number of calls we were making when submitting the form.

We can probably do this:
```
mapPropsToValues: ({ app, details }) => ({ ...app, ...details }),
```
and then update the `handleSubmit` method to check the right `props` (now it will have to check `props.details` in some spots, instead of `props.app`).

de-lib/src/main/java/org/iplantc/de/apps/client/presenter/submit/SubmitAppForPublicPresenter.java
```diff
@@ -188,26 +188,28 @@ PublishAppRequest getPublishAppRequest() {
...
+                view.loadReferences(parseRefLinks(result.get("references")));
             }
         });
     }
 
-    private List<AppRefLink> parseRefLinks(List<String> arr) {
+    private List<AppRefLink> parseRefLinks(Splittable arr) {
         List<AppRefLink> linksList = Lists.newArrayList();
-        for (String ref : arr) {
+       for (int i = 0; i < arr.size(); i++)  {
```
ash3rz: Minor, but spacing is misaligned

de-lib/src/main/java/org/iplantc/de/apps/client/presenter/list/AppsListPresenterImpl.java
```diff
@@ -246,7 +251,6 @@ protected void disableTypeFilterForHPC() {
             }
         }
         listView.setTypeFilter(filter);
```
ash3rz: Just picking a random spot in this file to comment - in the PR I had open, I also had to modify `onAppSelectionChanged`.  Is it no longer needed?  It may have been fixed in a different spot in a different way, so I may have missed it.

react-components/src/apps/admin/AdminAppDetails.js
sri2k1us: Good catch...i had my doubts when modifying this. Will add a fix for this.

de-lib/src/main/java/org/iplantc/de/apps/client/presenter/list/AppsListPresenterImpl.java
sri2k1us: I believe you are talking about https://github.com/cyverse-de/ui/blob/8548c7b3238e7457da0ac6708a5d09706d785387/de-lib/src/main/java/org/iplantc/de/apps/client/presenter/list/AppsListPresenterImpl.java#L187 ? That fix is no longer required. But I will double check it.

ash3rz: Yup, that's it.  Really, as long as the props don't have Boolean objects anymore, then I think this is fixed. This was a weird spot where things were getting flipped again for some reason.

react-components/src/apps/listing/AppGridListing.js
```diff
@@ -102,6 +102,10 @@ class AppGridListing extends Component {
                             const selected = isSelected(app.id);
                             const external = app.app_type !== appType.de;
                             const rowId = build(parentId, app.id);
+                            const disabled =
+                                app.isDisabled instanceof Boolean
```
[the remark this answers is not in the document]
sri2k1us: Nope. I didn't.

de-lib/src/main/java/org/iplantc/de/apps/client/presenter/list/AppsListPresenterImpl.java
ash3rz: Awesome - I'll close my PR then 👍

sri2k1us:
Thanks for the reviews.

PR #469, opened 2019-11-07
Core 634

sri2k1us:
Port tools info dialog to react.

image

image

image

image


react-components/src/tools/details/ToolInfoDialog.js
```diff
@@ -0,0 +1,227 @@
...
+                                selectable={false}
+                                deletable={false}
+                                enableMenu={false}
+                                isSelected={() => false}
+                                selectedApps={[]}
+                                onSortChange={() => {}}
+                                resetAppSelection={() => {}}
+                                handleAppSelection={() => {}}
+                            />
+                        )}
```
sri2k1us: Not sure if there is a different way to avoid errors in console and at the same time doing nothing on certain callbacks.

react-components/src/tools/manage/ManageTools.js
```diff
@@ -425,9 +440,10 @@ function ToolListing(props) {
                         >
                             <TableCell padding="none">
                                 <IconButton
-                                    onClick={() =>
-                                        presenter.onShowToolInfo(tool.id)
-                                    }
+                                    onClick={(event) => {
+                                        event.stopPropagation();
```
sri2k1us: This one tricky. Without this, every second click on the  same`info` will fail. See `onToolSelection` where selection is set to `null` when clicked the second time.

react-components/stories/apps/details/AppInfoDialog.stories.js
```diff
@@ -357,7 +357,7 @@ class AppInfoDialogTest extends Component {
                 baseDebugId="appDetails"
                 searchText={text("Search Pattern", "roger")}
                 app={app}
-                appDetails={appDetals}
```
sri2k1us: You can get all the `detals` about this typo here

react-components/src/tools/manage/ManageTools.js
sri2k1us: Just found another bug with this. Working on a fix....

sri2k1us: I went back and removed the logic that toggles selection on click. Since we don't have multi selection checkbox, I don't think we need it.

sri2k1us: This is ready for review now.

react-components/src/tools/details/ToolInfoDialog.js
```diff
@@ -0,0 +1,227 @@
...
+                                        <span className={labelClass}>
+                                            {getMessage("descriptionLabel")}:
+                                        </span>
+                                        <span className={valueClass}>
+                                            {selectedTool.description}
+                                        </span>
+                                    </Grid>
+                                    <Grid item xs={12}>
+                                        <Typography variant="h6">
+                                            {getMessage("restrictionsLabel")}
```
psarando: I know the original tool info dialog only included the attribution, description, and a few restrictions, but while we're making a new info dialog, I'd like to add more info to it.

It would be nice to be able to see more of the tool's container restrictions:
```javascript
tool.container.max_cpu_cores
tool.container.min_cpu_cores
tool.container.min_memory_limit
tool.container.min_disk_space
```
These could be helpful to an app integrator that needs to set some default/hidden app params, such as number of threads/cpu-cores.

Some of the tool's docker image info is already displayed in the listing grid, but it could be helpful to have it also available in this info dialog, especially for a user that might want to create their own app using a tool they don't have permission to edit (or for copy-paste into their own private tool for tweaking):
```javascript
tool.container.image.name
tool.container.image.tag
tool.container.image.url
tool.container.image.deprecated
tool.container.entrypoint
tool.container.uid
tool.container.working_directory
tool.version
```
These could also be added in a later PR.

react-components/src/tools/details/ToolInfoDialog.js
```diff
@@ -0,0 +1,227 @@
...
+                            />
+                        </Tabs>
+                        {tabIndex === 0 && tool && (
+                            <Paper id={baseDebugId} style={{ padding: 5 }}>
+                                <Grid
+                                    container
+                                    spacing={2}
+                                    style={{ paddingLeft: 5 }}
+                                >
+                                    <Grid item xs={12}>
```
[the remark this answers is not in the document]
sri2k1us: Removing `xs` value messes up the formatting.
image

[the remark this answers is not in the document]
psarando: 😖 I must be remembering something else, then.

react-components/stories/tools/ToolInfoDialog.stories.js
```diff
@@ -0,0 +1,186 @@
...
+                },
+            },
+            version: "0.0.3",
+            implementation: {
+                implementor: "Upendra Kumar Devisetty",
+                implementor_email: "upendra@cyverse.org",
+            },
+            time_limit_seconds: 0,
+        };
+        const toolInfo = {
```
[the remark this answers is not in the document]
sri2k1us: I wanted to check how a private tool vs a public tool rendered.

sri2k1us:
Updated screen shots. All the details that @psarando mentioned are added to details.

react-components/src/apps/listing/AppGridListing.js
```diff
@@ -221,7 +226,9 @@ class AppGridListing extends Component {
                     baseId={parentId}
                     ids={ids.FIELD}
                     columnData={columnData}
-                    onRequestSort={this.onRequestSort}
+                    onRequestSort={
+                        onSortChange ? this.onRequestSort : undefined
```
psarando: 👍

sri2k1us:
Thanks for the reviews.

PR #471, opened 2019-11-19
CORE-633 Update AppStepResourcesView editor with resource defaults.

psarando:
This PR will update the `AppStepResourcesView` editor to init the resource requirement fields with any defaults set in the resource limits model.

Currently only the relaunch endpoint includes default settings, added by cyverse-de/apps#196.


psarando:
Thanks for the review!

PR #475, opened 2020-01-14
Belphegor AllStats 

flynn1999:
Let me know if the comments are useful and understandable.

**Screenshots:**
Screen Shot 2020-01-14 at 2 25 36 PM
Screen Shot 2020-01-14 at 2 25 42 PM
Screen Shot 2020-01-14 at 2 25 51 PM
Screen Shot 2020-01-14 at 2 26 37 PM


react-components/src/apps/admin/AllStats/AllStats.js
```diff
@@ -0,0 +1,57 @@
...
+                            label={getMessage("startDate")}
+                            id={build(ids.MAIN_PAGE, ids.START_DATE)}
+                            className={classes.datePicker}
+                        />
+                        <DatePicker
+                            label={getMessage("endDate")}
+                            id={build(ids.MAIN_PAGE, ids.END_DATE)}
+                            className={classes.datePicker}
+                        />
+                        <ColorButton
```
ash3rz: Since this button uses the `primary` color (which is CyVerse blue), I don't think you need to use `ColorButton` at all, which seems to be trying to apply purple colors?  I think you could just stick with the regular material-ui `TextButton` and remove the ColorButton.js file.

react-components/src/apps/admin/AllStats/AppStatsComponents/NumberTextfield.js
```diff
@@ -0,0 +1,41 @@
...
+        marginRight: theme.spacing(1),
+        width: 200,
+    },
+}));
+
+function NumberTextfield() {
+    const classes = useStyles();
+
+    return (
+        <div>
```
ash3rz: I don't think you need this extra `div`

react-components/src/apps/admin/AllStats/AppStatsComponents/NumberTextfield.js
```diff
@@ -0,0 +1,41 @@
...
+            <form className={classes.container} noValidate autoComplete="off">
+                <TextField
+                    id="filled-number"
```
ash3rz: Let's build a static ID here like you've done in other files

react-components/src/apps/admin/AllStats/AppStatsComponents/AppsTable.js
```diff
@@ -0,0 +1,73 @@
...
+import ids from "./AllStatsIDs";
+import { getMessage, withI18N } from "@cyverse-de/ui-lib";
+import myMessagesFile from "./messages";
+import build from "@cyverse-de/ui-lib/src/util/DebugIDUtil";
+
+const rows = appsTableData.apps;
+
+function AppsTab() {
+    return (
+        <Paper
```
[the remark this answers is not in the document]
flynn1999: The Paper component is used just to layer various components on top of it. It basically acts as a blank canvas.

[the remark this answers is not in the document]
psarando: It also has styling that outlines the canvas with a subtle drop-shadow, so that the container looks like an actual piece of paper. I'm not sure, but I think there's a little shadow outlining the bottom of the table that may be coming from this wrapper component (unless the table has its own `paper` wrapper that is displaying that).

react-components/src/apps/admin/AllStats/AppStatsComponents/NavBarTabs.js
```diff
@@ -0,0 +1,111 @@
...
+import Box from "@material-ui/core/Box";
+import JobsTab from "./JobsTab";
+import AppsTab from "./AppsTab";
+import UsersTab from "./UsersTab";
+import ids from "./AllStatsIDs";
+import { getMessage, withI18N } from "@cyverse-de/ui-lib";
+import myMessagesFile from "./messages.js";
+import build from "@cyverse-de/ui-lib/src/util/DebugIDUtil";
+
+function TabPanel(props) {
```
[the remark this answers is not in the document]
psarando: Although there are some nice things about the example code. I do like `a11yProps` (though we may want to `build` a debug ID in there).

I also wonder if this version might be more performant than our simple version, where we change the contents that are rendered based on the `tabIndex` 🆚 this version that renders all the contents and hides all of them except the currently selected index. It's possible that our old version could thrash the DOM more. 🤔

That said, I think we don't need the `LinkTab` components that imitate links (we don't have routing that allows the user to navigate our app with the browser's back/forward buttons). So maybe stick with the [Simple Tabs](https://material-ui.com/components/tabs/#simple-tabs) example.

react-components/src/apps/admin/AllStats/AppStatsComponents/AppsTab.js
```diff
@@ -0,0 +1,38 @@
...
+import NumberTextfield from "./NumberTextfield";
+import styles from "../AllStatsStyle";
+import { withStyles } from "@material-ui/core";
+import ids from "./AllStatsIDs";
+import messages from "./messages";
+import { withI18N, getMessage, build } from "@cyverse-de/ui-lib";
+
+function AppTab(props) {
+    const { classes } = props;
+    const { baseId } = props.id;
```
psarando: This [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) doesn't look quite right. I think the `props` object looks something like `{classes: {...}, id: "..."}`, in which case these 2 lines could be combined to destructure both `classes` and `id` at once:
```javascript
     const { classes, id } = props;
```
And if you really want to name the variable `baseId`, then:
```javascript
     const { classes, id: baseId } = props;
```

react-components/src/apps/admin/AllStats/AppStatsComponents/AppsTab.js
```diff
@@ -0,0 +1,38 @@
...
+
+    return (
+        <div id={baseId}>
+            <div
+                className={classes.appSelectBar}
+                id={build(props.id, ids.HEADER)}
```
psarando: This could also be `baseId` instead of `props.id`.

react-components/src/apps/admin/AllStats/AppStatsComponents/UsersTab.js
```diff
@@ -0,0 +1,45 @@
...
+import Paper from "@material-ui/core/Paper";
+import usersTableData from "./dataFiles/usersData";
+import ids from "./AllStatsIDs";
+import { getMessage, withI18N, build } from "@cyverse-de/ui-lib";
+import myMessagesFile from "./messages";
+
+const rows = usersTableData.users;
+
+function UsersTab(props) {
+    const { baseId } = props.id;
```
psarando: See my destructuring comment in the `AppsTab` file.

react-components/src/apps/admin/AllStats/AppStatsComponents/NavBarTabs.js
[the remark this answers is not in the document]
ash3rz: 🤔 good point, it might actually be more performant which would explain the change in their docs.  Agree with sticking with `Simple Tabs` and getting rid of `LinkTab`.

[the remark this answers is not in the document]
flynn1999: @psarando What is a debug ID?

react-components/src/apps/admin/AllStats/AppStatsComponents/dataFiles/jobsData.js
```diff
@@ -0,0 +1,67 @@
+const jobsTableData = {
+    count: 2,
+    jobs: [
+        {
+            Category: "DE",
```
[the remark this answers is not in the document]
flynn1999: Yup double checked it. That's the way it is.

react-components/src/apps/admin/AllStats/AppStatsComponents/NavBarTabs.js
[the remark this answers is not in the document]
psarando: Oh, sorry, that's what they are called in GWT. I just meant a static ID, but dynamically built with the ui-lib's DebugIDUtil `build` function.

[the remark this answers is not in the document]
psarando: So `a11yProps` could return something like `id: build(baseId, ids.NAV_TAB, index)`.

react-components/src/apps/admin/AllStats/AllStatsStyle.js
```diff
@@ -0,0 +1,56 @@
...
+
+const styles = (theme) =>
+    createStyles({
+        datePicker: {
+            float: "left",
+            marginRight: 40,
+        },
+
+        datePickers: {
+            borderBottom: "solid 60px",
```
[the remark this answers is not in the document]
flynn1999: Those border styles define the "grey header" of the page.

react-components/src/apps/admin/AllStats/AppStatsComponents/NavBarTabs.js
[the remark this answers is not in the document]
flynn1999: I've made the necessary changes.

react-components/src/apps/admin/AllStats/AppStatsComponents/NavBarTabs.js
```diff
@@ -0,0 +1,91 @@
...
+
+function NavTabs(props) {
+    const classes = props;
+    const [value, setValue] = React.useState(0);
+
+    const handleChange = (event, newValue) => {
+        setValue(newValue);
+    };
+
+    function a11yProps(index) {
```
psarando: Ah, but I still think it's worthwhile to keep the `aria-controls` key and value in the return. I think its value could be the same (it doesn't need a `baseId`).

Also, it looks like `baseId` is now used for the `id` value 👍, but `baseId` isn't getting passed into this function; only `index` is getting passed in.

psarando: I would recommend uncommenting out the original function, but updating it to also accept a `baseId` parameter, and update its `id` field similar to this one.

flynn1999: What is the use of `aria-controls`?
(Also I've made the necessary changes)

psarando: Already discussed in Slack, but also posting here for future reference: https://www.w3.org/TR/wai-aria/

react-components/src/apps/admin/AllStats/AppStatsComponents/datePicker.js
```diff
@@ -0,0 +1,41 @@
...
+} from "@material-ui/pickers";
+import DateFnsUtils from "@date-io/date-fns";
+import { formatDateObject } from "@cyverse-de/ui-lib/src/util/DateFormatter";
+
+function DatePicker(props) {
+    const [selectedDate, setSelectedDate] = React.useState(new Date());
+    const handleDateChange = (date) => {
+        let formattedDate = formatDateObject(date, "YYYY-MM-DD");
+        console.log(formattedDate);
+        setSelectedDate(date);
```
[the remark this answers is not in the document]
psarando: One more refactor is required that will allow a date change to trigger an API call, which will reload these components with new data. Just as the story is now passing in mock data for these components to render, the story should also pass down a mock API callback function that can be used when the date changes.

In our current design pattern, we are wrapping API calls in a `presenter` object that is also passed into the view components, just as the data is now passed in.

Take a look at the example in the Teams story:
https://github.com/cyverse-de/ui/blob/5778ad4aa31b03168e65f92fb26dcea18c935a8b/react-components/stories/teams/Teams.stories.js#L14-L23

This mock presenter is defining a few mock callback functions. You can create a simple API callback mock similar to this `onTeamNameSelected` method. Your story's mock callback will replace this `handleDateChange` and it can format and log the date just like this one.

I'm also thinking your story component will also have to manage the state of the selected dates, and the dates will have to be passed down to these components, just as the mock data is currently passed down.

react-components/src/apps/admin/AllStats/AppStatsComponents/JobsTab.js
```diff
@@ -0,0 +1,71 @@
...
+                        <TableCell>
+                            {formatMessage(props.intl, "distinctLogins")}
+                        </TableCell>
+                        <TableCell align="center">
+                            {distinctLoginData.count}
+                        </TableCell>
+                    </TableRow>
+                    <TableRow>
+                        <TableCell>
+                            {formatMessage(props.intl, "totalLogins")}
```
ash3rz: Just a minor thing, but I don't think `formatMessage` is necessary in a table - `getMessage` should work just like in the TableCell a few rows up.

react-components/src/apps/admin/AllStats/AppStatsComponents/NavBarTabs.js
```diff
@@ -0,0 +1,92 @@
...
+TabPanel.propTypes = {
+    children: PropTypes.node,
+    index: PropTypes.any.isRequired,
+    value: PropTypes.any.isRequired,
+};
+
+function a11yProps(baseId, index) {
+    return {
+        id: build(baseId, index),
+        "aria-controls": `nav-tabpanel-${index}`,
```
ash3rz: Minor bug here with [aria-controls](https://www.w3.org/WAI/PF/aria-1.1/states_and_properties#aria-controls).  Basically `aria-controls` needs to match up with the `id` of the element it controls, and in your case they're mismatched.  `aria-controls` will have `nav-tabpanel-jobsTab`, while the element ID in `TabPanel` will be `nav-tabpanel-0`.  In other words, in `a11yProps`, `index` is not actually an index, it's part of the static ID from your IDs file, while in `TabPanel` `index` is actually an index.

react-components/src/apps/admin/AllStats/AppStatsComponents/NumberTextfield.js
```diff
@@ -0,0 +1,42 @@
...
+
+function NumberTextfield(props) {
+    const classes = useStyles();
+    const { baseId } = props;
+
+    return (
+        <form className={classes.container} noValidate autoComplete="off">
+            <TextField
+                id={baseId}
+                label="Number"
```
ash3rz: This label should come from your messages file

react-components/stories/apps/admin/AllStats/AllStats.stories.js
```diff
@@ -0,0 +1,21 @@
...
+import usersData from "./dataFiles/usersData";
+
+class AllStatsTest extends Component {
+    render() {
+        return (
+            <AllStats
+                appsData={appsTableData.apps}
+                distinctLoginData={distinctLoginData}
+                jobsData={jobsData.jobs}
+                usersData={usersData.users}
```
ash3rz: 👍

react-components/src/apps/admin/AllStats/AppStatsComponents/datePicker.js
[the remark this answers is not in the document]
ash3rz: Agreed.  Although I'm not 100% sure how different this will look in sonora, this is still a good pattern to follow to create the mental model of that flow between the view and the API.

react-components/src/apps/admin/AllStats/AppStatsComponents/NavBarTabs.js
psarando: Good catch. I was assuming this was the same as an `aria-label` and could be any descriptive text.

react-components/src/apps/admin/AllStats/AppStatsComponents/JobsTab.js
psarando: Right, since `getMessage` will return a React component that renders as the message, and `formatMessage` only returns a string, which is required in places that can't render a component, such as `label` properties.

The `getMessage` call looks a bit cleaner and doesn't require the `intl` param, so we generally prefer that where we can use it.

flynn1999:
I've made the changes for the date change. I'll finish up with the styling issues tomorrow. 

Also I had a styling suggestion. @psarando @ash3rz 

I was thinking of adding a space between `=` in all instances. 
For example:
Changing this
`presenter={presenter}`
to this
`presenter = {presenter}`

I think it makes things a lot more neat and easier to read. What are your thoughts on it?

ash3rz:
> Also I had a styling suggestion. @psarando @ash3rz
> 
> I was thinking of adding a space between `=` in all instances.
> For example:
> Changing this
> `presenter={presenter}`
> to this
> `presenter = {presenter}`
> 
> I think it makes things a lot more neat and easier to read. What are your thoughts on it?

Personally, I find it harder to read and I feel like it creates whitespace around something that's a simple assignment/attribute rather than complicated logic where whitespace might make things easier to parse.
Also, if you check out the `package.json` file in any of our 3 UI libs, you'll find that we have tools (eslint and prettier) to handle a lot of common code formatting issues/opinions upon commit.  I'm guessing if you've tried to implement this style and commit it as a change, you've found that your commit actually has the spacing removed. These tools are great because they enforce a level of consistency without requiring a dev to do it manually themselves, which in the end makes code easier to read and PRs easier to review.
That being said, I've not personally seen any other sample code or repo that has spacing around prop assignments.  I think most are following the conventions shown in the official React docs/tutorials.

react-components/stories/apps/admin/AllStats/AllStats.stories.js
```diff
@@ -4,18 +4,38 @@ import appsTableData from "./dataFiles/appsData";
...
+    const presenter = {
+        handleStartDateChange: (date) => {
+            let formattedDate = formatDateObject(date, "YYYY-MM-DD");
+            console.log(formattedDate);
+            setSelectedStartDate(formattedDate);
+        },
+        handleEndDateChange: (date) => {
+            let formattedDate = formatDateObject(date, "YYYY-MM-DD");
+            console.log(formattedDate);
+            setSelectedEndDate(formattedDate);
```
psarando: I think these `setSelectedStartDate` and `setSelectedEndDate` functions should be setting the new `date` object, rather than the `formattedDate` string.

I'm guessing that's what `KeyboardDatePicker` is expecting, and that's also how the original `handleDateChange` function worked in `datePicker`:
```javascript
    const [selectedDate, setSelectedDate] = React.useState(new Date());
    const handleDateChange = (date) => {
        let formattedDate = formatDateObject(date, "YYYY-MM-DD");
        console.log(formattedDate);
        setSelectedDate(date);
    };
```

react-components/src/apps/admin/AllStats/AppStatsComponents/datePicker.js
```diff
@@ -38,4 +36,13 @@ function DatePicker(props) {
     );
 }
 
+DatePicker.propTypes = {
+    presenter: PropTypes.shape({
+        handleStartDateChange: PropTypes.func.isRequired,
+        handleEndDateChange: PropTypes.func.isRequired,
+    }),
+    selectedStartDate: PropTypes.string.isRequired,
+    selectedEndDate: PropTypes.string.isRequired,
```
psarando: These `selectedStartDate` and `selectedEndDate` props should probably be required date objects, rather than required strings.

react-components/stories/apps/admin/AllStats/AllStats.stories.js
flynn1999: You're right. `formattedDate` was the one that is supposed to be sent to Jack's code and date is the one that `KeyboardDatePicker` receives. I'll make the necessary changes. Thanks for noticing this 👍

react-components/src/apps/admin/AllStats/AppStatsComponents/datePicker.js
```diff
@@ -0,0 +1,48 @@
...
+                    KeyboardButtonProps={{
+                        "aria-label": "change date",
+                    }}
+                />
+            </MuiPickersUtilsProvider>
+        </div>
+    );
+}
+
+DatePicker.propTypes = {
```
psarando: Sorry I didn't notice this before, but this `DatePicker.propTypes` is declaring that `DatePicker` requires the `presenter` and selected date props. The actual `DatePicker` function above is only using `dateChange` and `selectedDate` from its `props` though.

It looks like these Prop Types should actually be assigned to `AllStats.propTypes`.

flynn1999: How come it didn't show any errors when it was DatePicker.propTypes?

psarando: There should have been some warnings in the console about failed prop types.

I also seem to remember that when you're running Storybook, sometimes warnings will only display once at the start, but not again after code changes and the page auto-reloads. I think you can get the warnings to display again if you manually refresh the browser tab, or by restarting Storybook.

react-components/src/apps/admin/AllStats/AppStatsComponents/NavBarTabs.js
flynn1999: @ash3rz I had resolved this because I had fixed this by changing the function to this 
`function a11yProps(baseId, index) {
    return {
        id: build(baseId, index),
        "aria-controls": build(baseId, index),
    };
}`

I'm pretty sure that is not how I am supposed to do it but using this `aria-controls` matches with the `id`.
Let me know what I can do to fix this

ash3rz: Yes, you have `aria-controls` matching the id of the current element, rather than the element it controls, the `Typography` within `TabPanel`.  If you double check the sample code from material-ui's Tab demo, you'll see what needs to match between the different `id`, `aria-labelledby`, and `aria-controls` props.

flynn1999: Okay, I've made some changes that match the pattern shown in material ui's tab demo.

react-components/src/apps/admin/AllStats/AppStatsComponents/JobsTab.js
ash3rz: Now that `formatMessage` is gone, you can also remove the `formatMessage` and `injectIntl` imports.

react-components/src/apps/admin/AllStats/AppStatsComponents/NavBarTabs.js
ash3rz: Still not quite 😄 In MUI's example, you have basically 2 elements with their own distinct static IDs - the tab and the tab panel. Based on those 2 distinct IDs, you establish the `aria-controls` and `aria-labelledby` relationship between those 2 elements.
Basically, you need 
```
...
<Typography
  id=a
  aria-labelledby=b
  ...
/>
...
function a11yProps(baseId, index) {
  return {
    id: b
    aria-controls: a
  }
}
```

react-components/src/apps/admin/AllStats/AppStatsComponents/NavBarTabs.js
```diff
@@ -0,0 +1,92 @@
...
+TabPanel.propTypes = {
+    children: PropTypes.node,
+    index: PropTypes.any.isRequired,
+    value: PropTypes.any.isRequired,
+};
+
+function a11yProps(baseId, index) {
+    return {
+        id: build(baseId, index),
+        "aria-controls": `${baseId}-nav-tabpanel-${index}`,
```
ash3rz: Starting a new thread because it was getting harder to find the old one 😄 
This is still not quite right. In their example, there are only 2 IDs that they're dealing with: 'simple-**tabpanel**-${index}' and 'simple-**tab**-${index}'. And what you need to do is get them to match between your **tab** and **tabpanel** code.  At this stage, you have 3 different IDs, which is incorrect.  

Here's what I'm talking about from MUI's example to show what's supposed to match:
image

Since you're not using an actual index, I would rename your variables because maybe that's also causing confusion. 
Ultimately, you just need something like `build(baseId, tabName)` and either `build(baseId, panelName)` **or** `build(baseId, tabName, "panel")` if you want to do something similar to their example. Please let me know if this is still not making sense.

PR #477, opened 2020-01-30
CORE-879 Add admin config settings for max tool resource requests.

psarando:
This PR will add admin config settings for max tool resource requests, and update the `EditTool` view to use these max admin settings when building tool container resource limit drop-down lists in the admin view.


psarando:
Thanks for the reviews! 👍

PR #478, opened 2020-02-05
CORE-877 Add mutli-select relaunch support to AnalysesView

psarando:
This PR will add mutli-select relaunch support to the AnalysesView, allowing users to select more than 1 analysis to relaunch with the new analyses auto-relaunch endpoint (cyverse-de/apps#197).

This PR will also add range-selection support to the AnalysesView.

### DIFF
--- a/react-components/src/analysis/messages.js
+++ b/react-components/src/analysis/messages.js
@@ -63,6 +63,12 @@ var intlData = {
             " to extend the time limit?",
         analysesExecDeleteWarning:
             "This will remove the selected analyses and the parameters information associated with those analyses. Outputs can still be viewed in the Data window within the folder created by these analyses.",
+        analysesMultiRelaunchWarning:
+            "Relaunching more than one analysis at once will relaunch all selected analyses, reusing their original parameters and analysis names." +
+            " If the selected analyses are sub-jobs of an HT analysis, then those selected analyses will still be nested under that parent HT analysis," +
+            " and their output folders will also be grouped under that parent HT analysis' output folder," +
+            " but the relaunched sub-jobs will be renamed with a `-redo-#` suffix to differentiate them from their original sub-jobs." +
+            " Otherwise, relaunched analyses will be treated as new analyses, even though they reuse the same name and parameters as their original analyses.",
 
         shareDisclaimer:
             '<span style="overflow:auto;">I agree to <a href="https://wiki.cyverse.org/wiki/x/KBfHAQ" target="_blank">share my analysis</a> details, output file(s) and logs with CyVerse Support. If you want any of the <a href="https://wiki.cyverse.org/wiki/x/XRfHAQ" target="_blank">Science Informaticians</a> to recuse themselves from assisting with this problem, please list their name(s) in comments.</span>',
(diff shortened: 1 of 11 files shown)


react-components/src/analysis/view/AnalysesView.js
```diff
@@ -58,6 +58,20 @@ import analysesCollapseIcon from "../../resources/images/analyses-collapseList.s
...
+    if (deSelect) {
+        // deselect range
+        return selected.filter(
+            (selectedID) => !newSelectionRange.includes(selectedID)
+        );
+    }
+
+    // append range, removing duplicates
+    return [...new Set([...selected, ...newSelectionRange])];
+};
```
psarando: This `rangeSelect` logic probably didn't have to be pulled out into its own function, but I extracted it in case we wanted something general like this in `ui-lib` for other listings to use. If we do use it outside this view, it could probably use some null checks for `data` and `selected`.

react-components/src/analysis/view/AnalysesView.js
```diff
@@ -584,35 +599,59 @@ class AnalysesView extends Component {
...
-                );
+            if (selectRange && lastSelectedIndex !== index) {
+                const allIDs = data.map((analysis) => analysis.id);
+
+                newSelected =
+                    lastSelectedIndex < index
+                        ? rangeSelect(
+                              allIDs,
+                              selected,
+                              isSelected,
```
psarando: I'm adding or removing everything in the range solely based on whether the last item in the range was already selected or not. 

In other words, if you shift+click on an item that wasn't already selected, then everything in the range becomes selected, even if some of them already were selected. If the last item was already selected, then everything in the range becomes deselected, even if only one other item in the range was selected.

This is also how Gmail behaves.

react-components/src/analysis/view/AnalysesView.js
```diff
@@ -584,35 +599,59 @@ class AnalysesView extends Component {
...
+                              index,
+                              lastSelectedIndex
+                          );
+            } else {
+                newSelected = isSelected
+                    ? selected.filter((selectedID) => selectedID !== id)
+                    : [...selected, id];
```
psarando: The normal, non-range selection was reduced to this.

react-components/src/analysis/view/AnalysesView.js
```diff
@@ -538,7 +552,8 @@ class AnalysesView extends Component {
     }
 
     handleSelectAllClick(event, checked) {
-        if (checked) {
+        const { selected } = this.state;
+        if (checked && !selected.length) {
```
psarando: I added this check so that everything will be selected only if the selection was already empty.

When only some items in the listing are selected, the select-all checkbox icon becomes the indeterminate icon, which looks like a minus sign to me ➖ . Clicking on this will now remove all selections, rather than selecting everything. This is also how Gmail behaves.

react-components/src/analysis/messages.js
```diff
@@ -63,6 +63,12 @@ var intlData = {
             " to extend the time limit?",
         analysesExecDeleteWarning:
             "This will remove the selected analyses and the parameters information associated with those analyses. Outputs can still be viewed in the Data window within the folder created by these analyses.",
+        analysesMultiRelaunchWarning:
+            "Relaunching more than one analysis at once will relaunch all selected analyses, reusing their original parameters and analysis names." +
+            " If the selected analyses are sub-jobs of an HT analysis, then those selected analyses will still be nested under that parent HT analysis," +
+            " and their output folders will also be grouped under that parent HT analysis' output folder," +
+            " but the relaunched sub-jobs will be renamed with a `-redo-#` suffix to differentiate them from their original sub-jobs." +
+            " Otherwise, relaunched analyses will be treated as new analyses, even though they reuse the same name and parameters as their original analyses.",
```
psarando: I'm open to suggestions on the wording for this confirmation dialog (shown to confirm relaunching multiple analyses).

psarando:
Thanks for the review!

PR #479, opened 2020-02-14
CORE-763 Fix AVU number values in Metadata Template forms.

psarando:
The metadata presenter strips number/integer values from AVUs when converting to Autobeans in prep for submission to metadata endpoints.

This fix will convert number values to strings in the metadata form's `handleSubmit` function before passing the metadata to the presenter.


psarando:
Thanks for the review 👍

PR #480, opened 2020-02-15
CORE-414 Improve behavior of "Save" button in Metadata Template forms

psarando:
The Metadata Template "Save" button is no longer disabled when the form has validation errors.
Instead an alert dialog displays informing the user there are errors.

Also removed FAB buttons in favor of Icon buttons.

### DIFF
--- a/react-components/src/metadata/messages.js
+++ b/react-components/src/metadata/messages.js
@@ -44,6 +44,9 @@ export default {
         edit: "Edit",
         enumDefaultLabel: "Default Selection?",
         enumValues: "Enum Values",
+        errMetadataTemplateAlertHeading: "Please fix all errors.",
+        errMetadataTemplateAlertMsg:
+            "1 or more Attributes (or a child attribute) have 1 or more errors.",
         errAVUEditFormTooltip: "Please fix all errors.",
         errAttrHasErrors:
             "Attribute (or a child attribute) has 1 or more errors.",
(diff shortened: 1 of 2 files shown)


react-components/src/metadata/messages.js
```diff
@@ -44,6 +44,9 @@ export default {
         edit: "Edit",
         enumDefaultLabel: "Default Selection?",
         enumValues: "Enum Values",
+        errMetadataTemplateAlertHeading: "Please fix all errors.",
+        errMetadataTemplateAlertMsg:
+            "1 or more Attributes (or a child attribute) have 1 or more errors.",
```
ash3rz: The only thing that might make this better is if (and I don't remember, so maybe not) `formik` provides a list of which form elements contain errors.

psarando: I agree, that would be much better.

Formik allows us to build our error object however we want, so it is mirroring the structure of the template, but it only contains an `error` flag and a validation message currently.

I could modify the error object to also include the name of the attribute that contains the error, then I could build a list of attributes that have validation errors for this message. It would require a little experimenting to figure out how to display nested (child) attributes with validation errors 🤔

ash3rz: Ah ok, I thought I remembered something vaguely like that. Probably better spent putting that effort into the sonora version and leaving the generic message for now?

psarando: Agreed. I thought I would time-box an effort to about an hour, and I was able to build the list of invalid attribute names. Unfortunately I'm not sure how to convert the `<li>` list into a plain string that can be formatted by our i18n functions for `DEAlertDialog`.

So I'll merge this PR as-is for now, and revisit generating error lists like that in Sonora.

psarando:
Thanks for the review 👍

PR #482, opened 2020-03-18
Fix NPE and loading mask issues with Collaborators window/dialog

ash3rz:
* Fix NPE when attempting to add a collaborator for sharing via the "Choose Collaborators" option in the sharing dialog
* Fix Collaborators List dialog hiding automatically after clicking OK to save a Collaborator List.  The user was getting no feedback on whether the List didn't save or whether it was still in the process of saving.


ash3rz:
Thanks for the review!

PR #483, opened 2020-04-03
CORE-968 Fix Analysis View statusClick and handleRowClick.

psarando:
* Updating `statusClick` to use `index` instead of an `analysis` object was missed in #478.
* The `handleRowClick` logic also should have been updated in #478 to set the selected items to only the row clicked.
* The `DotMenu` was also updated to select that row when clicked, so that the single-item actions are available in that menu.
* The `AnalysesView` story `presenter` was updated to also log analysis info in console logs for various methods.

### DIFF
--- a/react-components/stories/analysis/view/AnalysesView.stories.js
+++ b/react-components/stories/analysis/view/AnalysesView.stories.js
@@ -121,7 +121,7 @@ class AnalysesViewTest extends Component {
                 },
                 {
                     description: "",
-                    name: "Word_Count_analysis1",
+                    name: "Word_Count_analysis2",
                     can_share: true,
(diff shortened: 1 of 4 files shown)


react-components/stories/analysis/view/AnalysesView.stories.js
```diff
@@ -305,11 +310,14 @@ class AnalysesViewTest extends Component {
             handleViewAllIconClick: () => {
                 console.log("View All");
             },
+            getVICELogs: (analysisId, name) => {
+                console.log("Get VICE Logs", analysisId, name);
+            },
```
ash3rz, line 315: 👍

psarando:
Thanks for the reviews 👍

PR #484, opened 2020-04-15
CORE-975 Add concurrency limit dialog for interactive jobs

ash3rz:
Wording can definitely be modified for clarity.

image


de-lib/src/main/java/org/iplantc/de/apps/widgets/client/presenter/AppLaunchPresenterImpl.java
```diff
@@ -300,13 +303,30 @@ private HandlerManager createHandlerManager() {
         return new HandlerManager(this);
     }
 
+    private boolean hasConcurrencyError(Throwable caught) {
+        String errMessage = caught.getMessage();
+        return errMessage.matches("\\S+ is already running \\d+ or more concurrent jobs\\n");
```
ash3rz, line 308: Right now, this seems to be the only message returned to the UI.  If the error response gets updated to provide a predefined key, this can be updated to check for that instead.

ash3rz:
Thanks for the review!

PR #487, opened 2020-07-01
Updated VICE limit dialog

ash3rz:
Not yet approved:
image

Approved, but limit reached:
image

Denied:
image


sri2k1us:
Here is the link to Google form - https://forms.gle/JH65bNTj4nVh9qWY7

edwins:
I had one question and one suggestion:

Regarding the approval process, is the DE restricting the launch of a vice app initially or can users still launch a vice app, but just not be able to access the web url without the initial request form?

Instead of a one-time submission of the form, how about showing the user's current total job limit (initially it will be 0) as in the second dialog (e.g. "your account has reached the limit of x concurrent jobs" or "new accounts have a default limit of 0 concurrent jobs") and then let the user fill out the form to increase whenever their limit has been hit. The google form should also reflect the the fact that users may need to submit with each new request, but that probably won't be often.

ash3rz:
> I had a couple of suggestions:
> 
> A suggested edit for the first form, first sentence to briefly provide context about the restriction: "New accounts are restricted from launching interactive apps. To unlock this restriction, please complete the VICE Access Request Form."
> 
> Instead of a one-time submission of the form, how about showing the user's current total job limit (initially it will be 0) as in the second dialog (e.g. "your account has reached the limit of x concurrent jobs" or "new accounts have a default limit of 0 concurrent jobs") and then let the user fill out the form to increase whenever their limit has been hit. The google form should also reflect the the fact that users may need to submit with each new request, but that probably won't be often.

@edwins Yes, any wording suggestions, I'll gladly take.  I will send this to MMS and Tina as well to make sure things sound clear, now that you mention it.

Regarding using the same form for new users and users who need more than the current 2 job limit, I feel a little hesitant to take that approach.  I think the goal with the form is to screen out any bad actors.  Once they've been screened, I like the idea of letting our legit users bypass filling out the form and simply request to have their limit bumped up. We could also mention in the Google form that they can request more resources in the future by filing a support ticket, so it's clear.  How does that sound to you?

edwins:
+1 for having Mary Margaret and Tina involved in proofreading.

I still think it would be useful to gather information about why a user who already has been granted vice app access needs to increase their concurrency limit, in much of the same way we are asking new users to justify access in the first place. Also, it would be good to simplify the google form as well so that its easier to complete. For example I suggest combining resource fields like cores, gpus, etc as one big text field ("Please describe your cpu, ram, and storage requirements for running VICE, if you have any")  -- btw, if you ask a scientific user if they want GPU access, they'll probably say yes -- and things like orcid id should be optional, otherwise this will be a huge burden for students and instructors, esp in a workshop situation. 

My suggestion stems from the UX perspective that there is a preference for processes that are consistent and streamlined. If it helps with an example for when a user needs more resources, Atmosphere has a simple in-app form that has two fields, what resources they need and why they need it. Atmosphere also shows them how much they allocation they have and how much they are using. We have users who sometimes fill out the form multiple times a month, but they always know where to go and what form to use.

But, as I said, these are only a suggestions. :)

sri2k1us:
> +1 for having Mary Margaret and Tina involved in proofreading.
> 
> I still think it would be useful to gather information about why a user who already has been granted vice app access needs to increase their concurrency limit, in much of the same way we are asking new users to justify access in the first place. Also, it would be good to simplify the google form as well so that its easier to complete. For example I suggest combining resource fields like cores, gpus, etc as one big text field ("Please describe your cpu, ram, and storage requirements for running VICE, if you have any") -- btw, if you ask a scientific user if they want GPU access, they'll probably say yes -- and things like orcid id should be optional, otherwise this will be a huge burden for students and instructors, esp in a workshop situation.
> 
> My suggestion stems from the UX perspective that there is a preference for processes that are consistent and streamlined. If it helps with an example for when a user needs more resources, Atmosphere has a simple in-app form that has two fields, what resources they need and why they need it. Atmosphere also shows them how much they allocation they have and how much they are using. We have users who sometimes fill out the form multiple times a month, but they always know where to go and what form to use.
> 
> But, as I said, these are only a suggestions. :)

ORCID id and GPU suggestions were made by Tyson and Reetu (Reetu is using ORCID id for some her Data Mining project). We can make it optional. For `GPUs` question, I will add a `details` sections and will ask the user to fill in additional details in case if they respond `Yes` to `GPUs`.

Coming back to having the user fill the form for additional resources, I agree it makes sense to have them go to one place to request something and that works out well when the form is built inside the app itself. We can revisit this in Sonora for sure.

ash3rz:
Thanks for the feedback and review!  I've pushed the form URL into the code changes and updated the screenshots with the wording updates I've gotten so far.

PR #489, opened 2020-07-22
CORE-1103 Update the DE to handle async tasks

ash3rz:
This will update the UI so that users can be more aware that async tasks are a thing now and to wait for moves/deletes/renames to complete.

Upon initiating a move, users will still see a loading screen in the Data window while waiting for the initial response to come back.  If the response comes back successful, they will get an announcer message letting them know the move is in progress and that a separate notification will be incoming once the transition is completed:
image

When the move has finished, the announcer and the notification dropdown will display a summary message from the notification.  If users want to see the entirety of the message, which will include full file path sources and destinations, they can view that in the Notification window.
image

image


de-lib/src/main/java/org/iplantc/de/notifications/client/utils/NotificationUtil.java
```diff
@@ -129,13 +129,18 @@ public NotificationMessage getMessage(Notification notification,
                         AutoBeanCodex.decode(notFactory, PayloadData.class, payload).as();
                 String dataAction = dataPayload.getAction();
 
-                if ("file_uploaded".equals(dataAction)) {
+                if ("file_uploaded".equals(dataAction) || "UPLOAD_COMPLETE".equals(dataAction)) {
```
ianmcorvidae, line 132: Why is just this one uppercase? Not that completed uploads are part of the async tasks stuff really, but I'm curious, heh

ash3rz: Good question!  I tried to search for where this payload action is set, but I can't find it.  But basically when I clicked on a file upload notification, the Data window wasn't opening like I expected, and then I noticed the actual action in the payload was `UPLOAD_COMPLETE`. Not sure if that means `file_uploaded` is no longer used or if that's an action for something else, so I kept them both for now.

ash3rz:
Thanks for the review!  Since this is a relatively small change, I'll go ahead and merge, but more reviews still welcome!

PR #490, opened 2020-07-24
Remove DiskResourceNameValidator from BulkMetadataViewImpl

ash3rz:
Since the file selector includes the entire resource path, having this validation would always fail on the `/` character being present in the path (e.g. /iplant/home/username).  Considering users can only pick items from the data store and cannot manually type in a file/folder, validating the resource name is unnecessary at this stage anyway.


ash3rz:
Thanks for the review!

PR #492, opened 2020-08-11
Remove totalBad field from all usage, since it isn't really being used anyway

ianmcorvidae:
Pretty straightforward, hopefully, if it's me putting up an old-ui PR :laughing:

Just trying to make it so the UI doesn't use that field at all, so that it won't care if it's there or not.


ianmcorvidae:
Cool, thanks for checking on it. I'll merge it now, I think.

PR #495, opened 2020-11-30
core-1177 fix date formatting in app details.

sri2k1us:


psarando (approved):
LGTM, though I would just suggest allowing `formatDateObject` to return its own empty date string (same suggestions as cyverse-de/sonora#239).

sri2k1us:
Thanks for the review and suggestion.

PR #496, opened 2020-12-19
Update notification unseen_total and total keys as integers

ash3rz:
With the old service, `unseen_total` is a String, and with the new service it's a number.

### DIFF
--- a/de-lib/src/main/java/org/iplantc/de/client/models/notifications/NotificationCategory.java
+++ b/de-lib/src/main/java/org/iplantc/de/client/models/notifications/NotificationCategory.java
@@ -16,7 +16,7 @@ public enum NotificationCategory {
     /** Analysis notifications */
     ANALYSIS("Analysis"),
     /** tool rquest status update notification */
-    TOOLREQUEST("Tool Request"),
+    TOOL_REQUEST("Tool Request"),
     /** Tool sharing/unsharing notifications **/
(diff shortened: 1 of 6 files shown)


ash3rz:
Oops, one more bug I need to fix that I just found

de-lib/src/main/java/org/iplantc/de/client/models/notifications/NotificationCategory.java
```diff
@@ -16,7 +16,7 @@
     /** Analysis notifications */
     ANALYSIS("Analysis"),
     /** tool rquest status update notification */
-    TOOLREQUEST("Tool Request"),
+    TOOL_REQUEST("Tool Request"),
```
ash3rz, line 19: I noticed these notification category types have changed from no spaces to having underlines, so I updated them.  
The only little snaggling bug right now is that Tool Request notifications show up as having no message body in the DE UI for some reason.  Will look into this after the break if nobody else has.

psarando (approved):
LGTM :shipit:

ash3rz:
I'm going to close this PR.  After some discussions with Sarah, we realized that I can't actually make these changes to the UI. At this point, QA has the updated messages service and production does not.  Making a change to the UI in QA to accommodate for the new service will break things in production's UI until it also gets the updated service.  

Changes to the messages service will be incoming.

